PR #114, opened 2023-09-11
Update/create time period entries efficiently

raviks789:
Currently a new time period gets created whenever a time period entry with the same group of contacts and/or contact groups for a given schedule. This nullifies the reason for existence of time periods.

Time periods should act as a bridge between the time period entries and the schedules.

Hence, time periods for a group of contacts and/or contact groups should be same across all time period entries for in a given schedule.

A new time period for a new time period entry should be generated only if there is no existing time period for the given schedule and the group of contacts and/contact groups. If another time period entry is created for the same group of contacts and/or contact groups in the same schedule, existing time period or these entries should be used.

### DIFF
--- a/application/forms/EntryForm.php
+++ b/application/forms/EntryForm.php
@@ -13,6 +13,7 @@ use Icinga\Module\Notifications\Model\ScheduleMember;
 use Icinga\Module\Notifications\Model\TimeperiodEntry;
 use Icinga\Web\Session;
 use ipl\Html\HtmlDocument;
+use ipl\Orm\Behavior\Binary;
 use ipl\Scheduler\Contract\Frequency;
 use ipl\Scheduler\OneOff;
 use ipl\Scheduler\RRule;
@@ -123,6 +124,10 @@ class EntryForm extends CompatForm
 
         $recipients = [];
         foreach ($members as $member) {
+            if (! isset($values['membership_hash'])) {
+                $values['membership_hash'] = (new Binary([]))->toDb($member->membership_hash, null, null);
+            }
+
             if ($member->contact_id !== null) {
                 $recipients[] = 'contact:' . $member->contact_id;
             } else {
@@ -220,6 +225,7 @@ class EntryForm extends CompatForm
         };
 
         $this->addElement('hidden', 'timeperiod_id');
+        $this->addElement('hidden', 'membership_hash');
 
         $this->addElement('textarea', 'description', [
             'label' => $this->translate('Description'),
@@ -311,6 +317,7 @@ class EntryForm extends CompatForm
     public function addEntry(int $scheduleId): void
     {
         $data = $this->formValuesToDb();
+        $membershipHash = $this->getMembershipHashFromRecipients($scheduleId);
         $recipients = array_map(function ($recipient) {
             return explode(':', $recipient, 2);
         }, explode(',', $this->getValue('recipient')));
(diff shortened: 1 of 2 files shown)


application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
+        $members = ScheduleMember::on(Database::get())
+            ->with('timeperiod')
+            ->filter(
+                Filter::all(
+                    Filter::equal('timeperiod.owned_by_schedule_id', $scheduleId),
+                    Filter::equal('schedule_member.membership_hash', $membershipHash)
+                )
+            );
+
+        if ($members->count() > 0) {
```
nilmerg, line 337: I'd rather select from `Timeperiod` directly. That's the only information you're really interested in. The count of members for example, you're using it to check whether there are any members at all. That's a common mistake, I mean, to count although the count itself is of no interest. So by fetching the timeperiod directly, you already verified that at least one member's hash matches. (Given that the filter is still the same)

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
-                    'contact_id' => $id
-                ]);
-            } elseif ($type === 'group') {
-                $db->insert('schedule_member', [
-                    'schedule_id' => $scheduleId,
-                    'timeperiod_id' => $timeperiodId,
-                    'contactgroup_id' => $id
-                ]);
+        if ($members->count() === 0) {
+            $binaryMemberHash = (new Binary([]))->toDb($membershipHash, null, null);
```
nilmerg, line 347:
```suggestion
            $q = ScheduleMember::on(Database::get());
            $binaryMemberHash = $q->getResolver()->getBehaviors($q->getModel())->persistProperty($membershipHash, 'membership_hash');
```

This way you don't hardcode the binary behavior here and it will still work once are compatible with MySQL.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
     }
 
+    /**
+     * Get membership hash for the recipients in timeperiod entry for the given schedule
+     *
+     * @param int $scheduleID
+     *
+     * @return string
+     */
+    public function getMembershipHashFromRecipients(int $scheduleID): string
```
nilmerg, line 377: Please use `$scheduleId` (with a lowercase `d` at the end) instead. `addEntry` does it already.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
+    {
+        $recipients = explode(',', $this->getValue('recipient'));
+        sort($recipients);
+        return sha1(sprintf('schedule:%d,%s', $scheduleID, implode(',', $recipients)), true);
```
nilmerg, line 381: empty line before return, pls

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
+                    ]);
+                }
             }
         }
 
         $db->commitTransaction();
     }
 
+    /**
+     * Get membership hash for the recipients in timeperiod entry for the given schedule
```
nilmerg, line 371: Please be more specific here and less verbose. You don't need to mention timeperiod entries here, that's obvious from the class already. Then, `get` imposes that the value may be be cached or so. At least, I don't see it fit with the implementation. It's not a getter. So maybe rename the method to `calculate...` or `generate...`, or even `hash...`.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
-        $recipients = explode(',', $this->getValue('recipient'));
+        if ($changedMembers->count() > 0) {
+            // Update the entries with timeperiod_id of the changed membership
+            $db->update(
+                'timeperiod_entry',
+                $data + ['timeperiod_id' => (int) $changedMembers->first()->timeperiod_id],
+                ['id = ?' => $id]
+            );
+            $prevTimeperiodEntriesCount -= 1;
+        } elseif ($prevTimeperiodEntriesCount === 1) {
```
nilmerg, line 418: This case doesn't update the entry at all

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
-        $timeperiodId = (int) $this->getValue('timeperiod_id');
-
         $db = Database::get();
-        $db->beginTransaction();
+        $binaryBehavior = new Binary([]);
+
+        $prevTimeperiodId = $this->getValue('timeperiod_id');
+        $suppliedHash = $this->getValue('membership_hash');
+        $calculatedHash = $this->getMembershipHashFromRecipients($scheduleId);
+        $changedMembers = ScheduleMember::on(Database::get())
```
nilmerg, line 393: Here you're also actually only interested in the timeperiod.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
+    }
+
     public function editEntry(int $scheduleId, int $id): void
     {
         $data = $this->formValuesToDb();
-        $timeperiodId = (int) $this->getValue('timeperiod_id');
-
         $db = Database::get();
-        $db->beginTransaction();
+        $binaryBehavior = new Binary([]);
```
nilmerg, line 388: Don't use the behavior directly here as well.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
+            $timeperiodId = $db->lastInsertId();
+
+            $db->update(
+                'timeperiod_entry',
+                $data + ['timeperiod_id' => $timeperiodId],
+                ['id = ?' => $id]
+            );
+
+            $prevTimeperiodEntriesCount -= 1;
+            if ($changedMembers->count() === 0) {
```
nilmerg, line 513: that check is redundant. If this case runs, the count is always zero.

application/forms/EntryForm.php
```diff
@@ -311,114 +317,233 @@ protected function assemble()
...
-                    $usersToRemove[] = $member->contact_id;
-                } else {
-                    unset($users[$member->contact_id]);
-                }
-            } else {
-                if (! isset($groups[$member->contactgroup_id])) {
-                    $groupsToRemove[] = $member->contactgroup_id;
+            $usersToRemove = [];
+            $groupsToRemove = [];
+            $prevMembers = ScheduleMember::on(Database::get())
```
nilmerg, line 437: With this, you'll fetch members of other timeperiods that happen to match the hash by accident.

nilmerg (changes requested):
You didn't adjust `removeEntry`. So it's currently not possible to remove an entry once it's not the only one in a timeperiod.

application/forms/EntryForm.php
nilmerg: Oh, never mind...

nilmerg:
As discussed. This is obsolete since it only complicates the code to achieve a so small benefit and is only relevant in a very specific edge-case.

PR #121, opened 2023-09-19
Add data base target `_main` to controller index action tabs

yhabteab:
When switching to another tab in `col1` while `col2` is active, `col2` must also be closed, as it is not related to the currently active tab in `col1`.

Bildschirmfoto 2023-09-19 um 11 02 59


nilmerg (changes requested):
please adjust gettabs instead

nilmerg (approved):

PR #124, opened 2023-09-21
Show additional daily appointments link

### LINKED ISSUE #109: Schedule configuration doesn't show more than 4 entries per day/week
# Problem

If you define, say 5, entries that start at the same day (not necessarily at the same time), the fifth is not visible.

# Notes

We should provide a hint that there's more and a way to access them. I previously envisioned a *sort-of* pagination for the calendar, which lets the user switch between different layers (pages).

It must be considered though, that different entries can end at different times. So an entry which's start is invisible due to this problem, might still end at a day where enough room exists to show it. Though, currently it's not shown as well and I don't think it's a good idea to show it then as the start is not visible.

# Preparations

I guess a mockup is the best way to come to a solution here. Your turn @flourish86 :wink:

raviks789:
fixes #109

### DIFF
--- a/library/Notifications/Widget/Calendar/BaseGrid.php
+++ b/library/Notifications/Widget/Calendar/BaseGrid.php
@@ -4,6 +4,7 @@
 
 namespace Icinga\Module\Notifications\Widget\Calendar;
 
+use DateInterval;
 use DateTime;
 use DateTimeInterface;
 use Icinga\Module\Notifications\Widget\Calendar;
@@ -34,6 +35,9 @@ abstract class BaseGrid extends BaseHtmlElement
 
     /** @var DateTime */
     protected $end;
+
+    /** @var array Extra counts stored as [date1 => count1, date2 => count2]*/
+    protected $extraEntriesCount = [];
 
     /**
      * Create a new calendar
@@ -137,6 +141,18 @@ abstract class BaseGrid extends BaseHtmlElement
         return $overlay;
     }
 
+    /**
+     * Fetch the count of additional entries for the given date
+     *
+     * @param DateTime $date
+     *
+     * @return int
+     */
+    public function getExtraEntryCount(DateTime $date): int
+    {
+        return $this->extraEntriesCount[$date->format('Y-m-d')] ?? 0;
+    }
+
     protected function assembleGridOverlay(BaseHtmlElement $overlay): void
     {
         $style = (new Style())->setNonce(Csp::getStyleNonce());
@@ -218,21 +234,42 @@ abstract class BaseGrid extends BaseHtmlElement
             }
         }
 
+        $this->extraEntriesCount = [];
         foreach ($occupiedCells as $entry) {
             $continuation = false;
             $rows = $occupiedCells->getInfo();
             foreach ($rows as $row => $hours) {
                 list($rowStart, $rowSpan) = $rowPlacements[spl_object_id($entry)][$row];
+                $colStart = min($hours);
+                $colEnd = max($hours);
+
+                // Calculate number of entries that are not displayed in the grid for each date
                 if ($rowStart > $row + $sectionsPerStep) {
-                    // TODO: Register as +1
+                    $startOffset = (int) (($row / $sectionsPerStep) * ($gridBorderAt / 48) + $colStart / 48);
+                    $endOffset = (int) (($row / $sectionsPerStep) * ($gridBorderAt / 48) + $colEnd / 48);
+                    $startDate = (clone $this->getGridStart())->add(new DateInterval("P$startOffset" . 'D'));
+                    $duration = $endOffset - $startOffset;
+                    for ($i = 0; $i <= $duration; $i++) {
+                        $countIdx = $startDate->format('Y-m-d');
+                        if (! isset($this->extraEntriesCount[$countIdx])) {
+                            $this->extraEntriesCount[$countIdx] = 1;
+                        } else {
+                            $this->extraEntriesCount[$countIdx] += 1;
+                        }
+
+                        $startDate->add(new DateInterval('P1D'));
+                    }
+
                     continue;
                 }
 
-                $rowEnd = $rowStart + $rowSpan;
-                $colStart = min($hours) + 1;
-                $colEnd = max($hours) + 2;
+                $gridArea = $this->getGridArea(
+                    $rowStart,
+                    $rowStart + $rowSpan,
+                    $colStart + 1,
+                    $colEnd + 2
+                );
 
-                $gridArea = $this->getGridArea($rowStart, $rowEnd, $colStart, $colEnd);
                 $entryClass = 'area-' . implode('-', $gridArea);
 
                 $style->add(".$entryClass", [
--- a/library/Notifications/Widget/Calendar/ExtraEntryCount.php
+++ b/library/Notifications/Widget/Calendar/ExtraEntryCount.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Icinga\Module\Notifications\Widget\Calendar;
+
+use DateTime;
+use ipl\I18n\Translation;
+use ipl\Web\Widget\ButtonLink;
+
+class ExtraEntryCount extends ButtonLink
+{
+    use Translation;
+
+    /** @var BaseGrid Grid this extra count is tied to*/
+    protected $grid;
+
+    /** @var DateTime Grid step for which the extra count is being registered */
+    protected $gridStep;
+
+    /**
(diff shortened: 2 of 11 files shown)


nilmerg (changes requested):
* The fourth entry is still shown if there are more than four
* The counter shows `+0 entries` if there are exactly four entries
* The counter shows `+2 entries` even if only one is hidden

library/Notifications/Widget/Schedule.php
```diff
@@ -44,47 +47,59 @@ protected function assembleCalendar(Calendar $calendar): void
...
+            )
+        );
+
+        foreach ($entries->filter($entryFilter) as $entry) {
+            $members = ScheduleMember::on($db)
+                ->with(['timeperiod', 'contact', 'contactgroup'])
+                ->filter(Filter::all(
+                    Filter::equal('schedule_id', $this->schedule->id),
+                    Filter::equal('timeperiod_id', $entry->timeperiod_id)
+                ))
```
nilmerg, line 90: Now the `Filter::all` makes sense. But this way the schedule is filtered multiple times. The entries are linked to a schedule, the members also, but only indirectly. So please accept my suggestion above so that the entries are already filtered.

nilmerg (changes requested):
just a follow-up to my previous review

nilmerg (changes requested):
will continue tomorrow

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -73,6 +77,13 @@ abstract protected function calculateGridEnd(): DateTime;
 
     abstract protected function getNoOfVisuallyConnectedHours(): int;
 
+    /**
+     * Get Number of days in the grid
+     *
+     * @return int
+     */
+    abstract protected function getNoOfDays(): int;
```
nilmerg, line 85: Please remove this, as discussed

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -218,23 +241,40 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
                 $colEnd = max($hours) + 2;
 
                 $gridArea = $this->getGridArea($rowStart, $rowEnd, $colStart, $colEnd);
                 $entryClass = 'area-' . implode('-', $gridArea);
 
+                // Calculate number of entries that are not displayed in the grid for each date
+                if ($rowStart > $row + $sectionsPerStep) {
+                    $stepRow = $rowStart - $this->getSectionsPerStep();
+                    $startOffset = $this->getStepOffset($stepRow, $gridArea[1] - 1);
+                    $endOffset = $this->getStepOffset($stepRow, $gridArea[3] - 2);
```
nilmerg, line 261:
```suggestion
                    $startOffset = (int) (($row / $sectionsPerStep) * ($gridBorderAt / 48) + $colStart / 48);
                    $endOffset = (int) (($row / $sectionsPerStep) * ($gridBorderAt / 48) + $colEnd / 48);
```

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -218,23 +241,40 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
+                            $this->extraEntriesCount[$countIdx] = 1;
+                        } else {
+                            $this->extraEntriesCount[$countIdx] += 1;
+                        }
+
+                        $startDate->add(new DateInterval('P1D'));
+                    }
+
+                    continue;
+                }
```
nilmerg, line 276:
```suggestion
                }

                $gridArea = $this->getGridArea(
                    $rowStart,
                    $rowStart + $rowSpan,
                    $colStart + 1,
                    $colEnd + 2
                );
                $entryClass = 'area-' . implode('-', $gridArea);
```

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -218,23 +241,40 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
             foreach ($rows as $row => $hours) {
                 list($rowStart, $rowSpan) = $rowPlacements[spl_object_id($entry)][$row];
-                if ($rowStart > $row + $sectionsPerStep) {
-                    // TODO: Register as +1
-                    continue;
-                }
-
                 $rowEnd = $rowStart + $rowSpan;
                 $colStart = min($hours) + 1;
                 $colEnd = max($hours) + 2;
```
nilmerg, line 252:
```suggestion
                $colStart = min($hours);
                $colEnd = max($hours);
```

nilmerg (changes requested):

public/css/calendar.less
```diff
@@ -168,6 +178,50 @@
...
   }
 
+  .step {
+    grid-column-end: span @columnsPerDay;
+    grid-row-end: span @rowsPerHour;
+    position: relative;
+  }
+
+  .extra-count {
+    height: auto;
```
nilmerg, line 188: That's the same in every mode, although not in the day mode, yeah. But there is no extra count anyway, so I don't think the rule hurts if applied to every grid. The position isn't that much of a deal as well, I guess.

public/css/calendar.less
```diff
@@ -168,6 +178,50 @@
...
+.extra-count {
+  z-index: 99;
+  background-color: @text-color-inverted;
+  border-radius: 0.25em;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  padding: 0 .25em;
+  color: @icinga-blue;
+}
```
nilmerg, line 201: Please add this to the `.calendar-grid` ruleset. I don't like this being in the global scope.

public/css/calendar.less
```diff
@@ -190,6 +245,16 @@
...
+      .entry-count {
+        display: flex;
+        flex-direction: row-reverse;
+        pointer-events: all;
+        white-space: nowrap;
+
+        a {
+          color: @icinga-blue;
+        }
+      }
```
nilmerg, line 257: remove this, the entry count isn't in the overlay anymore

public/css/calendar.less
```diff
@@ -214,9 +283,11 @@
 /* Design */
 
 .calendar-controls {
+  input[type="date"],
   input[type="month"],
   input[type="week"] {
     background-color: @low-sat-blue;
+    width: auto;
```
nilmerg, line 290: The width should be added to forms.less

public/css/calendar.less
```diff
@@ -246,6 +317,10 @@
     > a {
       text-decoration: none;
     }
+
+    .extra-count:hover {
+      text-decoration: underline;
+    }
```
nilmerg, line 323: Remove this. By using a `ButtonLink` this isn't required.

public/css/calendar.less
nilmerg: And also adjust it like this, since by making it a `ButtonLink` standard styles already apply.

```suggestion
.extra-count {
  z-index: 99;
  position: absolute;
  bottom: 1px; // 1px is the border-width of an entry, this makes it so it doesn't overlap
  right: 1px;
}
```

library/Notifications/Widget/Calendar/ExtraEntryCount.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace Icinga\Module\Notifications\Widget\Calendar;
+
+use DateTime;
+use ipl\I18n\Translation;
+use ipl\Web\Widget\Link;
+
+class ExtraEntryCount extends Link
```
nilmerg, line 9: Please extend `ButtonLink`

public/css/calendar.less
nilmerg: It should look like this in the end:
image

nilmerg (changes requested):

public/css/calendar.less
```diff
@@ -71,6 +71,11 @@
     }
   }
 
+  .overlay {
+    position: relative;
+    z-index: 1;
```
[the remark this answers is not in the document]
raviks789: z-index is required because removing it covers the part of the overlay belonging to the step.

public/css/calendar.less
raviks789: instead, I think, it would be better to set the height of  anchor `a` to 100% if it is not belonging to class `extra-count`.

public/css/calendar.less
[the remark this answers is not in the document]
raviks789: As discussed, will remove the `z-index`, but the `position` property stays.

PR #125, opened 2023-09-25
Reference channels by their ID

yhabteab:


julianbrost (commented):
Gave this a test after a `git rebase main` together with https://github.com/Icinga/icinga-notifications/pull/100 and have no obvious complaints (i.e. didn't find anything that didn't work, but also I didn't have the intention to test every little detail).

Daemon PR should be ready to be merged except for the naming of the schema upgrade file.

library/Notifications/Model/Channel.php
```diff
@@ -45,4 +47,37 @@ public function getDefaultSort()
...
+
+    /**
+     * Fetch and map all the configured channel types to a key => value array
+     *
+     * @return array<string>
+     */
+    public static function fetchChannelTypes(): array
+    {
+        $channels = [];
+        foreach (Channel::on(Database::get()) as $channel) {
```
nilmerg, line 66: Please require the database connection as argument.

library/Notifications/Model/Channel.php
```diff
@@ -45,4 +47,37 @@ public function getDefaultSort()
...
+    {
+        $relations->hasMany('contact', Contact::class)->setJoinType('LEFT');
+        $relations->hasMany('incident_history', IncidentHistory::class)->setJoinType('LEFT');
+        $relations->hasMany('rule_escalation_recipient', RuleEscalationRecipient::class)->setJoinType('LEFT');
+    }
+
+    /**
+     * Fetch and map all the configured channel types to a key => value array
+     *
+     * @return array<string>
```
nilmerg, line 61: Since the keys are not positions, please include them here with `int, ` and document them as being ids

library/Notifications/Model/Contact.php
```diff
@@ -57,6 +57,7 @@ public function getDefaultSort()
 
     public function createRelations(Relations $relations)
     {
+        $relations->belongsTo('channel', Channel::class);
```
nilmerg, line 60: Don't you have to define the candidate key, since it's not just `channel_id`?

library/Notifications/Model/IncidentHistory.php
```diff
@@ -57,7 +57,7 @@ public function getColumnDefinitions()
             'schedule_id'                   => t('Schedule Id'),
             'contactgroup_id'               => t('Contact Group Id'),
             'caused_by_incident_history_id' => t('Caused By Incident History Id'),
-            'channel_type'                  => t('Channel Type'),
+            'channel_id'                  => t('Channel ID'),
```
nilmerg, line 60: please keep the indentation level

library/Notifications/Model/RuleEscalationRecipient.php
```diff
@@ -37,15 +37,10 @@ public function getColumnDefinitions()
             'contact_id'            => t('Contact ID'),
             'contactgroup_id'       => t('Contactgroup ID'),
             'schedule_id'           => t('Schedule ID'),
-            'channel_type'          => t('Channel Type')
+            'channel_id'          => t('Channel ID')
```
nilmerg, line 40: Again, indentation level ;)

nilmerg (changes requested):

library/Notifications/Model/Channel.php
yhabteab: I don't see the benefit of doing that.

nilmerg: It's a non-testable code part less. The form already uses the singleton, so it's fine to redo it once more. But introducing the same dependency in other parts is no option.

nilmerg (approved):

PR #127, opened 2023-09-25
Add pending notifications incident history messages

yhabteab:


julianbrost:
Note that even though I already merged https://github.com/Icinga/icinga-notifications/pull/99, I didn't really test this PR because the current state doesn't work (presumably due to the change channel references to IDs change) and I can't rebase without conflicts (don't look too bad, but I didn't bother resolving them).

nilmerg (changes requested):
The state must be translatable. The value in the database is not a label.

library/Notifications/Model/IncidentHistory.php
```diff
@@ -84,4 +91,18 @@ public function createRelations(Relations $relations)
         $relations->belongsTo('rule_escalation', RuleEscalation::class)->setJoinType('LEFT');
         $relations->belongsTo('channel', Channel::class)->setJoinType('LEFT');
     }
+
+    public static function translateNotificationState(string $state): string
```
nilmerg, line 95: Add documentation

library/Notifications/Model/IncidentHistory.php
```diff
@@ -84,4 +91,18 @@ public function createRelations(Relations $relations)
...
+    {
+        switch ($state) {
+            case 'sent':
+                return t('sent', 'contact.notifications.state');
```
nilmerg, line 99: *notifications.transmission.state* fits better IMHO

nilmerg (changes requested):

library/Notifications/Model/IncidentHistory.php
nilmerg: A docblock is obligatory.

nilmerg (approved):

nilmerg (changes requested):
wait, there are conflicts

nilmerg (approved):

PR #131, opened 2023-09-29
Move `getFilter()` method to `ipl\Web\Compat\CompatController::class`

raviks789:
Depends on: Icinga/ipl-web#210


raviks789:
Will be implented in notifications module for now.

PR #134, opened 2023-10-16
Improve longer appointment style

### LINKED ISSUE #129: Calendar: Visualize entries with end or start scheduled beyond current view
Here are some mockups for visualizing entries that start or end on a point in time that is not within the view.

### Daily
Frame 20

### Weekly
Frame 22

### e.g. Daily
Frame 24

## Click to reveal details

Frame 27

## Monthly

Frame 28

raviks789:
fixes #129

### DIFF
--- a/library/Notifications/Widget/Calendar/BaseGrid.php
+++ b/library/Notifications/Widget/Calendar/BaseGrid.php
@@ -384,13 +384,6 @@ abstract class BaseGrid extends BaseHtmlElement
         }
 
         if ($startText) {
-            $title->addHtml(
-                HtmlElement::create(
-                    'div',
-                    ['class' => 'starts-at'],
-                    $startText
-                )
-            );
             $titleAttr = $startText . ' ' . $titleAttr;
         }
 
@@ -406,7 +399,7 @@ abstract class BaseGrid extends BaseHtmlElement
                 Attributes::create([
                     'datetime' => $entry->getStart()->format(DateTimeInterface::ATOM)
                 ]),
-                Text::create($entry->getStart()->format('H:i'))
+                Text::create($entry->getStart()->format($startText ? 'd/m/y H:i' : 'H:i'))
             ));
         }
 
--- a/library/Notifications/Widget/Calendar/BaseGrid.php
+++ b/library/Notifications/Widget/Calendar/BaseGrid.php
@@ -19,9 +19,24 @@ use ipl\Web\Widget\Link;
 use SplObjectStorage;
 use Traversable;
 
+/**
+ * @phpstan-type ContinuationType self::ACROSS_GRID | self::FROM_PREV_GRID | self::TO_NEXT_GRID | self::ACROSS_EDGES
+ */
 abstract class BaseGrid extends BaseHtmlElement
 {
     use Translation;
+
+    /** @var string Continuation type of the entry row continuing from the previous grid */
+    public const FROM_PREV_GRID = 'from-prev-grid';
+
+    /** @var string Continuation type of the entry row continuing to the next grid */
+    public const TO_NEXT_GRID = 'to-next-grid';
+
+    /** @var string Continuation type of the entry row continuing from the previous grid to the next grid */
+    public const ACROSS_GRID = 'across-grid';
+
+    /** @var string Continuation type of the entry row continuing across edges of the grid */
+    public const ACROSS_EDGES = 'across-edges';
 
     protected $tag = 'div';
 
@@ -169,6 +184,7 @@ abstract class BaseGrid extends BaseHtmlElement
         $gridBorderAt = $this->getNoOfVisuallyConnectedHours() * 2;
 
         $cellOccupiers = [];
+        /** @var SplObjectStorage<Entry, int[][]> $occupiedCells */
         $occupiedCells = new SplObjectStorage();
         foreach ($this->calendar->getEntries() as $entry) {
             $actualStart = $this->roundToNearestThirtyMinute($entry->getStart());
@@ -236,8 +252,12 @@ abstract class BaseGrid extends BaseHtmlElement
 
         $this->extraEntriesCount = [];
         foreach ($occupiedCells as $entry) {
-            $continuation = false;
+            $continuationType = null;
             $rows = $occupiedCells->getInfo();
+            $fromPrevGrid = $gridStartsAt > $entry->getStart();
+            $remainingRows = count($rows);
+            $toNextGrid = false;
+
             foreach ($rows as $row => $hours) {
                 list($rowStart, $rowSpan) = $rowPlacements[spl_object_id($entry)][$row];
                 $colStart = min($hours);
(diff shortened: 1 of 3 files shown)


library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -289,16 +340,92 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
+    /**
+     * Fetch properties for entryHtml
+     *
+     * @param string $entryColor
+     * @param string|null $direction
+     * @param string|null $opposite
+     *
+     * @return array
+     */
+    protected function getEntryHtmlProperties(
```
nilmerg, line 361: Please define as much as possible in CSS already. The only variables are the contact's color values. Since you need them now for more than just a background and border color, just set some CSS variables and reference them instead of generating the entire rules here. I suppose once you did this, this method is obsolete.

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -289,16 +340,92 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
                         'data-col-end' => $gridArea[3]
                     ])
                 );
-                $this->assembleEntry($entryHtml, $entry, $continuation);
+
+                $this->assembleEntry($entryHtml, $entry, $startText, $endText, $continuation);
                 $overlay->addHtml($entryHtml);
 
                 $continuation = true;
```
nilmerg, line 347: Doesn't this already help in deciding whether a gradient is required or not? The only missing thing here, is that an entry which starts outside the current view isn't handled. This only covers a single direction, forward, and needs a counterpart. I didn't check how many of the calculations above are redundant, but since you're not using nor replacing this flag, I think some of them are.

nilmerg (changes requested):

public/css/calendar.less
```diff
@@ -1,4 +1,5 @@
 /* Layout */
+@entry-color: var(--entry-color, @default-bg);
```
nilmerg, line 2: What is this for? Just use `var(--entry-bg)` where this is used instead.

public/css/calendar.less
```diff
@@ -338,6 +425,10 @@
     }
   }
 
+  .column-title:not(:first-of-type) {
+    border-left: 1px solid @gray-lighter;
```
nilmerg, line 429: This doesn't align with the grid. Why is this needed?

nilmerg (changes requested):

public/css/calendar.less
raviks789: I would have to align this, but it is needed. Because in case of Week, it could be confusing to differentiate the recipients if you have many recipients in an entry.

raviks789: As discussed offline this change will be removed from this PR, and an [issue](https://github.com/Icinga/icinga-notifications-web/issues/168) has been opened for this.

public/css/calendar.less
nilmerg, line 1: If the description of an entry is long enough, the end date is not visible.

public/css/calendar.less
```diff
@@ -183,6 +188,10 @@
   .grid,
   .overlay {
     border-left: none;
+
+    .title {
+      width: 100%;
```
nilmerg, line 193: This lets the description begin below the title, needlessly. (Since the end time isn't visible anyway)

public/css/calendar.less
```diff
@@ -258,6 +283,17 @@
 
     .entry {
       pointer-events: all;
+
+      .starts-at,
+      .ends-at {
+        font-weight: normal;
+        color: @text-color-light;
```
nilmerg, line 290: This is style!

public/css/calendar.less
nilmerg, line 1: I'll fix the commented issues in this file myself. I'd like to get this done asap.

nilmerg (changes requested):

public/css/calendar.less
```diff
@@ -324,6 +408,10 @@
         opacity: .8;
       }
     }
+
+    .description {
+      mix-blend-mode: screen;
```
nilmerg: I've checked this with Florian again, and it seems to have no effect, not for the description nor for the title.

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -307,7 +358,47 @@ protected function assembleEntry(BaseHtmlElement $html, Entry $entry, bool $isCo
...
+                HtmlElement::create(
+                    'div',
+                    ['class' => 'starts-at'],
+                    $startText
+                )
+            );
+            $titleAttr = $startText . ' ' . $titleAttr;
+        }
+
+        if ($endText) {
```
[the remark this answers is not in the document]
raviks789: Here, we do not need an elseif. Imagine if you are in day mode and you have both the start text and the end text (I need to add a continuation type for this `FROM_PREV_TO_NEXT_GRID`). In that case you need them both in the title attribute.

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -271,34 +285,71 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
                         'data-col-end' => $gridArea[3]
                     ])
                 );
-                $this->assembleEntry($entryHtml, $entry, $continuation);
+
+                if ($fromPrevGrid) {
+                    $continuationType = self::FROM_PREV_GRID;
+                } elseif ($toNextGrid) {
+                    $continuationType = self::TO_NEXT_GRID;
+                } else {
```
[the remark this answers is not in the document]
raviks789: You are right. It should be only set to `self::TO_NEXT_GRID`, when the entry continues to the next row.

library/Notifications/Widget/Calendar/BaseGrid.php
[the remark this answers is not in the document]
raviks789: Constant `ACROSS_GRID` has been added to address the entries in day mode, that starts in previous grid and continues to next grid.

library/Notifications/Widget/Calendar/BaseGrid.php
```diff
@@ -271,33 +291,66 @@ protected function assembleGridOverlay(BaseHtmlElement $overlay): void
...
     }
 
-    protected function assembleEntry(BaseHtmlElement $html, Entry $entry, bool $isContinuation): void
+    /**
+     * Assemble the entry in the grid
+     *
+     * @param BaseHtmlElement $html Container where to add the entry's HTML
+     * @param Entry $entry The entry to assemble
+     * @param ?ContinuationType $continuationType Continuation type of the entry's HTML
+     */
```
nilmerg, line 352: missing return type in doc

nilmerg (changes requested):
please rebase

nilmerg (approved):

PR #137, opened 2023-10-20
Calendar: Make `$url` nullable

### LINKED ISSUE #136: `Calendar::url` property should be nullable
If a schedule has not yet been created, the `Calendar::url` property is not set and hence it should be nullable.

raviks789:
fixes #136

### DIFF
--- a/library/Notifications/Widget/Calendar.php
+++ b/library/Notifications/Widget/Calendar.php
@@ -51,7 +51,7 @@ class Calendar extends BaseHtmlElement
     /** @var Url */
     protected $addEntryUrl;
 
-    /** @var Url */
+    /** @var ?Url */
     protected $url;
 
     public function setControls(Controls $controls): self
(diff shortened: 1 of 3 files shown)


library/Notifications/Widget/Calendar.php
```diff
@@ -89,13 +89,14 @@ public function setUrl(?Url $url): self
         return $this;
     }
 
-    public function prepareDayViewUrl(DateTime $date): Url
+    public function prepareDayViewUrl(DateTime $date): ?Url
```
nilmerg, line 92: The callers still pass on. Please handle it properly.

nilmerg (changes requested):

library/Notifications/Widget/Calendar/WeekGrid.php
```diff
@@ -118,8 +118,8 @@ protected function createSidebar(): BaseHtmlElement
 
     protected function assembleGridStep(BaseHtmlElement $content, DateTime $step): void
     {
-        if ($step->format('H') === '23') {
-            $dayViewUrl = $this->calendar->prepareDayViewUrl($step);
+        $dayViewUrl = $this->calendar->prepareDayViewUrl($step);
```
nilmerg, line 121: If you'd define this inside the first `if`, the url wouldn't be cloned and whatnot in 99% (23 times per day) of the time.

nilmerg (changes requested):

PR #140, opened 2023-10-23
EntryForm: Validate entry duration against its frequency

### LINKED ISSUE #139: Recurrent entries should not overlap with their previous occurrence
Currently you could choose the duration of a recurrent entry greater than its repetition time. For example an entry with a duration of 2 days can be set to repeat daily. This results in the overlapping of the entry with its previous occurrence which should not be possible.

raviks789:
Entry duration should be shorter than their frequency for recurrent entries.

resolves #139

### DIFF
--- a/application/forms/EntryForm.php
+++ b/application/forms/EntryForm.php
@@ -180,6 +180,20 @@ class EntryForm extends CompatForm
                             );
 
                             return false;
+                        } elseif ($this->getValue('frequency') !== self::NO_REPEAT) {
+                            /** @var Frequency $frequency */
+                            $frequency = $this->getValue();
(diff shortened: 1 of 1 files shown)


application/forms/EntryForm.php
```diff
@@ -179,6 +179,20 @@ protected function assemble()
                             );
 
                             return false;
+                        } elseif ($this->getValue('frequency') !== self::NO_REPEAT) {
+                            $next = $this
+                                ->getValue()
+                                ->getNextDue($this->getValue('start'));
```
nilmerg, line 185: `getValue('start') === $value`, right?

nilmerg (approved):

Issue #135, opened 2023-10-18
Desktop Notifications


nilmerg:
## What's this about?

The idea that Icinga Web displays notifications on the desktop is a very [old one](https://github.com/Icinga/icingaweb2/issues/2857). What better reason to add/research that is it, that we are about to re-imagine how notifications in Icinga work?

## Research?

We don't want to show only in-app notifications. Desktop notifications are to be preferred here, as they appear even if the browser tab is not focused. Though, we don't want to rely on a push service such as Google, due to privacy concerns. Thus we need to push the notifications on our own.

For this we need a *service worker* and our own persistent communication. Since this does not exist in Icinga Web until now, we first have to play with it to see how we'd like to integrate this into our framework.

The goal is to integrate this eventually into Icinga Web natively and Icinga Notifications is only a component relying on it. Though, until we chose a final implementation, Icinga Notifications is the sole playground. That way we are independent of the Icinga Web release cycle and have the freedom to make significant changes.

## Related Technologies

* [ServiceWorker API](https://caniuse.com/mdn-api_serviceworker)
* [WebSocket API](https://caniuse.com/mdn-api_websocket)
* [HTTP/2](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_2)

## Architecture

Browser ⇆ Webserver ⇆ Backend ⟶ Database ⟵ Notifications

The browser establishes a connection to our backend using the usual endpoint. The backend is a daemon written in PHP which runs continuously besides the web-server and polls the database.

The question here is, whether websockets or HTTP/2 should be used. Websockets seem to be fairly old already and HTTP/2 would save us some hassles I believe.

The advantages and disadvantages need to be identified here, especially with authentication in mind (as noted below). This might also influence whether we actually need a daemon in the backend. (Even if it only changes a **must** to a **can**)

## Authentication

A connection must be authenticated. Icinga Web has the only authority to do so. So the connection must only be established after the user logged successfully into Icinga Web. The best would be if there are no or just minimal changes to the webserver configuration necessary.

This would allow us to still rely on the browser transmitting the session cookie with each connection attempt, so the request can be authenticated with the current session mechanism in Icinga Web.

> **Hint**
> Since the functionality won't be integrated into Icinga Web for now, it might be difficult to authenticate the user over a websocket connection. Notifications Web might need to provide a `AuthenticationHook` and persist the session on its own so that the backend daemon has access to it.

## Browser Integration

[...]

PR #146, opened 2023-11-27
Introduce desktop notifications

ncosta-ic:
Targets #135

Depends on https://github.com/Icinga/icinga-notifications/pull/136
Depends on https://github.com/Icinga/icinga-notifications/pull/215


ncosta-ic:
## Current observations
The general logic that triggers the desktop notifications is pretty stable now. However, there are some issues with mobile devices, that are currently unsolved.

## iOS
- iOS 16 or greater is required, as the Notification API only got introduced with Safari (iOS) 16.4.
- A web application is only allowed to make use of the Notification API if it's being run as a home-screen application (share -> add to home-screen).
- The logic itself works fine for as long as the home-screen application stays active.
- Leaving the application or locking the phone results in the operating system (iOS) freezing the browser activities after a short amount of time (around 5 seconds).
This behaviour kills all external connections while not running any browser logic, which causes some unwanted issues.
- After unlocking the phone or by going back to the home-screen application, the browser itself gets unfreezed, which causes the tabs and web workers to continue their workflows.
- The event-stream should normally reconnect once it looses its connection [(see second to last paragraph)](https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model). This won't happen as it's proxied by the service worker, which itself pipes the proper event-stream from the daemon back to the tab. During the freeze phase, this connection gets dropped and the stream stays empty and locked (only an assumption, it's not possible to *remote* debug service workers on iOS).

This results in the app stopping to show notifications after the home-screen application was put into background or if the phone got locked. 

## Android
- Untested

## Possible fix?
One way around this problem, albeit being a real ugly one, would be to start a freshness timer in the service worker context once the tab gets the [visiblestate event](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange event) fired with a value of `hidden`. This event is always fired once the browser stops rendering a tab, even on mobile phones and even when putting a PWA into the background or locking the screen.
As the service worker is going to get freezed as well, the freshness checks will get delayed. Once it unfreezes again, it continues with its freshness checks but notices a time delay from the last check to the current.
This could then in return trigger a reconnection on the tabs and the whole feature would continue to work. The repeating freshness checks could then be stopped as the visibilitychange event would be retriggered with a value of  `visible`.

public/js/notifications-worker.js
```diff
@@ -0,0 +1,278 @@
+// noinspection JSUnresolvedReference
+
+const _PREFIX = '[notifications-worker] - ';
+const _SERVER_CONNECTIONS = {};
+let _BASE_URL = '';
+
+if (!(self instanceof ServiceWorkerGlobalScope)) {
```
raviks789, line 7:
```suggestion
if (! (self instanceof ServiceWorkerGlobalScope)) {
```

raviks789: I would suggest having a space between the operator and expression.

raviks789 (changes requested):

run.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use Icinga\Application\Modules\Module;
+
+/* Icinga Web 2 | (c) 2018 Icinga Development Team | GPLv2+ */
```
raviks789, line 5:
```suggestion
/* Icinga Web 2 | (c) 2023 Icinga Development Team | GPLv2+ */
```

library/Notifications/Daemon/Server.php
```diff
@@ -0,0 +1,470 @@
...
+     *
+     * @param ?string $username
+     *
+     * @return ?int contact identifier or null if no contact could be matched
+     */
+    protected function matchContact(?string $username): ?int
+    {
+        /**
+         * TODO(nc): the matching needs to be properly rewritten once we decide about how we want to handle the contacts
+         *  in the notifications module
```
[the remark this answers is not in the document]
ncosta-ic: @raviks789 The space has a purpose. This allows a multiline todo comment in PhpStorm (see: https://www.jetbrains.com/help/phpstorm/using-todo.html#add-multiline-todo)

raviks789 (dismissed):

library/Notifications/Daemon/Daemon.php
```diff
@@ -0,0 +1,361 @@
...
+
+                                break;
+                            case 'service':
+                                $service = $tag->value;
+
+                                break;
+                        }
+                    }
+
+                    self::$logger::warning(self::PREFIX . "Host: " . $host . " | Service: " . $service);
```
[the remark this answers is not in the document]
raviks789: Otherwise the log message for host would look something like this:
```
[daemon] - Host: test-host | Service:
```

Expected:
```
[daemon] - Host: dummy-902
```

raviks789 (commented):
So far looks good to me.

public/js/notifications-worker.js
```diff
@@ -0,0 +1,276 @@
...
+                    title += "'" + notification.payload.host + "'";
+
+                    self.registration.showNotification(
+                        title,
+                        {
+                            icon: _BASE_URL + '/img/notifications/icinga-notifications-' + severity + '.webp',
+                            body: 'changed to severity ' + severity,
+                            data: {
+                                url:
+                                    _BASE_URL
```
sukhwinder33445, line 112: The `_BASE_URL` is somehow still empty because the `View Incident` does not contain the prefix `icingaweb2`. Which leads to `404` page.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+
+            // listen to controller (service worker) changes
+            navigator.serviceWorker.addEventListener('controllerchange', (event) => {
+                this.logger.debug(this.prefix + "new controller attached ", event.target.controller);
+                if (event.target.controller !== null) {
+                    // reset eventsource and handshake flag
+                    this.allowedToOperate = false;
+                    if (this.eventSource !== null && this.eventSource.readyState !== EventSource.CLOSED) {
+                        this.eventSource.close();
+                    }
```
sukhwinder33445, line 86: This code exists more than twice. Please wrap in into a function `closeEventSource()`.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+            this.getServiceWorker()
+                .then((serviceWorker) => {
+                    if (! serviceWorker) {
+                        // no service worker registered yet, registering it
+                        self.navigator.serviceWorker.register(icinga.config.baseUrl + '/notifications-worker.js', {
+                            scope: icinga.config.baseUrl + '/',
+                            type: 'classic'
+                        }).then((registration) => {
+                            let callback = (event) => {
+                                console.log("YEP");
```
sukhwinder33445, line 174: Remove this please.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+
+                try {
+                    this.logger.debug(this.prefix + "opening event source");
+                    this.eventSource = new EventSource(icinga.config.baseUrl + '/notifications/daemon', {withCredentials: true});
+                    this.eventSource.addEventListener('message', () => {
+                        // this._logger.debug(this._prefix + `got message from event-stream: ${event.data}`);
+                    });
+                    this.eventSource.addEventListener('error', () => {
+                        // this._logger.debug(this._prefix +`got an error from event-stream: ${event}`);
+                    });
```
sukhwinder33445, line 238: Please remove these empty listeners.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                    this.eventSource.addEventListener('icinga2.notification', (event) => {
+                        if (this.hasNotificationPermission() && this.hasNotificationsEnabled()) {
```
sukhwinder33445, line 240: As described above, please return earlier if the condition does not match.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                                                baseUrl: icinga.config.baseUrl
+                                            })
+                                        );
+                                    }
+                                });
+                        }
+                    });
+                } catch (error) {
+                    // this._logger.error(this._prefix + `got an error: ${error}`);
+                }
```
sukhwinder33445, line 258: Please uncomment or remove.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+            this.logger.debug(this.prefix + "unloaded");
+        }
+
+        reload () {
+            this.unload();
+            this.load();
+        }
+
+        openEventStream () {
+            if (this.hasNotificationPermission() && this.hasNotificationsEnabled()) {
```
sukhwinder33445, line 224: I would flip the if, else condition and return earlier.
```suggestion
            if (! this.hasNotificationPermission() || ! this.hasNotificationsEnabled()) {
             // the log from else condition
             
                return;
            }
            
            // if condition body
```

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                    // this._logger.error(this._prefix + `got an error: ${error}`);
+                }
+            } else {
+                this.logger.warn(this.prefix + "denied opening event-stream as the notification permissions" +
+                    " are missing or the notifications themselves disabled");
+            }
+        }
+
+        renderHandler (event) {
+            if (event.type === 'rendered') {
```
sukhwinder33445, line 266: This condition must be superfluous, as the event type must always be the `rendered`.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                 *  This in turn requires a user to be logged in and their session validated.
+                 *  In the future, we should introduce a proper login event and tie the initial event-stream connection
+                 *  to this specific event (SSO should ALSO trigger the login event as the user lands in the
+                 *  interface with an authenticated session).
+                 */
+                if (_this.initialized === false) {
+                    _this.initialized = true;
+                }
+
+                if (url.pathname === _this.icinga.config.baseUrl + '/account') {
```
sukhwinder33445, line 283: As above, please return earlier if the condition is not met. This prevents indentation.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                    // check permissions and storage flag
+                    const state = (
```
sukhwinder33445, line 285: Please remove the superfluous parenthesis.

public/js/notifications-worker.js
```diff
@@ -0,0 +1,315 @@
...
+                self.clients.openWindow(event.notification.data.url).then();
+                break;
+            case 'dismiss':
+                break;
+        }
+    }
+});
+
+function processMessage (event) {
+    if (event.data) {
```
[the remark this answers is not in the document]
ncosta-ic: Not sure about the benefit this brings, but I changed it anyways

public/js/notifications-worker.js
```diff
@@ -0,0 +1,315 @@
...
+selfSW.addEventListener('install', (event) => {
+    event.waitUntil(selfSW.skipWaiting().then());
+})
+selfSW.addEventListener('fetch', (event) => {
+    // self.console.log(_PREFIX + 'fetch event triggered with: ', event);
+    const request = event.request;
+    const url = new URL(event.request.url);
+
+    // only check dedicated event stream requests towards the daemon
+    if (request.headers.get('accept').startsWith('text/event-stream') && url.pathname.trim() === '/icingaweb2/notifications/daemon') {
```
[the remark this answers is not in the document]
sukhwinder33445: However, the main reason for the suggested change is that the entire function body is unnecessarily included in this if condition. We will avoid this indentation.

public/js/notifications-worker.js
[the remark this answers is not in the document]
sukhwinder33445: The benefit is that you avoid unnecessary indentation. It is not best practice to unnecessarily enclose the entire function body in a condition.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+                            command: 'handshake',
+                            version: VERSION
+                        })
+                    );
+                }
+            });
+
+            // listen to messages from the controller (service worker)
+            self.navigator.serviceWorker.addEventListener('message', (event) => {
+                if (event.data) {
```
sukhwinder33445, line 100: Please return earlier here.

sukhwinder33445: And make debug logs single-line if the line length is `> 120`.

public/js/notifications.js
```diff
@@ -0,0 +1,433 @@
...
+
+                                // the controller declared us as an outdated script version
+                                this.icinga.loader.createNotice(
+                                    'warning',
+                                    'This tab is running an outdated script version. Please reload the page!',
+                                    true
+                                );
+
+                                this.allowedToOperate = false;
+                            } else if (data.status === 'accepted') {
```
sukhwinder33445, line 118: The condition seems to be superfluous here.

public/js/notifications.js
ncosta-ic: the line length is capped to a maximum of 120 in this file though

public/js/notifications.js
ncosta-ic: This has been done on purpose, as in the future other status types could be introduced by the service worker. We currently only check for the tab version validity, but it's technically not limited to that.
As an example, the service worker could accept the handshake but with additional requirements.

public/js/notifications-worker.js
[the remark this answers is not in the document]
ncosta-ic: Yeah I somehow missed that this was suggested for an early return

ncosta-ic:
@raviks789 
> So far looks good to me.

As it seems like you're fine with the changes that I implemented, you might want to approve your review. It still needs the approval of @sukhwinder33445 and @nilmerg either way.
Thanks!

nilmerg (approved):
I did only functional tests and directly fixed issues. Didn't look at 98% of the code.

nilmerg:
Problems that are left open:

* Apache seems not to immediately close connections upon disconnect of either peer.
* A high number of transmitted messages might cause the daemon to continuously error and not recover from it.

ncosta-ic:
> Problems that are left open:
> 
>     * Apache seems not to immediately close connections upon disconnect of either peer.
> 
>     * A high number of transmitted messages might cause the daemon to continuously error and not recover from it.

### And another small issue
`Cannot read properties of null (reading 'querySelector')` is getting thrown through the `parentNode` call at [notifications.js#L285](https://github.com/Icinga/icinga-notifications-web/blob/9f9517701ff8b83dd98387a8a903f4b99aee9bb0/public/js/notifications.js#L285) when login takes place with a redirect onto the account page, e.g through: http://localhost/icingaweb2/authentication/login?redirect=account

nilmerg (changes requested):
Please reflect required changes for these here:

* https://github.com/Icinga/icinga-notifications-web/issues/194
* https://github.com/Icinga/icinga-notifications-web/issues/195

ncosta-ic:
> Please reflect required changes for these here:
> 
>     * [Schema: browser_session: let only php_session_id be the PK #194](https://github.com/Icinga/icinga-notifications-web/issues/194)
> 
>     * [Schema: browser_session: let the code insert authenticated_at #195](https://github.com/Icinga/icinga-notifications-web/issues/195)

Those issues get resolved in https://github.com/Icinga/icinga-notifications/pull/215

nilmerg (approved):

nilmerg:
@ncosta-ic Please hit "Merge pull request" once https://github.com/Icinga/icinga-notifications/pull/215 is merged. But please make sure it's still up to date with main prior.

PR #156, opened 2024-01-16
Add source configuration UI

### LINKED ISSUE #147: Add config forms for source table
Currently, the `source` table has to be populated manually by the user by issuing insert statements to the database. As the daemon loads most of it's configuration from the database, additional source configuration should also be managed from the web module.

With upcoming PRs in the daemon, there will be extensions to the `source` table, that should be possible to configure from web. This issue is kind of a heads-up notice, the details will be finalized in https://github.com/Icinga/icinga-notifications/pull/132 and https://github.com/Icinga/icinga-notifications/pull/112 but roughly are the following:

* `type = 'icinga2'` will become special in the way that it enables extra functionality in the daemon to pull events from the Icinga 2 API. Therefore, extra columns will be added for the API URL/user/password/TLS config that should be populated only for this type.
* All other types will use the existing HTTP listener that allow submitting events via HTTP POST request. This will be protected with basic auth and therefore, a column for a password hash (bcrypt) will be added (no column for the username, that's basically the source ID).

nilmerg:
resolves #147


oxzi:
FYI, @julianbrost's [comment regarding a custom certificate Common Name](https://github.com/Icinga/icinga-notifications-web/issues/147#issuecomment-1893785058) was now implemented in Icinga/icinga-notifications#112's commit Icinga/icinga-notifications@24a484370dd372d6f657f92a2c554ec0965228bd.

nilmerg:
@oxzi A functional review and whether the options are correctly managed/stored is sufficient :wink:

application/forms/SourceForm.php
```diff
@@ -0,0 +1,311 @@
...
+                            'The CN of the API certificate. Only required if it differs from the host name'
+                        )
+                    ]
+                );
+
+                $this->addElement(
+                    'textarea',
+                    'icinga2_ca_pem',
+                    [
+                        'cols'          => 65,
```
oxzi, line 140:
```
root@dc0735581235:/go/src/github.com/Icinga/icinga-notifications# openssl s_client -connect docker-master:5665 -showcerts < /dev/null 2> /dev/null | awk '/BEGIN CERTIFICATE/ || p { p = 1; print } /END CERTIFICATE/ { exit }' | awk '{ max = (length($0) > max) ? length($0) : max } END { print max }'
64
```

*SCNR*

application/forms/SourceForm.php
```diff
@@ -0,0 +1,311 @@
...
+                        'rows'          => 28,
```
oxzi, line 141:
```
root@dc0735581235:/go/src/github.com/Icinga/icinga-notifications# openssl s_client -connect docker-master:5665 -showcerts < /dev/null 2> /dev/null | awk '/BEGIN CERTIFICATE/ || p { p = 1; print } /END CERTIFICATE/ { exit }' | wc -l
29
```

I'm getting a scroll bar for one line.
But, on the other hand, this value is only useful as long we are dealing with 4096 bit RSA keys ¯\\\_(ツ)_/¯

However, would it be possible to use a monospace font here?

application/forms/SourceForm.php
```diff
@@ -0,0 +1,311 @@
...
+            /** @var CheckboxElement $insecureBox */
+            $insecureBox = $this->getElement('icinga2_insecure_tls');
+            if ($insecureBox->isChecked()) {
+                $this->addElement(
+                    'text',
+                    'icinga2_common_name',
+                    [
+                        'label'         => $this->translate('Common Name'),
+                        'description'   => $this->translate(
+                            'The CN of the API certificate. Only required if it differs from the host name'
```
oxzi, line 131: Maybe emphasize that this is the FQDN from the _API URL_ field above as no other "host name" was mentioned otherwise?

application/forms/SourceForm.php
nilmerg: I deliberately used what fits the certificate I got from Icinga's api setup. And I don't get a scrollbar. :thinking:

Using a monospace font makes absolutely sense. Thought of this already but forgot about it.

application/forms/SourceForm.php
nilmerg: What is the fqdn? The host in the url or the cn of the certificate? My understanding is, if the cn in the certificate doesn't match the actual hostname (as in the url field), one needs to define it here.

nilmerg:
> However, I realized that it is impossible to remove a source once an object was created linked to it, but this might be another day's issue:

Yep. That's not only a problem here. Any other configuration is affected by this as well.

application/forms/SourceForm.php
oxzi: That's how it's looking for me with a certificate generated from Icinga 2:
2024-01-19-155045_798x919_scrot

application/forms/SourceForm.php
oxzi: Sorry, the Fully Qualified Domain Name of the URL from the `icinga2_base_url` field, labeled as "API URL".

application/forms/SourceForm.php
nilmerg: I'm a little confused now. What the command you posted returns, isn't this the peer certificate? Though, what the source configuration requires is a certificate to validate this. i.e. the CA certificate. Which has 28 lines. :D

oxzi: I tricked myself and haven't used the CA, but the Icinga 2 master's certificate.

```
icinga@0e990fad2029:/$ openssl s_client -connect icinga2-master1:5665 -showcerts < /dev/null
[ . . . ]
Certificate chain
 0 s:CN = icinga2-master1
   i:CN = Icinga CA
-----BEGIN CERTIFICATE-----
[ 29 lines including BEGIN and END lines ]
-----END CERTIFICATE-----
 1 s:CN = Icinga CA
   i:CN = Icinga CA
-----BEGIN CERTIFICATE-----
[ 28 lines including BEGIN and END lines ]
-----END CERTIFICATE-----
[ . . . ]
```

The script, however, returns the first certificate within the chain, which is unfortunately not the CA.

Thus, this should be fine, I guess.

PR #159, opened 2024-02-21
Event rule config enhancement

raviks789:
Use a single form for event rule configuration.

Blocked by Icinga/icingaweb2#5190


application/forms/EventRuleConfigElements/EscalationRecipient.php
```diff
@@ -0,0 +1,266 @@
...
+
+        return $options;
+    }
+
+    /**
+     * Create remove button for the recipient in the given position
+     *
+     * @param int $pos
+     *
+     * @return FormElement|null
```
sukhwinder33445, line 188: `?FormElement`

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -0,0 +1,320 @@
...
+                ]
+            );
+
+            $operators = ['=', '>', '>=', '<', '<=', '!='];
+            /** @var BaseFormElement $op */
+            $op = $this->createElement(
+                'select',
+                $opName,
+                [
+                    'class'    => ['class' => 'operator-input', 'autosubmit'],
```
sukhwinder33445, line 113: Why? must be a mistake, right?

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -0,0 +1,132 @@
...
+        if ($showSearchBar === '0') {
+            /** @var SubmitButtonElement $filterElement */
+            $filterElement = $addFilterButton;
+            $this->addAttributes(['class' => 'empty-filter']);
+        } else {
+            $editorOpener = new Link(
+                new Icon('cog'),
+                $this->getSearchEditorUrl(),
+                Attributes::create([
+                    'class'               => 'search-editor-opener control-button',
```
sukhwinder33445, line 62: Please use array.

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -0,0 +1,132 @@
...
+                    'title'               => t('Adjust Filter'),
+                    'data-icinga-modal'   => true,
+                    'data-no-icinga-ajax' => true,
+                ])
+            );
+
+            $searchBar = new TextElement(
+                'searchbar',
+                [
+                    'class'    => 'filter-input control-button',
```
sukhwinder33445, line 72: Please use array.

library/Notifications/Widget/ItemList/Escalation.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Set if escalation has remove button
+     *
+     * @param bool $state
+     *
+     * @return $this
+     */
+    public function setHasNoRemovedButton(bool $state): self
```
sukhwinder33445, line 90: A default value for the parameter please.

sukhwinder33445: Cannot see any usage, maybe this method is no more required?

sukhwinder33445 (changes requested):

library/Notifications/Widget/ItemList/Escalation.php
raviks789: This method is unnecessary

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -0,0 +1,132 @@
...
+            $searchBar = new TextElement(
+                'searchbar',
+                [
+                    'class'    => ['filter-input', 'control-button'],
+                    'readonly' => true,
+                    'value'    => $this->objectFilter
+                ]
+            );
+
+            $filterElement = Html::tag('div', ['class' => 'search-controls icinga-controls']);
```
sukhwinder33445, line 78: Classes in array, please.

And since your content is static, please try to use `new HtmlElement()`. The processing of `Html::tag` is time-consuming.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,698 @@
...
+
+        foreach ($prefixesMap as $key => $prefixMap) {
+            if ($removePosition === $prefixMap) {
+                if ($zeroConditionEscalation === $prefixMap) {
+                    $zeroConditionEscalation = null;
+                }
+
+                unset($prefixesMap[$key]);
+                $this->getElement('prefixes-map')->setValue(implode(',', $prefixesMap));
+                continue;
```
sukhwinder33445, line 201: Please add a new line before continue.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,698 @@
...
+            }
+
+            $escalationCondition = new EscalationCondition(
+                'escalation-condition_' . $prefixMap
+            );
+
+            $escalationRecipient = new EscalationRecipient(
+                'escalation-recipient_' . $prefixMap
+            );
```
sukhwinder33445, line 210: One line please.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,698 @@
...
+            if ($escalationId && is_numeric($escalationId)) {
+                $incident->filter(Filter::equal('rule_escalation.id', $escalationId));
+                if ($incident->count() > 0) {
+                    $disableRemoveButton = true;
+                }
+            }
+        }
+
+        /** @var SubmitButtonElement $button */
+        $button = $this->createElement(
```
sukhwinder33445, line 435: Use `SubmitButtonElement` please, so it matches the return type.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,698 @@
...
+            ->getAttributes()
+            ->registerAttributeCallback('disabled', function () use ($disableRemoveButton) {
+                return $disableRemoveButton;
+            })
+            ->registerAttributeCallback('title', function () use ($disableRemoveButton) {
+                if ($disableRemoveButton) {
+                    return $this->translate(
+                        'There exist active incidents for this escalation and hence cannot be removed'
+                    );
+                }
```
sukhwinder33445, line 461: I think callbacks are obsolete here. You can add the attributes directly.

library/Notifications/Widget/ItemList/EscalationConditionList.php
```diff
@@ -0,0 +1,34 @@
...
+    public function __construct(array $conditions)
+    {
+        $this->conditions = $conditions;
+    }
+
+    protected function assemble(): void
+    {
+        $this->add([
+            $this->conditions
+        ]);
```
sukhwinder33445, line 32:
```suggestion
        $this->add($this->conditions);
```

library/Notifications/Widget/ItemList/EscalationRecipientList.php
```diff
@@ -0,0 +1,34 @@
...
+    public function __construct(array $recipients)
+    {
+        $this->recipients = $recipients;
+    }
+
+    protected function assemble(): void
+    {
+        $this->add([
+            $this->recipients
+        ]);
```
sukhwinder33445, line 32:
```suggestion
        $this->add($this->recipients);
```

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->removeButton = $removeButton;
+    }
+
+    protected function assemble(): void
+    {
+        $this->add([
+            $this->recipient,
+            $this->channel,
+            $this->removeButton
+        ]);
```
sukhwinder33445, line 42: You can call this method directly in constructor.

library/Notifications/Widget/ItemList/EscalationRecipientList.php
sukhwinder33445, line 32: You can call this method directly in constructor.

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    protected function assemble(): void
+    {
+        $this->add([
+            $this->conditionType,
+            $this->operator,
+            $this->conditionVal,
+            $this->removeButton
+        ]);
```
sukhwinder33445, line 48: You can call this method directly in constructor. And remove the assemble method.

library/Notifications/Widget/ItemList/EscalationConditionList.php
sukhwinder33445, line 32: You can call this method directly in the constructor. This makes the assemble method and the class properties superfluous.

library/Notifications/Widget/ItemList/Escalations.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->escalations = $escalations;
+        $this->addButton = $addButton;
+    }
+
+    protected function assemble(): void
+    {
+        $this->add([
+            $this->escalations,
+            Html::tag('li', ['class' => 'add-escalation'], $this->addButton)
+        ]);
```
sukhwinder33445, line 40: You can call this method directly in constructor.

sukhwinder33445 (changes requested):

sukhwinder33445 (changes requested):
I think you can move all `edit/add` and `search-editor` related methods from `EventRulesController` to `EventRuleController`. Since the `EventRulesController` should only show the `add` button and the list items.

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
raviks789: There is a reason, this is done in assemble. We want the recipient list assembled as late as possible, as in case of single recipient the remove button would be remove from recipient fieldset.

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
raviks789: There is a reason, this is done in assemble. While in case of single condition you would have to not show the remove button based on the condition count and whether there is an escalation with no condition. We want the condition list to be assembled as late as possible.

application/controllers/EventRuleController.php
```diff
@@ -32,81 +34,120 @@ class EventRuleController extends CompatController
...
+                    'class'          => 'btn-discard-changes',
+                    'formnovalidate' => true,
+                ]
+            ));
+            $disableSave = false;
         }
 
-        $disableRemoveButton = false;
-        if (ctype_digit($ruleId)) {
+        $buttonsWrapper = new HtmlElement('div', Attributes::create(['class' => ['icinga-controls', 'save-config']]));
```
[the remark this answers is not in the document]
raviks789: This is not longer than 120 characters.

application/controllers/EventRulesController.php
```diff
@@ -181,47 +176,68 @@ public function completeAction(): void
...
-                        ['use_cache' => true]
+                        ['id' => $ruleId]
                     )
                 );
         });
 
         $editor->handleRequest($this->getServerRequest());
 
-        $this->getDocument()->addHtml($editor);
+        $this->getDocument()->add($editor);
```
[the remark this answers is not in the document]
raviks789: Don't remember the reason exactly. But I think it was something to do with circumventing the phpstan error. Will undo these changes.

application/controllers/EventRuleController.php
```diff
@@ -32,81 +34,120 @@ class EventRuleController extends CompatController
...
+            'delete',
+            [
+                'label'          => t('Delete'),
+                'form'           => 'event-rule-config-form',
+                'class'          => 'btn-remove',
+                'formnovalidate' => true
+            ]
+        ));
+
+        $buttonsWrapper->add(
```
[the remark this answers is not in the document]
raviks789: This is a good suggestion. I will try making these changes in the PR #170. As certain changes would become obsolete after moving the buttons.

raviks789:
> I think you can move all `edit/add` and `search-editor` related methods from `EventRulesController` to `EventRuleController`. Since the `EventRulesController` should only show the `add` button and the list items.

This changes will be made in PR #170. As certain changes made while moving the `EventRulesController::addAction` to `EventRuleController` will become obsolete in PR #170.

application/forms/EventRuleConfigElements/EscalationRecipient.php
```diff
@@ -0,0 +1,241 @@
...
+                'title'          => $this->translate('Add Recipient'),
+                'formnovalidate' => true
+            ]
+        );
+
+        $this->registerElement($addRecipientButton);
+        $recipientCount = $this->getValue('recipient-count');
+        if ($addRecipientButton->hasBeenPressed()) {
+            $recipientCount += 1;
+            $this->getElement('recipient-count')->setValue($recipientCount);
```
sukhwinder33445, line 42:
```suggestion
            $this->getElement('recipient-count')->setValue(++$recipientCount);
```

application/forms/EventRuleConfigElements/EscalationRecipient.php
```diff
@@ -0,0 +1,241 @@
...
+     * @param int $pos
+     *
+     * @return ?SubmitButtonElement
+     */
+    protected function createRemoveButton(int $pos): ?SubmitButtonElement
+    {
+        $recipientCount = $this->getValue('recipient-count');
+        if ((int) $recipientCount === 1) {
+            return null;
+        }
```
sukhwinder33445, line 175: Please call `createRemoveButton` method conditionally instead.

sukhwinder33445 (changes requested):

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -0,0 +1,283 @@
...
+    public $allowZeroConditions;
+
+    /**
+     * Set whether the zero conditions is allowed for the escalation
+     *
+     * @param bool $allowZeroConditions
+     *
+     * @return $this
+     */
+    public function setAllowZeroConditions(bool $allowZeroConditions = false): self
```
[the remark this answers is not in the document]
raviks789: This is because by default, the escalations should not have zero condition, except for that one escalation with zero conditions.

application/controllers/EventRulesController.php
```diff
@@ -110,19 +108,50 @@ public function addAction(): void
...
+            ->populate($config);
+
+        $eventRuleConfig
+            ->on(Form::ON_SUCCESS, function (EventRuleConfigForm $form) use ($config) {
+                $ruleId = $config['id'];
+                $ruleName = $config['name'];
+                $form->addOrUpdateRule($ruleId, $config);
+                $this->sessionNamespace->delete($ruleId);
+                Notification::success(sprintf(t('Successfully add event rule %s'), $ruleName));
+                $this->redirectNow('__CLOSE__');
```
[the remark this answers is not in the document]
raviks789: This was done to match the behavior with other configuration, we have like Contacts. But this seem to be unnecessary.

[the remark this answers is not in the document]
raviks789: Will undo the changes.

application/controllers/EventRulesController.php
```diff
@@ -110,19 +108,50 @@ public function addAction(): void
...
+            })
+            ->on(EventRuleConfigForm::ON_CHANGE, function (EventRuleConfigForm $form) use ($config) {
+                $config = array_merge($config, $form->getValues());
+                $config['rule_escalation'] = $form->getValues()['rule_escalation'];
```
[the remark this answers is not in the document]
raviks789: I wanted to merge while holding the values for escalations from the form and the order of keys in the array. `array_merge` will retain the values from the first array if the key already is present. I could reverse the order of merge but I didn't want to lose the order of the keys.

application/controllers/EventRuleController.php
```diff
@@ -32,81 +34,119 @@ class EventRuleController extends CompatController
...
+                    sprintf(
+                        t('Successfully discarded changes to event rule %s'),
+                        $configValues['name']
+                    )
+                );
+                $this->redirectNow(Links::eventRule((int) $ruleId));
+            })
+            ->on(EventRuleConfigForm::ON_CHANGE, function (EventRuleConfigForm $form) use ($ruleId, $configValues) {
+                $configValues = array_merge($configValues, $form->getValues());
+                $configValues['rule_escalation'] = $form->getValues()['rule_escalation'];
```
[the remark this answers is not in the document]
raviks789: It is not superfluous, as I have mentioned above the reason I do this.

application/controllers/EventRuleController.php
```diff
@@ -221,52 +260,70 @@ public function searchEditorAction(): void
...
-        if (isset($cache) || $ruleId === '-1') {
-            $config = $cache ?? [];
-        } else {
-            $config = $this->fromDb((int) $ruleId);
+        $config = $this->sessionNamespace->get($ruleId);
+        if ($config === null) {
+            if ($ruleId === '-1') {
+                $config = ['id' => $ruleId];
+            } else {
+                $config = $this->fromDb((int) $ruleId);
```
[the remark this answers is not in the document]
raviks789: I would say, you still need this, while you store the config in session. But you are right, we should query the columns directly. However, this must be fixed in PR #170 as there we do not need the session storage.

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -0,0 +1,133 @@
...
+                'class'          => ['add-button', 'control-button', 'spinner'],
+                'label'          => new Icon('plus'),
+                'formnovalidate' => true,
+                'title'          => $this->translate('Add filter')
+            ]
+        );
+
+        $this->registerElement($addFilterButton);
+        /** @var string $showSearchBar */
+        $showSearchBar = $this->getValue('show-searchbar');
```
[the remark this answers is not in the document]
raviks789: Now I look at it, I don't think the element 'show-searchbar' is necessary.

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
```diff
@@ -0,0 +1,44 @@
...
+    protected $tag = 'li';
+
+    /** @var ?SubmitButtonElement Remove button for the recipient */
+    public $removeButton;
+
+    /** @var FormElement Recipient name */
+    public $recipient;
+
+    /** @var FormElement Recipient channel */
+    public $channel;
```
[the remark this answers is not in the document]
raviks789: While the the attributes of form elements are replaced in `EscalationRecipient`.

application/controllers/EventRulesController.php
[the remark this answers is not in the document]
sukhwinder33445: You are calling `$form->getValues()` twice, which is unnecessary (db fetch, loop and all). However, this is moved to `EventRuleController` in https://github.com/Icinga/icinga-notifications-web/pull/170 and there `getValues()` is not called twice, so it's fine.

library/Notifications/Widget/ItemList/Escalation.php
```diff
@@ -0,0 +1,94 @@
...
+    protected $tag = 'li';
+
+    /** @var ?SubmitButtonElement Remove button of the escalation */
+    public $removeButton;
+
+    /** @var EscalationCondition Escalation condition fieldset */
+    public $condition;
+
+    /** @var EscalationRecipient Escalation recipient fieldset */
+    public $recipient;
```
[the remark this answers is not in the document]
raviks789: Again, same reason as I mentioned for recipients and conditions.

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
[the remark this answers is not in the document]
sukhwinder33445: Please use `getter/setter` for it.

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -0,0 +1,133 @@
...
+    /** @var Url Url of the search editor */
+    protected $searchEditorUrl;
+
+    /** @var ?string Event rule's object filter */
+    protected $objectFilter;
+
+    protected $defaultAttributes = ['class' => 'config-filter'];
+
+    protected function assemble(): void
+    {
```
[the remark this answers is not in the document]
raviks789: I think we do not need 'show-searchbar' element anymore and I have already rewritten the assemble method.

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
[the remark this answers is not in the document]
raviks789: After discussing with @nilmerg, we have come to a conclusion, that this has to be designed in a different way by taking advantage of `registerAttributeCallback` and deferred assembling of the list. Same for conditions and escalations.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * Get the rendered condition
+     *
+     * @return string
+     */
+    public function getCondition(): string
+    {
+        $filter = Filter::any();
+        /** @var int $count */
+        $count = $this->getValue('condition-count');
```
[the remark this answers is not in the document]
raviks789: Casting to int is required while after autosubmit the values that are populated from the post request and this would be a string after the autosubmit, but as you said casting back to string is redundant.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -166,7 +161,9 @@ protected function assemble(): void
             $this->registerElement($col);
             $this->registerElement($op);
             $this->registerElement($val);
-            $removeButton = $this->createRemoveButton($i);
+            $removeButton = $conditionCount === 1 && ! $this->allowZeroConditions
```
sukhwinder33445, line 164: Please use `if` condition to keep it simple.
```suggestion
if (..) {
                $removeButton = $this->createRemoveButton($i);
                
                if ($removeButton->hasBeenPressed()) {
                    $removePosition = $i;
                }
            }
```

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -254,16 +245,15 @@ public function hasValue(): bool
...
-        /** @var int $count */
-        $count = $this->getValue('condition-count');
-        $removePosition = $this->getValue('remove');
+        $count = (int) $this->getValue('condition-count');
+        $removePosition = (int) $this->getValue('remove');
         if ($removePosition) {
             $count += 1;
         }
 
         if ($count > 0) { // if count is 0, loop runs in reverse direction
```
sukhwinder33445, line 254: You can return earlier in this case.
```suggestion
public function getCondition(): string
    {
        $count = (int) $this->getValue('condition-count');
        if ($count === 0) {
            return '';
        }
        
        $filter = Filter::any();
        ...
```

application/forms/EventRuleConfigElements/EscalationRecipient.php
```diff
@@ -100,7 +99,7 @@ protected function assemble(): void
             }
 
             $this->registerElement($val);
-            $removeButton = $this->createRemoveButton($i);
+            $removeButton = $recipientCount > 1 ? $this->createRemoveButton($i) : null;
```
sukhwinder33445, line 102: Please use `if` condition.

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -22,103 +21,63 @@ class EventRuleConfigFilter extends FieldsetElement
...
-            );
-
-            $filterElement->add([$searchBar, $editorOpener]);
+            $this->registerElement($addFilterButton);
+            if ($addFilterButton->hasBeenPressed()) {
+                $this->removeAttribute('class', 'empty-filter');
+            } else {
+                $this->addAttributes(['class' => 'empty-filter']);
+                $this->addHtml($addFilterButton);
+                return;
```
sukhwinder33445, line 49: Please add a new line before `return` statement.

application/forms/EventRuleConfigElements/EventRuleConfigFilter.php
```diff
@@ -22,103 +21,63 @@ class EventRuleConfigFilter extends FieldsetElement
...
             );
-
-            $filterElement = new HtmlElement(
-                'div',
-                Attributes::create(['class' => ['search-controls', 'icinga-controls']])
-            );
-
-            $filterElement->add([$searchBar, $editorOpener]);
+            $this->registerElement($addFilterButton);
+            if ($addFilterButton->hasBeenPressed()) {
```
sukhwinder33445, line 44: A new line before `if`please.

sukhwinder33445 (changes requested):

application/controllers/EventRuleController.php
```diff
@@ -283,7 +281,7 @@ public static function createFilterString(Filter\Rule $filters): string
 
         $filterStr = QueryString::render($filters);
 
-        return ! empty($filterStr) ? $filterStr : '';
+        return ! empty($filterStr) ? rawurldecode($filterStr) : '';
```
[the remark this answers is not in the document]
raviks789: This is to store and display certain special characters like '/' as it is instead of in its url encoded form. This was earlier done in `setObjectString` method in `EventRuleObjectFilter`.

application/controllers/EventRuleController.php
```diff
@@ -131,9 +131,7 @@ public function indexAction(): void
             ]
         );
 
-        $buttonsWrapper->add(
-            [$eventRuleConfigSubmitButton, $discardChangesButton, $deleteButton]
-        );
+        $buttonsWrapper->add([$eventRuleConfigSubmitButton, $discardChangesButton, $deleteButton]);
```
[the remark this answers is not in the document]
raviks789: I would use `add()` here as `$discardChangesButton` could be null. But `addHtml` could be use in PR #170, as you do not need that button.

application/controllers/EventRuleController.php
```diff
@@ -295,7 +293,12 @@ public function editAction(): void
             if ($ruleId === '-1') {
                 $config = ['id' => $ruleId];
             } else {
-                $config = $this->fromDb((int) $ruleId);
+                $config = iterator_to_array(
+                    Rule::on(Database::get())
+                        ->withColumns(['id'])
+                        ->filter(Filter::equal('id', $ruleId))
+                        ->first()
+                );
```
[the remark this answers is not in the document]
raviks789: You do not need to mention the columns, as by default only the columns from `rule` table will be fetched.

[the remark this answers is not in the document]
sukhwinder33445: Yes, but you don't want all rule columns.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+        $formValues = [];
+        $formValues['prefixes-map'] = $this->getPrefixesMap(count($values['rule_escalation']));
+
+        foreach ($values['rule_escalation'] as $position => $escalation) {
+            $conditions = explode('|', $escalation['condition'] ?? '');
+            $conditionFormValues = [];
+            $conditionFormValues['condition-count'] = count($conditions);
+            $conditionFormValues['id'] = $escalation['id'] ?? bin2hex(random_bytes(4));
+            foreach ($conditions as $key => $condition) {
+                if ($condition === '' && ! isset($formValues['zero-condition-escalation'])) {
```
sukhwinder33445, line 279: Please use `empty()` instead.

sukhwinder33445 (changes requested):
The delete button doesn't show any hint why it is deactivated. I can't find the previous comment, but I've requested this change before also.

raviks789:
> The delete button doesn't show any hint why it is deactivated. I can't find the previous comment, but I've requested this change before also.

This comment is present in PR#170. You are right, it must be fixed here.

application/controllers/EventRuleController.php
```diff
@@ -32,81 +34,117 @@ class EventRuleController extends CompatController
     /** @var Session\SessionNamespace */
     private $sessionNamespace;
 
+    /** @var ?string Event rule config filter */
+    protected $filter;
```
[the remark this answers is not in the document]
raviks789: I think after code changes, this property has no usages and must be removed.

application/controllers/EventRuleController.php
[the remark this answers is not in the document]
raviks789: I don't understand why is this an issue here. As I have mentioned in my earlier comment, I would instead fix this in PR #170.

application/controllers/EventRuleController.php
```diff
@@ -221,52 +256,71 @@ public function searchEditorAction(): void
...
+            }
         }
 
         $eventRuleForm = (new EventRuleForm())
             ->populate($config)
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUCCESS, function ($form) use ($ruleId, $cache, $config) {
+            ->on(Form::ON_SUCCESS, function ($form) use ($ruleId, $config) {
                 $config['name'] = $form->getValue('name');
                 $config['is_active'] = $form->getValue('is_active');
```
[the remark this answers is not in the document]
raviks789: as I am just replacing the values for keys `name` and `is_active` and rest of the config should remain how it was.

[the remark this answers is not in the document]
sukhwinder33445: I see.

application/controllers/EventRuleController.php
```diff
@@ -32,81 +34,117 @@ class EventRuleController extends CompatController
...
+
     public function init()
     {
         $this->sessionNamespace = Session::getSession()->getNamespace('notifications');
     }
 
     public function indexAction(): void
     {
-        $this->assertPermission('notifications/config/event-rules');
+        $this->assertPermission('notifications/config/event-rule');
```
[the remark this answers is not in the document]
sukhwinder33445: Why not? If it is the same condition or base condition for each action, it can be moved to `init`.

application/controllers/EventRuleController.php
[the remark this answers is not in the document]
sukhwinder33445: My only thought was that if the code is introduced here, it can also be changed here.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+        $escalations = $this->config['rule_escalation'] ?? [];
+
+        $pos = hex2bin($prefix);
+        $disableRemoveButton = false;
+        if ($pos) {
+            $escalationId = $escalations[$pos]['id'] ?? null;
+            $incident = Incident::on(Database::get())
+                ->with('rule_escalation');
+
+            if ($escalationId && ctype_digit($escalationId)) {
```
sukhwinder33445, line 388: Why `ctype_digit` check? Is casting to `int` not sufficient? It looks kind of weird.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+     * @return SubmitButtonElement
+     */
+    protected function createRemoveButton(string $prefix): SubmitButtonElement
+    {
+        /** @var array<int, array<string, mixed>> $escalations */
+        $escalations = $this->config['rule_escalation'] ?? [];
+
+        $pos = hex2bin($prefix);
+        $disableRemoveButton = false;
+        if ($pos) {
```
sukhwinder33445, line 383: Why the `$pos` can be false? Is the `$prefix` known at this point or not?

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+        if ($disableRemoveButton) {
+            $button->addAttributes([
+                'disabled' => true,
+                'title'    => $this->translate(
+                    'There exist active incidents for this escalation and hence cannot be removed'
+                )
+            ]);
+        } else {
+            $button->addAttributes(['title' => $this->translate('Remove escalation')]);
+        }
```
sukhwinder33445, line 415:
```suggestion
        $button = new SubmitButtonElement(
            'remove-escalation',
            [
                'class'             => ['remove-escalation', 'remove-button', 'control-button', 'spinner'],
                'label'             => new Icon('minus'),
                'formnovalidate'    => true,
                'value'             => $prefix,
                'disabled'          => $disableRemoveButton,
                'title'             => $disableRemoveButton
                    ? $this->translate('There exist active incidents for this escalation and hence cannot be removed')
                    : $this->translate('Remove escalation')
            ]
        );
```

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+     *
+     * @param int $escalationCount
+     *
+     * @return string
+     */
+    protected function getPrefixesMap(int $escalationCount): string
+    {
+        $prefixesMap = [];
+        for ($i = 1; $i <= $escalationCount; $i++) {
+            $prefixesMap[(string) $i] = bin2hex((string) $i);
```
sukhwinder33445, line 649: Why can't the `key` be omitted?

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+    /**
+     * Create a new EventRuleConfigForm
+     *
+     * @param array<string, mixed> $config
+     * @param Url                  $searchEditorUrl
+     */
+    public function __construct(array $config, Url $searchEditorUrl)
+    {
+        $this->config = $config;
+        $this->config['object_filter'] = $this->config['object_filter'] ?? null;
```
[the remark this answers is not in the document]
raviks789: I get where you are coming from. But I would say this is an implementation detail and I would call it out as not correct. But changes will be made.

application/controllers/EventRulesController.php
```diff
@@ -110,19 +107,46 @@ public function addAction(): void
...
+                $this->sessionNamespace->set('-1', $config);
+            })
+            ->handleRequest($this->getServerRequest());
 
         $eventRuleForm = Html::tag('div', ['class' => 'event-rule-form'], [
-            Html::tag('h2', $eventRuleConfig->getConfig()['name'] ?? ''),
+            Html::tag('h2', $config['name'] ?? ''),
```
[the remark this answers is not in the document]
raviks789: The config will be empty on a hard refresh for new event rule.

application/forms/EventRuleConfigForm.php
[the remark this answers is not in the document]
sukhwinder33445: In my opinion, cuntructor should not manipulate the values it stores. It should store them as they are.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -0,0 +1,267 @@
...
+        $defaultCount = 1;
+        $configHasZeroConditionEscalation = $this->configForm->hasZeroConditionEscalation();
+        if ($zeroConditions && $configHasZeroConditionEscalation) {
+            $defaultCount = 0;
+            $conditionCount = $defaultCount;
+        } else {
+            $conditionCount = $conditionCount === null ? $defaultCount : (int) $conditionCount;
+        }
+
+        $this->addElement('hidden', 'id');
```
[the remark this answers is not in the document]
raviks789: This is needed to populate escalation_ids and save the recipient and conditions to correct escalation

application/forms/EventRuleConfigForm.php
raviks789: nope, because I am assigning fake escalation Id  for new escalations which a is not integer but hexadecimal.

application/forms/EventRuleConfigForm.php
raviks789: It is required to populate and remove right escalations.

application/forms/EventRuleConfigForm.php
raviks789: I think that a strict check for empty string is completely alright.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,654 @@
...
+     */
+    public function hasZeroConditionEscalation(): bool
+    {
+        return $this->hasZeroConditionEscalation;
+    }
+
+    protected function assemble(): void
+    {
+        $this->addElement($this->createCsrfCounterMeasure(Session::getSession()->getId()));
+        $this->add($this->createUidElement());
```
sukhwinder33445, line 110: This is now superfluous, please remove.

application/controllers/EventRuleController.php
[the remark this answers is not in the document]
raviks789:
```php
$buttonsWrapper->addHtml($eventRuleConfigSubmitButton, $deleteButton);
```

The change in PR #170 would look something like which is only one line change. And the if condition I introduce will become obsolete. So it is better to handle this in PR #170 than here.

application/controllers/EventRuleController.php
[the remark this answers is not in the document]
raviks789: Just an update this change is now done in PR #170. In this PR you require the entire config as you save the config to the session.

application/forms/EventRuleConfigForm.php
raviks789: No, probably the if condition has become superfluous after the code changes.

application/forms/EventRuleConfigForm.php
```diff
@@ -273,8 +269,10 @@ public function populate($values): self
         foreach ($values['rule_escalation'] as $position => $escalation) {
             $conditions = explode('|', $escalation['condition'] ?? '');
             $conditionFormValues = [];
-            $conditionFormValues['condition-count'] = count($conditions);
+            $conditionCount = count($conditions);
+            $conditionFormValues['condition-count'] = $conditionCount;
```
sukhwinder33445, line 274: Please remove the `$conditionCount` variable.

application/controllers/EventRuleController.php
```diff
@@ -139,12 +139,15 @@ public function indexAction(): void
                 ->filter(Filter::equal('rule.id', $ruleId));
 
             if ($incidents->count() > 0) {
-                $deleteButton->addAttributes(['disabled' => true]);
+                $deleteButton->addAttributes([
+                    'disabled' => true,
```
sukhwinder33445: I would chain the `count()`, as we discussed offline.

application/forms/EventRuleConfigForm.php
```diff
@@ -377,19 +375,15 @@ protected function createRemoveButton(string $prefix): SubmitButtonElement
...
-                $incident->filter(Filter::equal('rule_escalation.id', $escalationId));
-                if ($incident->count() > 0) {
-                    $disableRemoveButton = true;
-                }
+        $escalationId = $escalations[$pos]['id'] ?? null;
+
+        if ($escalationId && ctype_digit($escalationId)) {
+            $incident = Incident::on(Database::get())->with('rule_escalation');
+            $incident->filter(Filter::equal('rule_escalation.id', $escalationId));
+            if ($incident->count() > 0) {
```
sukhwinder33445, line 386: I would chain the `count()`, as we discussed offline.

application/forms/EventRuleConfigForm.php
sukhwinder33445: I can't find a use for it. Can you please show me the lines of code.

sukhwinder33445 (dismissed):
Looks good to me, works fine.

@nilmerg please have a look.

sukhwinder33445:
Please squash the commits.

sukhwinder33445:
Please do not forget to keep an eye on the phpstan errors.

sukhwinder33445 (dismissed):
LGTM

library/Notifications/Model/Contact.php
```diff
@@ -7,10 +7,24 @@
...
 use ipl\Orm\Behaviors;
 use ipl\Orm\Model;
+use ipl\Orm\Query;
 use ipl\Orm\Relations;
 
 /**
  * @property int $id
+ * @property string $full_name
+ * @property ?string $username
+ * @property string $color
```
sukhwinder33445, line 17: Remove please.

library/Notifications/Common/Links.php
```diff
@@ -110,4 +110,12 @@ public static function moveRotation(): Url
     {
         return Url::fromPath('notifications/schedule/move-rotation');
     }
+
+    public static function ruleFilterSuggestionUrl(int $id): Url
+    {
+        return Url::fromPath(
+            "notifications/event-rule/complete",
+            ['_disableLayout' => true, 'showCompact' => true, 'id' => $id]
```
sukhwinder33445, line 118: Please do not add any model-related parameters here. This class should only provide basic links. I already had a discussion with Johannes about this topic. You can add more parameters while calling this function, or remove them altogether and create them in a specific class if required.

library/Notifications/Model/Contact.php
```diff
@@ -7,10 +7,24 @@
...
+ * @property int $default_channel_id
+ *
+ * @property Channel | Query $channel
+ * @property ContactAddress | Query $contact_address
+ * @property Incident | Query $incident
+ * @property IncidentContact | Query $incident_contact
+ * @property IncidentHistory | Query $incident_history
+ * @property RuleEscalationRecipient | Query $rule_escalation_recipient
+ * @property RotationMember | Query $rotation_member
+ * @property ContactAddress | Query $contactgroup
```
sukhwinder33445, line 27: This is optional, Most of the existing and upcoming annotations are defined as following:
`@property Query|Channel $channel`. (without a space between `|`)
This way the code looks consistent.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,647 @@
...
+    {
+        if (! isset($values['rule_escalation'])) {
+            return parent::populate($values);
+        }
+
+        $formValues = [];
+        $formValues['prefixes-map'] = $this->getPrefixesMap(count($values['rule_escalation']));
+
+        foreach ($values['rule_escalation'] as $position => $escalation) {
+            $conditions = explode('|', $escalation['condition'] ?? '');
```
nilmerg, line 271: It's odd to me that these are parsed here. For one, in `getValues()` they aren't built, so that must be done somewhere else, so why are they parsed here? And second, they're parsed *by hand*, meaning, not by `QueryString::parse`. I hope this is just a mistake here, and where ever they're built, this is not also done *by hand*.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,647 @@
...
+
+            foreach ($conditions as $key => $condition) {
+                if ($condition === '' && ! isset($formValues['zero-condition-escalation'])) {
+                    $formValues['zero-condition-escalation'] = bin2hex($position);
+                    $conditionFormValues['condition-count'] = 0;
+
+                    continue;
+                }
+
+                $count = $key + 1;
```
nilmerg, line 284: $position, no?

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,647 @@
...
+                                $recipientFormValues['val_' . $count] = $elementValue;
+                            }
+                        }
+
+                        if (isset($recipient['id'])) {
+                            $recipientFormValues['id_' . $count] = (int) $recipient['id'];
+                        }
+                    }
+                }
+            }
```
nilmerg, line 326: Now I'm sure something is off here. Much of this is split here and in `EscalationCondition` as well as `EscalationRecipient`.

It's bad that processing of input and output is split that way. There should be only one location where this is done, and these are the iframes to me. They accept and provide form data as well as config. They know best what structure is needed where. No need to spread so much *intimate* knowledge around.

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,647 @@
...
+     */
+    protected function createRemoveButton(string $prefix): SubmitButtonElement
+    {
+        /** @var array<int, array<string, mixed>> $escalations */
+        $escalations = $this->config['rule_escalation'] ?? [];
+        $pos = hex2bin($prefix);
+        $disableRemoveButton = false;
+        $escalationId = $escalations[$pos]['id'] ?? null;
+
+        if ($escalationId && ctype_digit($escalationId)) {
```
nilmerg, line 382: What else could these be than digits?

application/forms/EventRuleConfigForm.php
```diff
@@ -0,0 +1,647 @@
...
+            $incidentCount = Incident::on(Database::get())
+                ->with('rule_escalation')
+                ->filter(Filter::equal('rule_escalation.id', $escalationId))
+                ->count();
+
+            $disableRemoveButton = $incidentCount > 0;
+        }
+
+        $button = new SubmitButtonElement(
```
nilmerg, line 391: Again, please use `$this->createElement`

nilmerg (changes requested):
Not finished yet. Will continue tomorrow.

library/Notifications/Widget/ItemList/EscalationConditionList.php
```diff
@@ -0,0 +1,49 @@
...
+
+/* Icinga Notifications Web | (c) 2023 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Notifications\Widget\ItemList;
+
+use ipl\Html\BaseHtmlElement;
+
+class EscalationConditionList extends BaseHtmlElement
+{
+    protected $defaultAttributes = ['class' => 'options'];
```
nilmerg, line 11: I'm not a fan of such generic class names. The problem with these is sometimes that they get undesired styles applied due to some random rule defined somewhere else. So please use a less generic name. Such as `escalation-condition-list` :wink:

library/Notifications/Widget/ItemList/EscalationConditionList.php
```diff
@@ -0,0 +1,49 @@
...
+
+    protected $tag = 'ul';
+
+    /** @var EscalationConditionListItem[] Condition list items */
+    protected $conditions;
+
+    /**
+     * Create conditions list of the escalation
```
nilmerg, line 19: *Create a list of escalation conditions*

library/Notifications/Widget/ItemList/EscalationConditionList.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $removedPosition = null;
+        foreach ($this->conditions as $position => $condition) {
+            if ($condition->hasBeenRemoved()) {
+                $removedPosition = $position;
+
+                continue;
+            }
+
+            if ($removedPosition) {
```
nilmerg, line 38: Just for a litte bit of safety: (I know positions start at 1)

```suggestion
            if ($removedPosition !== null) {
```

library/Notifications/Widget/ItemList/EscalationConditionList.php
```diff
@@ -0,0 +1,49 @@
...
+                $condition->setPosition($position - 1);
+            }
+        }
+
+        foreach ($this->conditions as $position => $condition) {
```
nilmerg, line 43: why two loops? One should suffice.

library/Notifications/Widget/ItemList/EscalationRecipientList.php
```diff
@@ -0,0 +1,55 @@
...
+
+/* Icinga Notifications Web | (c) 2023 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Notifications\Widget\ItemList;
+
+use ipl\Html\BaseHtmlElement;
+
+class EscalationRecipientList extends BaseHtmlElement
+{
+    protected $defaultAttributes = ['class' => 'options'];
```
nilmerg, line 11: `escalation-recipient-list`

library/Notifications/Widget/ItemList/EscalationRecipientList.php
```diff
@@ -0,0 +1,55 @@
...
+
+    protected $tag = 'ul';
+
+    /** @var EscalationRecipientListItem[] Recipient list items of the escalation */
+    protected $recipients;
+
+    /**
+     * Create recipients list of the escalation
```
nilmerg, line 19: *Create a list of escalation recipients*

library/Notifications/Widget/ItemList/EscalationRecipientList.php
```diff
@@ -0,0 +1,55 @@
...
+        $recipientCount = count($this->recipients);
+        foreach ($this->recipients as $position => $recipient) {
+            if ($recipient->hasBeenRemoved()) {
+                $removedPosition = $position;
+                --$recipientCount;
+
+                continue;
+            }
+
+            if ($removedPosition) {
```
nilmerg, line 40: !== null

library/Notifications/Widget/ItemList/EscalationRecipientList.php
```diff
@@ -0,0 +1,55 @@
...
+                $recipient->setPosition($position - 1);
+            }
+        }
+
+        foreach ($this->recipients as $position => $recipient) {
```
nilmerg, line 45: a single loop only please

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
```diff
@@ -0,0 +1,97 @@
...
+    protected $operator;
+
+    /** @var FormElement Condition value */
+    protected $conditionVal;
+
+    /** @var int */
+    protected $position;
+
+    /**
+     * Create the condition list item of the escalation
```
nilmerg, line 33: missing doc for $position

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Icinga\Module\Notifications\Widget\ItemList;
+
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\Contract\FormElement;
+use ipl\Html\FormElement\SubmitButtonElement;
+
+class EscalationConditionListItem extends BaseHtmlElement
+{
+    protected $defaultAttributes = ['class' => 'option'];
```
nilmerg, line 13: `escalation-condition-list-item`

or

`escalation-condition-list > li` (in css)

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
```diff
@@ -0,0 +1,97 @@
...
+     * @return $this
+     */
+    public function setPosition(int $position): self
+    {
+        $this->position = $position;
+
+        return $this;
+    }
+
+    public function setRemoveButton(?SubmitButtonElement $removeButton): self
```
nilmerg, line 78: missing doc

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Icinga\Module\Notifications\Widget\ItemList;
+
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\Contract\FormElement;
+use ipl\Html\FormElement\SubmitButtonElement;
+
+class EscalationRecipientListItem extends BaseHtmlElement
+{
+    protected $defaultAttributes = ['class' => 'option'];
```
nilmerg, line 13: you know the drill

library/Notifications/Widget/ItemList/EscalationRecipientListItem.php
```diff
@@ -0,0 +1,92 @@
...
+     * @return $this
+     */
+    public function setPosition(int $position): self
+    {
+        $this->position = $position;
+
+        return $this;
+    }
+
+    public function removeRemoveButton(): self
```
nilmerg, line 73: missing doc

library/Notifications/Widget/ItemList/Escalation.php
```diff
@@ -0,0 +1,97 @@
...
+    protected $removeButton;
+
+    /** @var EscalationCondition Escalation condition fieldset */
+    protected $condition;
+
+    /** @var EscalationRecipient Escalation recipient fieldset */
+    protected $recipient;
+
+    /** @var bool Whether the widget has a remove button */
+    protected $hasNoRemoveButton = false;
```
nilmerg, line 29: This is always false, if I'm not mistaken, and as such unused.

library/Notifications/Widget/ItemList/Escalation.php
```diff
@@ -0,0 +1,97 @@
...
+    public function lastConditionHasBeenRemoved(): bool
+    {
+        return $this->condition->getPopulatedValue('condition-count') === '1'
+            && $this->condition->getPopulatedValue('remove') === '1';
+    }
+
+    /**
+     * Create first component of the escalation widget
+     *
+     * @return ?FlowLine|SubmitButtonElement
```
nilmerg, line 72: this cannot return null

nilmerg (changes requested):
next: controllers, fieldsets, css

nilmerg (changes requested):
next: controllers, css

application/controllers/EventRulesController.php
```diff
@@ -45,6 +45,7 @@ public function init()
     public function indexAction(): void
     {
         $eventRules = Rule::on(Database::get());
+        $this->sessionNamespace->delete('-1');
```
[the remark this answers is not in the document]
nilmerg: Note that, as I just saw that the `New Event Rule` button used to have a parameter with such functionality, you should use a separate action for new event rules. This then should clear the session before writing anything to it upon submit.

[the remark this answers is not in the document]
nilmerg: /And now I noticed that this very action is already the `addAction` here, but that's used to set up a new rule. i.e. It's not just the endpoint for the very first modal, it's a detail.

I think that there doesn't need to be a distinction between `event-rule?id=-1` and `event-rule?id!=-1`. The `event-rule/index` route should be able to handle new rules (=-1) and existing ones (!=-1). There's not much difference between them, only a few buttons and event handlers. By transferring the responsibility to create buttons to the form, this is already not part of the action anymore. So that only leaves the events and those are just not emitted in case of a new rule, so it doesn't really hurt that they are defined.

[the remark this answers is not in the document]
nilmerg: So in the end:

`event-rules/add` is the endpoint for the the `New Event Rule` button, this resets the session upon submit
`event-rule/index` is the endpoint to configure new and existing rules
`event-rule/edit` is the endpoint to edit new and existing rules

nilmerg (changes requested):
Finito. For now.

I didn't look at CSS, will do this at the very end.

Since the controllers will receive a rather large rework, I didn't look that detailed over them as in the previous reviews. Will do this once they're adjusted.

sukhwinder33445:
Removing an escalation condition writes `''` (empty string) (rule_escalation.condition='') instead of `null` to the database.

application/forms/EventRuleConfigForm.php
raviks789: This was used because I am assigning hexadecimal values as fake IDs for escalations that are not yet saved. In this case using `is_numeric` to avoid the call to database does not work as that would be true even if the IDs are hexadecimal.

application/controllers/EventRuleController.php
```diff
@@ -28,146 +30,127 @@ class EventRuleController extends CompatController
...
-            }
-
-            $escalationRecipients = $re
-                ->rule_escalation_recipient
-                ->withoutColumns(['changed_at', 'deleted']);
-
-            foreach ($escalationRecipients as $recipient) {
-                $config[$re->getTableName()][$re->position]['recipient'][] = iterator_to_array($recipient);
-            }
+        $this->addControl(Html::tag('div', ['class' => 'event-rule-and-save-forms'], [
```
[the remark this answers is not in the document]
nilmerg: If that'd be the case, then why keep `Html` altogether? Remember, strictness may be preferable, but in case of code that is not repeated nor part of a public interface (not a library part), laziness wins.

application/forms/EventRuleConfigElements/NotificationConfigProvider.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function fetchRuleEscalations(array $columns, Rule $filter): iterable
+    {
+        return RuleEscalation::on(Database::get())->columns($columns)->filter($filter);
+    }
+
+    public function fetchRuleEscalationRecipients(array $columns, Rule $filter): iterable
+    {
+        return RuleEscalationRecipient::on(Database::get())->columns($columns)->filter($filter);
+    }
```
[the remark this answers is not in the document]
nilmerg: Why should it be required?, should be the question.

phpstan.neon
```diff
@@ -8,6 +8,7 @@ parameters:
     checkFunctionNameCase: true
     checkInternalClassCaseSensitivity: true
     treatPhpDocTypesAsCertain: false
+    reportUnmatchedIgnoredErrors: false
```
sukhwinder33445, line 11: We should remove this later once this PR is ready to merge and clean up the baseline instead.

public/css/detail/event-rule-detail.less
```diff
@@ -1,432 +1,282 @@
...
-          background: none;
-          cursor: not-allowed;
-          opacity: 0.5;
-        }
-      }
-
-      &.btn-discard-changes {
-        .event-rule-button();
-      }
+    .btn-discard-changes {
```
sukhwinder33445, line 274: This is now obsolete.

public/css/detail/event-rule-detail.less
```diff
@@ -1,432 +1,282 @@
...
-}
+// Other stuff
 
-.horizontal-line-long {
-  width: 14.5em;
+.event-rule-detail {
+  > #cache-notice:not(:empty) {
+    margin-bottom: 1em;
+  }
 }
```
sukhwinder33445, line 252: This is now obsolete.

sukhwinder33445 (changes requested):

sukhwinder33445:
Class `RightArrow` is not used anywhere and can be removed.

The class `FlowLine` can also be removed, as the CSS classes added by this class apply exactly the same rules. The getVerticalLine() method is not used. Instead of applying class `right-arrow` or `horizontal-line`, you can add class line `connector` or `connector-line`. Instead of the `right-arrow` and `horizontal-line` class, you can add the something like `connector` or `connector-line` class.

application/forms/EventRuleConfigElements/EscalationRecipient.php
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Create a list of recipients to use in a select element
+     *
+     * @return array<string, array<string, string>>
+     */
+    protected function selectRecipients(): array
+    {
+        $contacts = [];
+        foreach ($this->provider?->fetchContacts() ?? [] as $contact) {
```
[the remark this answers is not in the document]
nilmerg: This is actually only for static analysis as the property can be null but technically isn't. So I don't see any reason to do more than use a syntactic element introduced with PHP 8 that allows to easily silence such inspections ;)

sukhwinder33445:
> Fixed. Though, now the shown filter shows escaped chars again. I'm not sure whether it's necessary to avoid this here. The object filters will change anyway and when they do, I doubt that such a simple text input will remain. So let's keep that for now.

Now the database entry has the operator as a suffix, is that OK?

sukhwinder33445:
@oxzi, that should be fine as discussed offline, right?

oxzi:
> In case the daemon accepts it, sure :P

Well, based on the architectural change described in the "Custom variable support in Notifications" document, the event rules will not be evaluated in the Icinga Notifications daemon anymore, but will be moved to the source.

Implementation started in https://github.com/Icinga/icinga-notifications/pull/324 and https://github.com/Icinga/icingadb/pull/998.

Thus, I am a bit puzzled how this PR works with the upcoming changes, especially the intermediate representation mentioned in the docs.

> @oxzi, that should be fine as discussed offline, right?

I was unable to review +2,762 −2,829 changes in 34 files, but at least having a suffix is something which works for us :upside_down_face:

application/controllers/EventRuleController.php
```diff
@@ -186,7 +186,11 @@ public function searchEditorAction(): void
             );
 
         $editor->on(Form::ON_SUCCESS, function (SearchEditor $form) use ($ruleId) {
-            $this->session->set('object_filter', (new Renderer($form->getFilter()))->render());
+            $filter = (new Renderer($form->getFilter()))->render();
+            // TODO: Should not be needed for the new filter implementation
+            $filter = preg_replace('/[<>!=]{1,2}(?=[|&]|$)/', '', $filter);
```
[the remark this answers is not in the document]
nilmerg: fixed

PR #155, opened 2024-01-12
Fix inefficient list views which render object names

nilmerg:
fixes #153

requires https://github.com/Icinga/ipl-orm/pull/128, https://github.com/Icinga/ipl-orm/pull/126 and https://github.com/Icinga/ipl-orm/pull/124

PR #165, opened 2024-03-21
Hook implementation for event object display names

### LINKED ISSUE #132: Implement hooks to render objects
With the new changes in notifications, the object table will no longer have the `host`, `service` and will support all types of objects. See https://github.com/Icinga/icinga-notifications/issues/85

This makes it difficult to identify the object, whether it is a host, a service or a certificate. The search bar cannot explicitly search for hosts and services. Rendering lists (event, incident) no longer works as before.

There should be some hooks that can do the job based on the source.

raviks789:
resolves #132 

Blocked by: #155, #166

requires: https://github.com/Icinga/ipl-web/pull/217

### DIFF
--- a/library/Notifications/Hook/ObjectsRendererHook.php
+++ b/library/Notifications/Hook/ObjectsRendererHook.php
@@ -1,0 +1,211 @@
+<?php
+
+/* Icinga Notifications Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Notifications\Hook;
+
+use Exception;
+use Generator;
+use Icinga\Application\Hook;
+use Icinga\Application\Logger;
+use Icinga\Module\Notifications\Model\Objects;
+use ipl\Html\HtmlString;
+use ipl\Html\ValidHtml;
+use ipl\Web\Url;
+use ipl\Web\Widget\Link;
+
+/**
+ * Base hook to prepare and render objects
+ */
+abstract class ObjectsRendererHook
+{
+    /**
+     * Array of Object ID tags for each source
+     *
+     * It has the following structure : ['object source type' => ['object ID' => [object ID tags]]].
+     *
+     * @var array<string, array<string, array<string, string>>>
+     */
+    private static $objectIdTags = [];
+
+    /**
+     * Array of HTMLs for objects with their corresponding object IDs as keys
+     *
+     * It has the following structure : ['object ID' => registered HTML for the object name].
+     *
+     * @var array<string, ValidHtml>
+     */
+    private static $objectNameHtmls = [];
+
+    /**
+     * Get the HTML for the object names for the objects using the object ID tags
+     *
+     * @param array<array<string, string>> $objectIdTags Array of object ID tags of the objects belonging to the source
+     *
+     * @return Generator<array<string, string>, ValidHtml> Generator for object name HTMLs with their object ID tags
+     *                                                     as keys
+     */
+    abstract public function getHtmlForObjectNames(array $objectIdTags): Generator;
+
+    /**
+     * Get the source type of the objects
+     *
+     * @return string
+     */
+    abstract public function getSourceType(): string;
+
+    /**
+     * Create the object link for the given object ID tag
+     *
+     * @param array<string, string> $objectIdTag
+     *
+     * @return ValidHtml
+     */
+    abstract public function createObjectLink(array $objectIdTag): ValidHtml;
+
+    /**
+     * Register object ID tags to the cache
+     *
+     * @param Objects $obj
+     *
+     * @return void
+     */
+    final public static function register(Objects $obj): void
+    {
+        self::$objectIdTags[$obj->source->type][$obj->id] = $obj->id_tags;
+    }
+
+    /**
+     * Load HTMLs to be rendered for the object names to the cache using the cached objects
+     *
+     * @return void
+     */
+    final public static function load(): void
+    {
+        self::prepare(self::$objectIdTags);
+
+        self::$objectIdTags = [];
+    }
+
+    /**
+     * Prepare the objects to be rendered using the given object ID tags for each source
+     *
+     * The supplied object ID tags must have the following structure:
+     * ['object source type' => ['object ID' => [object ID tags]]].
+     *
+     * @param array<string, array<string, array<string, string>>> $objectIdTags Array of object ID tags for each source
+     *
+     * @return void
+     */
+    private static function prepare(array $objectIdTags): void
+    {
+        $idTagToObjectIdMap = [];
+        foreach ($objectIdTags as $sourceType => $objects) {
+            foreach ($objects as $objectId => $tags) {
+                if (! isset(self::$objectNameHtmls[$objectId])) {
+                    $idTagToObjectIdMap[$sourceType][] = [$objectId, $tags];
+                }
+            }
+        }
+
+        $objectDisplayNames = [];
+
+        /** @var self $hook */
+        foreach (Hook::all('Notifications\\ObjectsRenderer') as $hook) {
+            $source = $hook->getSourceType();
+
+            if (isset($idTagToObjectIdMap[$source])) {
+                try {
+                    $objectIDTagsForSource = array_map(
+                        function ($object) {
+                            return $object[1];
+                        },
+                        $idTagToObjectIdMap[$source]
+                    );
+
+                    /** @var array $objectIdTag */
+                    foreach ($hook->getHtmlForObjectNames($objectIDTagsForSource) as $objectIdTag => $validHtml) {
+                        foreach ($idTagToObjectIdMap[$source] as $key => $val) {
+                            $diff = array_intersect_assoc($val[1], $objectIdTag);
+                            if (count($diff) === count($val[1])) {
+                                unset($idTagToObjectIdMap[$key]);
+                                $objectDisplayNames[$val[0]] = $validHtml;
+
+                                continue 2;
+                            }
+                        }
+                    }
+                } catch (Exception $e) {
+                    Logger::error('Failed to load hook %s:', get_class($hook), $e);
+                }
+            }
+        }
+
+        self::$objectNameHtmls += $objectDisplayNames;
+    }
+
+    /**
+     * Get the object name of the given object
+     *
+     * If an HTML for the object name is not loaded, it is prepared using object ID tags and the same is returned.
+     *
+     * @param Objects $obj
+     *
+     * @return ValidHtml
+     */
+    final public static function getObjectName(Objects $obj): ValidHtml
+    {
+        $objId = $obj->id;
+        if (! isset(self::$objectNameHtmls[$objId])) {
+            self::prepare([$obj->source->type => [$objId => $obj->id_tags]]);
+        }
+
+        if (isset(self::$objectNameHtmls[$objId])) {
+            return self::$objectNameHtmls[$objId];
+        }
+
+        $objectTags = [];
+
+        foreach ($obj->id_tags as $tag => $value) {
+            $objectTags[] = sprintf('%s=%s', $tag, $value);
+        }
+
+        self::$objectNameHtmls[$objId] = new HtmlString(implode(', ', $objectTags));
+
+        return self::$objectNameHtmls[$objId];
+    }
+
+    /**
+     * Render object link for the given object
+     *
+     * @param Objects $object
+     *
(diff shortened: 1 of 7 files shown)


nilmerg:
Please rebase, fix conflicts and mark as ready for review.

sukhwinder33445:
Please rebase and resolve the conflicts.

sukhwinder33445 (changes requested):

library/Notifications/Widget/ItemList/EventList.php
```diff
@@ -27,10 +38,36 @@ public function __construct(ResultSet $data)
...
+    /**
+     * Get the rendered object display name for the given object ID
+     *
+     * @param string $objectID
+     *
+     * @return ?ValidHtml
+     */
+    public function getRenderedObjectDisplayName(string $objectID): ?ValidHtml
+    {
+        return $this->renderedObjectDisplayNames[$objectID] ?? null;
```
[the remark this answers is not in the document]
raviks789: This is a dead code.

library/Notifications/Hook/EventsObjectsInfoHook.php
```diff
@@ -0,0 +1,119 @@
...
+     * @return ?ValidHtml
+     */
+    final public static function getObjectListItemWidget(string $source, array $objectIdTag): ?ValidHtml
+    {
+        $objectListItemWidget = null;
+        foreach (Hook::all('Notifications\\EventsObjectsInfo') as $hook) {
+            /** @var self $hook */
+            try {
+                if ($source === $hook->getSource()) {
+                    $objectListItemWidget = $hook->getHtmlForObject($objectIdTag);
```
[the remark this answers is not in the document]
sukhwinder33445: You have still not returned in the loop. Should the first `$objectListItemWidget` found be overwritten if there is more than one hook?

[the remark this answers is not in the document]
raviks789: No, every object as a single source so this will not be replaced and that is also the reason for checking whether source's hook matches with object's hook. But I guess it is still better to return here.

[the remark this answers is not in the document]
sukhwinder33445: Yes, because further iterations are useless.

library/Notifications/Hook/EventsObjectsInfoHook.php
```diff
@@ -0,0 +1,118 @@
...
+     * @return array<string, ValidHtml>
+     */
+    abstract public function getObjectNameWidgets(array $objectIdTags): array;
+
+    /**
+     * Get the source of the objects
+     *
+     * @return string
+     */
+    abstract public function getSource(): string;
```
sukhwinder33445, line 35: It returns the source type, name the method accordingly, please.

library/Notifications/Hook/EventsObjectsInfoHook.php
```diff
@@ -0,0 +1,118 @@
...
+{
+    /** @var array<string, ValidHtml>  */
+    protected static $objectNameWidgets = [];
+
+    /**
+     * Get the object name widgets for the objects using the object ID tags
+     *
+     * @param array<string, array<string, string>> $objectIdTags
+     *
+     * @return array<string, ValidHtml>
```
sukhwinder33445, line 26: This type hint seems to be incorrect.
It should be `@return array<string, array<string, ValidHtml>` // Outer array key must be source type, inner array key must be object id. Please document this.

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,143 @@
...
+     * @return ValidHtml
+     */
+    abstract public function createObjectLink(array $objectIdTag): ValidHtml;
+
+    /**
+     * Register HTMLs to be rendered for the object names using the cached object ID tags
+     *
+     * @return void
+     */
+    final public static function registerHtmlForObjectNames(): void
```
sukhwinder33445, line 68: Please wrap the function body in a `new private static method` and call it here.  his new method can then also be called in `getObjectName` if this registration does not take place. For example, if the object is not a list item.

Function name : e.g. `prepare(Array)`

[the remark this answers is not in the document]
sukhwinder33445: Please change the name of the property accordingly.

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,143 @@
...
+     * @param Objects $obj
+     *
+     * @return ValidHtml
+     */
+    final public static function getObjectName(Objects $obj): ValidHtml
+    {
+        $objId = $obj->id;
+        if (isset(self::$registeredObjectNameHtmls[$objId])) {
+            return self::$registeredObjectNameHtmls[$objId];
+        }
```
sukhwinder33445, line 98: And here we can call the new method.
```suggestion
        if (! isset(self::$registeredObjectNameHtmls[$objId])) {
            $data[$obj->source->type][$objId] = $obj->id_tags;
            self::prepare($data);
        }
        
        if (isset(self::$registeredObjectNameHtmls[$objId])) {
            return self::$registeredObjectNameHtmls[$objId];
        }
```

library/Notifications/Widget/ItemList/EventListItem.php
```diff
@@ -105,14 +107,15 @@ protected function assembleTitle(BaseHtmlElement $title): void
 
         /** @var Objects $obj */
         $obj = $this->item->object;
-        $name = $obj->getName();
+        $name = (new DeferredText(function () use ($obj) {
```
sukhwinder33445, line 110: `DefferedText` is not required here, as the name has already been rendered. Please undo the change.

sukhwinder33445 (changes requested):

library/Notifications/Hook/ObjectsRendererHook.php
raviks789: If an HTML for the object name is not already loaded, I don't think doing this has any effect. If you are suggesting this because of Event/Incident detail view shown in a single column. It would still be unnecessary as the object is registered and the HTMLs for the object is already prepared. If not then it should be prepared from the object ID tags.

library/Notifications/Hook/ObjectsRendererHook.php
raviks789: I think this is superfluous.

library/Notifications/Hook/ObjectsRendererHook.php
raviks789: After discussing it offline. I think, I understand this suggestion better.

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -61,30 +64,58 @@ abstract public function getSourceType(): string;
...
+    private static function prepare(array $objectIdTags): void
     {
         $objectDisplayNames = [];
         foreach (Hook::all('Notifications\\ObjectsRenderer') as $hook) {
             /** @var self $hook */
             try {
-                $objectDisplayNames[] = $hook->getHtmlForObjectNames(self::$objectIdTagsCache[$hook->getSourceType()]);
+                $objectDisplayNames[] = $hook->getHtmlForObjectNames($objectIdTags[$hook->getSourceType()]);
+
+                unset($objectDisplayNames[$hook->getSourceType()]);
```
sukhwinder33445, line 106: I would prefer to reset the entire array after the loop. If there are 5 hooks, we call unset 5 times, which is unnecessary in my opinion.

sukhwinder33445 (dismissed):

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @return void
+     */
+    private static function prepare(array $objectIdTags): void
+    {
+        $objectDisplayNames = [];
+        foreach (Hook::all('Notifications\\ObjectsRenderer') as $hook) {
+            /** @var self $hook */
+            try {
+                $objectDisplayNames[] = $hook->getHtmlForObjectNames($objectIdTags[$hook->getSourceType()]);
```
sukhwinder33445, line 106: The call `$objectIdTags[$hook->getSourceType()]` must be fixed. It could throw `undefined array key` if there is no `icingadb` hook, for example.

library/Notifications/Widget/Detail/IncidentDetail.php
```diff
@@ -52,39 +51,15 @@ protected function createContacts()
...
+        $objectUrl = ObjectsRendererHook::renderObjectLink($this->incident->object);
+
+        if (! $objectUrl) {
+            return [];
+        }
 
         return [
-            Html::tag('h2', t('Object')),
-            $list
+            new HtmlElement('h2', null, Text::create(t('Object'))),
```
sukhwinder33445, line 61: Please use `Related Object` as content, just like in `EventDatail`.

sukhwinder33445 (commented):
Please `squash` the commits.

sukhwinder33445 (approved):
LGTM

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,293 @@
...
+    protected static $objectNameHtmls = [];
+
+    /**
+     * Array of object names with their corresponding object IDs as keys
+     *
+     * It has the following structure : ['object ID' => 'object name'].
+     *
+     * @var array<string, string>
+     */
+    protected static $objectNames = [];
```
nilmerg, line 49: These should be private. I'd not want that implementations can change these.

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,293 @@
...
+
+    /**
+     * Get the object names for the objects using the object ID tags
+     *
+     * @param array<string, array<string, string>> $objectIdTags Array of object ID tags with their corresponding
```
nilmerg, line 54: the type doc doesn't match the description

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,293 @@
...
+     *                                                           ['object ID' => [object ID tags]]
+     *
+     * @return Generator<string, string> Generator for object names with their object ID tags as keys
+     */
+    abstract public function getObjectNames(array $objectIdTags): Generator;
+
+    /**
+     * Get the HTML for the object names for the objects using the object ID tags
+     *
+     * @param array<string, array<string, string>> $objectIdTags Array of object ID tags with their corresponding
```
nilmerg, line 65: the type doc doesn't match the description

nilmerg (changes requested):

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -0,0 +1,299 @@
...
+
+        return self::createObjectNameAsString($obj);
+    }
+
+    private static function createObjectNameAsString(Objects $obj)
+    {
+        $objId = $obj->id;
+        if (isset(self::$objectNames[$objId])) {
+            return self::$objectNames[$objId];
+        }
```
nilmerg, line 248: Move this to `getObjectNameAsString`.

nilmerg (approved):

library/Notifications/Hook/ObjectsRendererHook.php
```diff
@@ -277,13 +277,15 @@ final public static function renderObjectLink(Objects $object): ?ValidHtml
...
-                        ->addAttributes([
+                    $objectLink = $hook->createObjectLink($object->id_tags);
+                    if ($objectLink !== null) {
+                        return $objectLink->addAttributes([
                             'class' => [
                                 'icinga-module',
                                 'module-' . ($sourceType === 'icinga2' ? 'icingadb' : $sourceType)
                             ]
                         ]);
+                    }
```
nilmerg, line 288: The loop should end in this case, otherwise we allow multiple implementations for the same source type.

nilmerg (changes requested):

library/Notifications/Hook/ObjectsRendererHook.php
sukhwinder33445: There is a return statement in the if condition.

nilmerg: That was previously the case. Now it's conditional.

PR #169, opened 2024-04-03
Make incident age field intuitive

raviks789:


sukhwinder33445 (changes requested):
Cannot save the incident age to db.

raviks789:
> Cannot save the incident age to db.

As discussed offline this is a known issue in this PR, that does not need to be fixed as the PR #170 makes it obsolete.

library/Notifications/Widget/ItemList/EscalationConditionListItem.php
```diff
@@ -26,23 +26,29 @@ class EscalationConditionListItem extends BaseHtmlElement
     /** @var FormElement Condition value */
     public $conditionVal;
 
+    /** @var ?FormElement Unit of the condition value */
+    public $conditionUnit;
```
sukhwinder33445, line 30: Make it `protected` please.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -129,28 +130,21 @@ protected function assemble(): void
...
-                                        return false;
-                                    }
-
-                                    $validator->clearMessages();
-                                    return true;
-                                })
-                            ]
+                            'class'      => 'right-operand',
+                            'value'      => 1
+                        ]
```
sukhwinder33445, line 139: Only whole numbers are supported now, `1.5h` is invalid.
- Please add `0.5` as step. 
- Remove second.
- Minimum allowed minute should be `1m`, not `0.5m`.

sukhwinder33445 (changes requested):

sukhwinder33445 (commented):
- It is still possible to set `0.5m`, yes we cannot save it, or when reloading it becomes `1`, but it is a bug. You can simply add a default attribute `value` to the `$valUnit` element to prevent this behaviour.
- There is extra margin.
Screenshot 2024-05-15 at 12 57 05

application/forms/EventRuleConfigForm.php
```diff
@@ -296,7 +296,16 @@ public function populate($values): self
                 }
 
                 $conditionFormValues['operator_' . $count] = QueryString::getRuleSymbol($filter);
-                $conditionFormValues['val_' . $count] = $filter->getValue();
+                $conditionValue = $filter->getValue();
+                if (
+                    preg_match('~^(\d+(?:\.?\d*))?([hms])$~', $conditionValue, $matches)
+                    && count($matches) === 3
```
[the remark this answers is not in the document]
raviks789: I think this can be implemented in a more simpler way using `str_split`. I will make the changes.

sukhwinder33445 (changes requested):
Each automatic submit of the form resets the `incident_age` value.

raviks789:
> Each autosubmit of the form resets the `incident_age` value.

This should be fixed now.

sukhwinder33445 (approved):
LGTM

nilmerg:
Obsoleted by #159

PR #170, opened 2024-04-05
Remove event rule session storage

### LINKED ISSUE #173: Remove Event Rule session storage
Once the form for event rule configuration has been rewritten. Remove the session storage which would not be necessary.

raviks789:
resolves #173 

Blocked by #159


sukhwinder33445 (changes requested):
When I add a new event rule and refresh the form before submitting, the name disappears and I can no longer save the form.

raviks789:
> When I add a new event rule and refresh the form before submitting, the name disappears and I can no longer save the form.

You should not be able to save the form after clicking refresh, this is the right behavior. As you would lose the data in event rule config form. But the name should not disappear as it is independent from event rule config form.

application/forms/EventRuleConfigForm.php
```diff
@@ -236,7 +228,7 @@ protected function assemble(): void
                 ['class' => 'filter-wrapper'],
                 [
                     Html::tag('li', (new FlowLine())->getRightArrow()),
-                    Html::tag('li', $configFilter),
+                    Html::tag('li', ['id' => 'config-filter'], $configFilter),
                     Html::tag('li', (new FlowLine())->getHorizontalLine())
```
sukhwinder33445, line 232: Use `HtmlElement` instead, please.

application/forms/EventRuleConfigForm.php
```diff
@@ -383,13 +375,19 @@ public function getValues(): array
...
-            $escalations[$i]['id'] = $escalationCondition->getValue('id');
-            $escalations[$i]['recipients'] = $escalationRecipient->getRecipients();
+            if ($this->hasElement('escalation-condition_' . $prefixMap)) {
+                /** @var EscalationCondition $escalationCondition */
+                $escalationCondition = $this->getElement('escalation-condition_' . $prefixMap);
+                $escalations[$i]['condition'] = $escalationCondition->getCondition();
+                $escalations[$i]['id'] = $escalationCondition->getValue('id');
+            }
+
+            if ($this->hasElement('escalation-condition_' . $prefixMap)) {
```
sukhwinder33445, line 385: It's the same condition as above.

application/forms/EventRuleConfigForm.php
```diff
@@ -411,7 +409,7 @@ public function getValues(): array
     protected function createRemoveButton(string $prefix): SubmitButtonElement
     {
         /** @var array<int, array<string, mixed>> $escalations */
-        $escalations = $this->config['rule_escalation'] ?? [];
+        $escalations = $this->getValues()['rule_escalation'];
```
sukhwinder33445, line 412: The `getValues()` method is called too often and performs a query each time. This should cache the last changes to avoid the unnecessary process.

public/css/detail/event-rule-detail.less
```diff
@@ -38,6 +30,14 @@
       width: 25em;
     }
   }
+
+  .control-button.disabled {
```
sukhwinder33445, line 34: There is no hint for delete button, why it is disabled.

sukhwinder33445 (changes requested):

application/controllers/EventRuleController.php
```diff
@@ -286,17 +299,85 @@ public static function createFilterString(Filter\Rule $filters): string
...
+                    'title' => $this->translate('Cannot edit event rule until it is saved to database')
+                ],
+                (new Link(
+                    new Icon('edit'),
+                    Url::fromPath('notifications/event-rule/edit', [
+                        'id' => -1
+                    ]),
+                    ['class' => ['control-button', 'disabled']]
+                ))->openInModal()
+            )
```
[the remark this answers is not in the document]
raviks789: Makes sense.

application/controllers/EventRuleController.php
```diff
@@ -145,8 +120,8 @@ public function indexAction(): void
             }
         }
 
-        $eventRuleForm = Html::tag('div', ['class' => 'event-rule-form'], [
-            Html::tag('h2', $configValues['name'] ?? ''),
+        $eventRuleForm = Html::tag('div', ['class' => 'event-rule-form', 'id' => 'event-rule-form'], [
+            Html::tag('h2', $eventRuleConfigValues['name'] ?? ''),
```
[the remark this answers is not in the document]
raviks789: As discussed offline with @nilmerg, we are going to use session storage to temporarily store the new event rule details. The name can be empty/null again as the event rule details will be lost as soon as the user clicks refresh. So I don't have to change anything here.

application/controllers/EventRuleController.php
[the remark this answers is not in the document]
raviks789: As explained offline, since not much in the detail view as to be changed once the event rule is added. We will retain the edit link, but will just toggle between disabled and not disabled.

public/css/detail/event-rule-detail.less
raviks789: This has been fixed in the original PR #159

sukhwinder33445:
The `switch case` in `EventRuleConfigForm::insertOrUpdateEscalations() Line: 587` is superfluous. You can initialize the `$data` variable directly.

application/forms/EventRuleConfigForm.php
```diff
@@ -18,6 +18,7 @@
 use ipl\Html\Form;
 use ipl\Html\FormElement\SubmitButtonElement;
 use ipl\Html\HtmlElement;
+use ipl\Html\ValidHtml;
```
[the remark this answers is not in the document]
sukhwinder33445: I think the `ON_DELETE` can also be removed and can be checked in other ways.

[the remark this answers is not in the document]
raviks789: But I think it is not incorrect to handle it this way. And also, it is clear what the event is.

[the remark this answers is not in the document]
sukhwinder33445: There are two methods that are overridden to define this event, this can be omitted.

application/forms/EventRuleConfigForm.php
```diff
@@ -628,6 +642,33 @@ public function removeRule(int $id): void
...
+            }
+
+            return empty(array_udiff_assoc($a, $b, $checker)) ? 0 : 1;
+        };
+
+        $diff = array_udiff_assoc($values, $dbValuesToCompare, $checker);
+
+        // Search editor is not part of the form and the filter is dynamically updated in the config after its
+        // submission. Hence, it must be part of the changes.
+        $diff['object_filter'] = $this->config['object_filter'] ?? null;
```
sukhwinder33445, line 667: Now the returned value is never empty, which leads to false positives.
Do not edit anything and click on `Save`, then the success message will appear.

application/controllers/EventRuleController.php
```diff
@@ -291,31 +280,49 @@ public function editAction(): void
...
-                $config['is_active'] = $form->getValue('is_active');
+            ->on(Form::ON_SUCCESS, function ($form) use ($ruleId, $db) {
                 if ($ruleId === '-1') {
-                    $redirectUrl = Url::fromPath('notifications/event-rules/add', ['id' => '-1']);
-                } else {
-                    $redirectUrl = Url::fromPath('notifications/event-rule', ['id' => $ruleId]);
+                    $db->insert('rule', [
+                        'name'          => $form->getValue('name'),
+                        'timeperiod_id' => null,
+                        'object_filter' => null,
```
sukhwinder33445, line 301: Why are you inserting null? These columns are already nullable.

sukhwinder33445: You can call `$form->getValues()` to get both values.
```suggestion
    $db->insert('rule', $this->getValues());
```

application/controllers/EventRuleController.php
```diff
@@ -291,31 +280,49 @@ public function editAction(): void
...
+
+                    $id = $db->lastInsertId();
+
+                    $this->getResponse()->setHeader('X-Icinga-Container', 'col2');
                     $this->sendExtraUpdates(['#col1']);
+                    $this->redirectNow(Links::eventRule($id));
+                } else {
+                    $db->update('rule', [
+                        'name'          => $form->getValue('name'),
+                        'is_active'     => $form->getValue('is_active')
```
sukhwinder33445, line 313: Same here as above.

application/forms/EventRuleConfigForm.php
```diff
@@ -415,62 +459,50 @@ protected function createRemoveButton(string $prefix): SubmitButtonElement
      *
      * @return int
      */
-    public function addOrUpdateRule(int $id, array $config): int
+    public function updateRule(int $id, array $config): int
     {
         $db = Database::get();
-
         $db->beginTransaction();
+        $db->update('rule', ['object_filter' => $config['object_filter'] ?? null], ['id = ?' => $id]);
```
sukhwinder33445, line 466: In this way, the saved value is updated, even if it remains the same if there are no changes in the filter.

sukhwinder33445 (changes requested):

application/forms/EventRuleConfigForm.php
raviks789: I am aware of this. But since, there was no priority on the PR. I did not work on this issue. I will fix it.

sukhwinder33445 (commented):
I cannot remove a recently added escalation.

raviks789:
> I cannot remove a recently added escalation.

I think this is related to the `EventRuleConfigForm::getChanges()` method you have implemented. The method is returning an empty array instead of the changes. This is causing the form to not save the changes.

sukhwinder33445:
No the method works fine, the checked values are equal.

raviks789:
> No the method works fine, the checked values are equal.

I just checked. The checked values are different. I mean `$values` and `$dbValuesToCompare` are different. And this only happens when the last escalation is being deleted.

sukhwinder33445:
You're right, either it worked before or we simply overlooked the bug. This must be fixed.

sukhwinder33445:
`EventRuleConfigForm::getChanges()` should work now, please test.

application/forms/EventRuleConfigForm.php
```diff
@@ -415,62 +463,50 @@ protected function createRemoveButton(string $prefix): SubmitButtonElement
      *
      * @return int
      */
-    public function addOrUpdateRule(int $id, array $config): int
+    public function updateRule(int $id, array $config): int
     {
         $db = Database::get();
-
         $db->beginTransaction();
+        $db->update('rule', ['object_filter' => $this->objectFilter], ['id = ?' => $id]);
```
sukhwinder33445, line 470: Please only update `object_filter` if it has been changed.

raviks789: If there are no changes, the transaction does not even happen.

sukhwinder33445: As discussed offline, this is also triggered if we only add a new escalation and the filter remains the same.

PR #181, opened 2024-04-22
Move source icon in incident detail to the header

### LINKED ISSUE #161: Move source icon in incident detail to the header
Objects are tied to single sources by now. There's no reason to show the source icon for every history entry anymore.

But since it's tied to the object, it should instead be shown in the header alongside the object's name. (i.e. Aligned right, as previously)

A beneficial side effect should be, that the icon is also visible in the primary incident list from there on.

raviks789:
fixes #161

### DIFF
--- a/library/Notifications/Widget/ItemList/IncidentHistoryListItem.php
+++ b/library/Notifications/Widget/ItemList/IncidentHistoryListItem.php
@@ -14,7 +14,7 @@ use ipl\Web\Widget\Icon;
 use ipl\Web\Widget\TimeAgo;
 
 /**
- * Event item of an event list. Represents one database row.
+ * Incident history item of an incident history list. Represents one database row.
  */
 class IncidentHistoryListItem extends BaseListItem
 {
--- a/library/Notifications/Widget/ItemList/IncidentHistoryListItem.php
+++ b/library/Notifications/Widget/ItemList/IncidentHistoryListItem.php
@@ -8,7 +8,6 @@ use Icinga\Module\Notifications\Common\Icons;
 use Icinga\Module\Notifications\Model\Event;
 use Icinga\Module\Notifications\Model\IncidentHistory;
 use Icinga\Module\Notifications\Widget\IconBall;
-use Icinga\Module\Notifications\Widget\SourceIcon;
 use ipl\Html\BaseHtmlElement;
 use ipl\Web\Common\BaseListItem;
 use ipl\Web\Widget\Icon;
@@ -50,11 +49,6 @@ class IncidentHistoryListItem extends BaseListItem
         }
 
         $header->addHtml($this->createCaption());
-        if ($this->item->type === 'opened' || $this->item->type === 'incident_severity_changed') {
-            $header->add(
-                (new SourceIcon(SourceIcon::SIZE_BIG))->addHtml($this->item->incident->object->source->getIcon())
-            );
-        }
 
         $header->add(new TimeAgo($this->item->time->getTimestamp()));
     }
--- a/library/Notifications/Widget/ItemList/IncidentListItem.php
+++ b/library/Notifications/Widget/ItemList/IncidentListItem.php
@@ -23,7 +23,7 @@ use ipl\Web\Widget\TimeAgo;
 use ipl\Web\Widget\TimeSince;
 
 /**
- * Event item of an event list. Represents one database row.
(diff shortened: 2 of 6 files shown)


sukhwinder33445 (changes requested):
- The icon is no longer aligned in the center due to this change (https://github.com/Icinga/icinga-notifications-web/commit/bae69e5cbb08add2cf5978e76427356b0875aa5a)
- The icons in the list should be statically aligned.
Screenshot 2024-04-24 at 11 48 58

ncosta-ic:
> * The icon is no longer aligned in the center due to this change ([bae69e5](https://github.com/Icinga/icinga-notifications-web/commit/bae69e5cbb08add2cf5978e76427356b0875aa5a))
> 
> * The icons in the list should be statically aligned.
>

I stumbled accross the same problem and fixed it by adding a `margin-top: .15em` to the `i.icon` rule for `.source-icon` in `public/css/common.less`
I will not include this change in my PR (https://github.com/Icinga/icinga-notifications-web/pull/166) but it should be an easy fix.

sukhwinder33445:
> I stumbled accross the same problem and fixed it by adding a `margin-top: .15em` to the `i.icon` rule for `.source-icon` in `public/css/common.less`

It should also be fixed if you only remove this change. ([bae69e5](https://github.com/Icinga/icinga-notifications-web/commit/bae69e5cbb08add2cf5978e76427356b0875aa5a))

ncosta-ic:
> > I stumbled accross the same problem and fixed it by adding a `margin-top: .15em` to the `i.icon` rule for `.source-icon` in `public/css/common.less`
> 
> It should also be fixed if you only remove this change. ([bae69e5](https://github.com/Icinga/icinga-notifications-web/commit/bae69e5cbb08add2cf5978e76427356b0875aa5a))
source_icon_alignment
(The red box should be aligned with to the cyan guides)

I quickly compared all three variants:
- vertical-align: top
- vertical-align: top, margin-top: .15em
- without both css rules

From my point of view, the second option is also not quite right, but it's much closer to be perfectly aligned than the other two variants.

public/css/list/incident-list.less
```diff
@@ -0,0 +1,6 @@
+.incident-list .list-item {
```
nilmerg, line 1: This doesn't seem to be specific to the incident list. The event list could also need it, so please move this to common.less

nilmerg (changes requested):
The placement of the icon should be fixed separately.

library/Notifications/Widget/ItemList/IncidentHistoryListItem.php
```diff
@@ -76,11 +76,6 @@ protected function assembleHeader(BaseHtmlElement $header): void
         }
 
         $header->addHtml($this->createCaption());
-        if ($this->item->type === 'opened' || $this->item->type === 'incident_severity_changed') {
-            $header->add(
-                (new SourceIcon(SourceIcon::SIZE_BIG))->addHtml($this->item->event->object->source->getIcon())
```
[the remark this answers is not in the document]
raviks789: You were right, I have removed the join now.

library/Notifications/Widget/ItemList/IncidentListItem.php
```diff
@@ -73,19 +80,22 @@ protected function assembleTitle(BaseHtmlElement $title): void
...
-                ['class' => 'meta'],
-                [
-                    'closed ',
-                    new TimeAgo($this->item->recovered_at->getTimestamp())
-                ]
-            ));
+            $meta->add([
+                $this->translate('closed') . ' ',
+                new TimeAgo($this->item->recovered_at->getTimestamp())
+            ]);
```
nilmerg, line 93: With this, a translator has the option to move the time representation.

```suggestion
           $meta->addHtml(FormattedString::create(
               $this->translate('closed %s', '(incident) ... <relative time>'),
               new TimeAgo($this->item->recovered_at->getTimestamp())
           ));
```

nilmerg (changes requested):
> The placement of the icon should be fixed separately.

With this, I meant the placement of the icon inside the circle. Essentially, my answer to the discussion above.

Now, you've dismissed your own change that added a small gap to the right. I didn't mean you to do this! I still think it's suitable and a visual improvement. *If applied to the incident **and** events list*.

nilmerg (approved):

PR #183, opened 2024-04-24
Schedule redesign

### LINKED ISSUE #177: Redesigned Scheduler UI
### Step 1: Create a new Schedule
A separate schedules list is not only a better known structure within the Icinga Web 2 context, but helps navigating and organising schedules. A preview of the final schedule should be included to get a rough idea, what the schedule looks like. To add a new schedule list use the button "New Schedule".

Frame 2

### Step 2: Add a rotation

Starting with the schedule view featuring an empty state note. From there, the user can start adding Rotations by clicking the "Add rotation" button.

Frame 3

### Step 3: Configure rotation for office hours

There are 3 different modes of rotations (See below for details). Creating the office hour rotation leverages the Partial day rotation mode. The Modal features a list of rotation members. The order of the members defines their rotation order and can be edited in context of the rotation. Members can be contacts and contact groups. Click "Add rotation" to save and add the rotation to the schedule. The Modal should probably feature a preview of the configured rotation (not in mockup for now)

Frame 4

### Step 4: Update the schedule view

Adding a rotation updates the schedule preview. It displays the configure rotation as a row in the timeline. There's also a summary row on the bottom, that shows the result of the "stacked" rotations.

Frame 5

### Step 5: Add another partial day rotation for the out of office hours.

It's the same as before. The preview should visualise the free timeslots.

Frame 8

### Step 6: Display a second rotation row

Rotation rows can be reordered. The end result would not allow for any overlaps, lower rows win over higher ones. The "zoom factor" of the timeline can be altered from more detail to a larger timespan on the one hand. On the other hand there's also a calendar view that displays the final result in a calendar view (not in mockup).

Frame 9

### Step 7: Add another rotation for the weekend shifts.

Another rotation is added for the weekend shifts, which can be done with a multi-day rotation.

Frame 10

### Step 8: Visualising all 3 Rotations
Frame 11

nilmerg:
resolves #177 

**Requirements**

* https://github.com/Icinga/ipl-html/pull/137
* https://github.com/Icinga/ipl-web/pull/223
* https://github.com/Icinga/icinga-notifications/pull/204


sukhwinder33445:
The `Event Rules` and `Contacts` tabs add the `Add new *` button to the content, while the `Schedule` tab adds it to the controls.

julianbrost:
From my comment in the corresponding daemon PR (https://github.com/Icinga/icinga-notifications/pull/204#discussion_r1629084447):

> Editing rotations in Notifications Web still feels weird for me. As of me trying it, it's 10:55 and I'm editing a rotation that's active at that very moment, it sets `actual_handoff` to 11:00, i.e. if I save the currently rotation get disabled until a few minutes in the future. I'd expect it to set `actual_handoff` to now instead, so that it becomes immediately active.

It also looks like `actual_handoff` is not taken into account when rendering rotations in web, there an entry in a rotation with `actual_handoff` at 11:00 today is still shown to start at 10:00 today for me.

nilmerg:
> > Editing rotations in Notifications Web still feels weird for me. As of me trying it, it's 10:55 and I'm editing a rotation that's active at that very moment, it sets actual_handoff to 11:00, i.e. if I save the currently rotation get disabled until a few minutes in the future. I'd expect it to set actual_handoff to now instead, **so that it becomes immediately active.**

Which is the case, now.

> It also looks like actual_handoff is not taken into account when rendering rotations in web

I told you so, when I accepted your suggestion to not pre-fill the first handoff to a suitable value matching the pattern. If a rotation is being edited, and the first handoff isn't adjusted, only the actual handoff is updated.

ncosta-ic (approved):
Disclaimer: This is not a code-review. I just tested the functionalities provided and changed by this PR.


All rotation modes can be configured correctly and the resulting schedule is correct as well.
The rotations keep their defined date and there are no time derivations or gaps when adding or modifying them.
Additionally, all values are correctly saved into the database table. I was not able to produce data corruptions by using the web ui.

PR #186, opened 2024-04-30
Move schedule to primary list

### LINKED ISSUE #162: Move schedule selection to a primary list
## Current State

By navigating through the menu to schedules, the calendar plus the following header is shown:

image

The button on the right allows the user to create new schedules, while the dropdown on the left allows them to choose between existing ones. The button in the middle allows to adjust settings of the chosen schedule.

The dropdown is empty or not available if there are no schedules and the calendar [cannot be interacted with](https://github.com/Icinga/icinga-notifications-web/issues/104).

## Expected

As the linked issue already mentions, the calendar makes not much sense in case there a no schedules yet. Though, the proposed solution there is outdated now.

The calendar will undergo a full redesign in the near future. Not all is decided yet, though clear is that this must change fundamentally.

The new solution is to introduce a full blown primary list for schedules. Just like the one used for event rules:

* Pagination Control
* Limit Control
* Sort Control
* Search Bar with editor
* The button to create schedules is shown in the top left of the content
* The remaining content is an item list that uses items to link to existing schedules
  * The design of such an item is supposed to be very basic, only the name is shown, everything else is subject to the schedule redesign (but the design should already be similar to the `common` view type)
  * This should cause the calendar to be opened in the right column

### Routes

* The current route should be the new list
  * Causing it to be the primary entry point to schedules through the menu
* The calendar should be available at `notifications/schedule`
  * Currently used by the settings dialog, this must be moved to `notfications/schedule/settings`

### Changes to the Calendar

* The drop down plus label is removed
* The button to create schedules is removed
* The button to adjust a schedule's settings is kept as is, as changing it is again subject to the full redesign

sukhwinder33445:
resolves #162

### DIFF
--- a/application/controllers/ScheduleController.php
+++ b/application/controllers/ScheduleController.php
@@ -4,16 +4,56 @@
 
 namespace Icinga\Module\Notifications\Controllers;
 
+use Icinga\Module\Notifications\Common\Database;
+use Icinga\Module\Notifications\Common\Links;
 use Icinga\Module\Notifications\Forms\EntryForm;
 use Icinga\Module\Notifications\Forms\ScheduleForm;
+use Icinga\Module\Notifications\Model\Schedule;
+use Icinga\Module\Notifications\Widget\Calendar\Controls;
 use Icinga\Module\Notifications\Widget\RecipientSuggestions;
+use Icinga\Module\Notifications\Widget\Schedule as ScheduleWidget;
+use ipl\Html\Form;
 use ipl\Html\Html;
+use ipl\Stdlib\Filter;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Url;
+use ipl\Web\Widget\ButtonLink;
 
 class ScheduleController extends CompatController
 {
-    public function indexAction()
+    public function indexAction(): void
+    {
+        $id = (int) $this->params->getRequired('id');
+
+        $query = Schedule::on(Database::get())
+            ->filter(Filter::equal('schedule.id', $id));
+
+        /** @var ?Schedule $schedule */
+        $schedule = $query->first();
+        if ($schedule === null) {
+            $this->httpNotFound(t('Schedule not found'));
+        }
+
+        $this->addTitleTab(sprintf(t('Schedule: %s'), $schedule->name));
+
+        $this->controls->addHtml(
+            (new ButtonLink(
+                null,
+                Links::scheduleSettings($id),
+                'cog'
+            ))->openInModal()
+        );
+
+        $this->controls->addAttributes(['class' => 'schedule-controls']);
+
+        $calendarControls = (new Controls())
+            ->setAction(Url::fromRequest()->getAbsoluteUrl())
+            ->handleRequest($this->getServerRequest());
+
+        $this->addContent(new ScheduleWidget($calendarControls, $schedule));
+    }
+
+    public function settingsAction(): void
     {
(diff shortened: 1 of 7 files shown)


sukhwinder33445:
The `add/editEntryAction()` of `SchedulerController` uses `$this->sendExtraUpdates(['#col2'])` to update the entry in the calendar. But if there is only one column (list #col1 is closed), the hardcoded `#col2` in `sendExtraUpdates` cannot update the calendar.

library/Notifications/Widget/ItemList/ScheduleListItem.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Icinga\Module\Notifications\Widget\ItemList;
+
+use Icinga\Module\Notifications\Common\Links;
+use Icinga\Module\Notifications\Model\Schedule;
+use ipl\Html\BaseHtmlElement;
+use ipl\Web\Common\BaseListItem;
+use ipl\Web\Widget\Link;
+
+/**
+ * Channel item of a channel list. Represents one database row.
```
nilmerg, line 14: Ehh?

library/Notifications/Model/Schedule.php
```diff
@@ -5,28 +5,51 @@
...
     {
         return 'id';
     }
 
-    public function getColumns()
+    public function getColumns(): array
     {
-        return [
-            'name'
-        ];
```
nilmerg, line 26: undo this unnecessary change

application/controllers/ScheduleController.php
```diff
@@ -4,16 +4,59 @@
...
+                'cog'
+            ))->openInModal()
+        );
+
+        $calendarControls = (new Controls())
+            ->setAction(Url::fromRequest()->getAbsoluteUrl());
+        if ($this->getRequest()->getHeader('X-Icinga-Container') === 'modal-content') {
+            $this->getResponse()->setHeader('X-Icinga-Modal-Layout', 'wide');
+            $calendarControls->setBaseTarget('modal-content');
+        }
```
nilmerg, line 51: This is/was only for testing purposes which shouldn't have been introduced to `main` to begin with. Remove this please.

application/controllers/SchedulesController.php
```diff
@@ -5,102 +5,126 @@
...
-            $this->getResponse()->setHeader('X-Icinga-Modal-Layout', 'wide');
-            $calendarControls->setBaseTarget('modal-content');
+        $this->addContent(new ScheduleList($schedules));
+
+        if (! $searchBar->hasBeenSubmitted() && $searchBar->hasBeenSent()) {
+            $this->sendMultipartUpdate();
         }
 
-        $this->addControl($controls);
-        $this->controls->addAttributes(['class' => 'schedule-controls']);
```
nilmerg, line 80: I miss this in the new controller.

application/controllers/SchedulesController.php
```diff
@@ -5,102 +5,126 @@
...
+        $this->getDocument()->add($editor);
+        $this->setTitle(t('Adjust Filter'));
+    }
+
+    /**
+     * Get the filter created from query string parameters
+     *
+     * @return Filter\Rule
+     */
+    public function getFilter(): Filter\Rule
```
nilmerg, line 106: Make this private as well, as is the property

nilmerg (changes requested):

nilmerg (approved):
please rebase

Issue #174, opened 2024-04-10
Contactgroup Configuration


nilmerg:
## Goal

We need a way to configure contactgroups in the UI.

## How it Should Work

* A tab in the `Configuration` menu section should lead to a list of contactgroups
* This list allows to paginate, limit, sort, search for and create contactgroups
* The button to create a contactgroup should be on the top left of the list
  * Clicking on it, opens a modal which allows to choose a name and define the group's members by using a `TermInput`
  * Saving closes the modal and shows the group's detail in the right column
* A click on a list item opens up a detail on the right, showing the name of the group in the header
  * A button on the name's right side allows to edit it
  * The modal should be the same as when creating a group, plus a remove button
* The detail's content lists a group's members, clicking on such an item leads to the contact's configuration view

ncosta-ic:
I might consider updating the ContactGroup class once https://github.com/Icinga/icinga-notifications/issues/187 gets sorted out.
It currently links to the `contactgroup` table and this implies a class name of `Contactgroup`, which seems wrong.

PR #188, opened 2024-05-22
Contactgroup configuration

sukhwinder33445:
resolves https://github.com/Icinga/icinga-notifications-web/issues/174

### DIFF
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -87,7 +87,7 @@ class ContactGroupController extends CompatController
                     }
                 }
             })
-            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) use ($groupId) {
                 if ($form->editGroup()) {
                     Notification::success(sprintf(
                         t('Successfully updated contact group %s'),
@@ -95,7 +95,7 @@ class ContactGroupController extends CompatController
                     ));
                 }
 
-                $this->closeModalAndRefreshRemainingViews($this->getRequest()->getUrl()->getAbsoluteUrl());
+                $this->closeModalAndRefreshRemainingViews(Links::contactGroup($groupId));
             })
             ->handleRequest($this->getServerRequest());
 
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -68,7 +68,7 @@ class ContactGroupController extends CompatController
 
         $form = (new ContactGroupForm(Database::get()))
             ->loadContactgroup($groupId)
-            ->setAction($this->getRequest()->getUrl()->getAbsoluteUrl())
+            ->setAction((string) Links::contactGroupEdit($groupId))
             ->on(Form::ON_SENT, function (ContactGroupForm $form) {
                 if ($form->hasBeenRemoved()) {
                     $form->removeContactgroup();
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -5,8 +5,20 @@
 namespace Icinga\Module\Notifications\Controllers;
 
 use Icinga\Module\Notifications\Common\Database;
-use ipl\Sql\Connection;
+use Icinga\Module\Notifications\Common\Links;
+use Icinga\Module\Notifications\Forms\ContactGroupForm;
+use Icinga\Module\Notifications\Model\Contactgroup;
+use Icinga\Module\Notifications\Widget\ItemList\ContactList;
+use Icinga\Web\Notification;
+use ipl\Html\Attributes;
+use ipl\Html\Form;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Stdlib\Filter;
+use ipl\Web\Common\BaseItemList;
 use ipl\Web\Compat\CompatController;
+use ipl\Web\Widget\ButtonLink;
 
 class ContactGroupController extends CompatController
 {
@@ -15,9 +27,90 @@ class ContactGroupController extends CompatController
         $this->assertPermission('notifications/config/contact-groups');
     }
 
+    public function indexAction(): void
+    {
+        $groupId = $this->params->getRequired('id');
+
+        $query = Contactgroup::on(Database::get())
+            ->columns(['id', 'name'])
+            ->filter(Filter::equal('id', $groupId));
+
+        $group = $query->first();
+        if ($group === null) {
+            $this->httpNotFound(t('Contact group not found'));
+        }
+
+        $this->controls->addAttributes(['class' => 'contactgroup-detail']);
+
+        $this->addControl(new HtmlElement('div', new Attributes(['class' => 'header']), Text::create($group->name)));
+
+        $this->addControl($this->createPaginationControl($group->contact));
+        $this->addControl($this->createLimitControl());
+
+        $this->addContent(
+            (new ButtonLink(
+                Text::create(t('Edit Contact Group')),
+                Links::contactGroupEdit($groupId),
+                'edit',
+                ['class' => 'add-new-component']
+            ))->openInModal()
+        );
+
+        $this->addContent(new ContactList($group->contact));
+
+        $this->addTitleTab(t('Contact Group'));
+        $this->setTitle(sprintf(t('Contact Group: %s'), $group->name));
+    }
+
     public function editAction(): void
     {
-        // TODO: add edit logic
         $groupId = $this->params->getRequired('id');
+
+        $form = (new ContactGroupForm(Database::get()))
+            ->loadContactgroup($groupId)
+            ->setAction($this->getRequest()->getUrl()->getAbsoluteUrl())
+            ->on(Form::ON_SENT, function (ContactGroupForm $form) {
+                if ($form->hasBeenRemoved()) {
+                    $form->removeContactgroup();
+                    Notification::success(sprintf(
+                        t('Successfully removed contact group %s'),
+                        $form->getValue('group_name')
+                    ));
+                    $this->switchToSingleColumnLayout();
+                } elseif ($form->hasBeenCancelled()) {
+                    $this->closeModalAndRefreshRelatedView($this->getRequest()->getUrl()->getAbsoluteUrl());
+                } elseif (! $form->hasBeenSubmitted()) {
+                    foreach ($form->getPartUpdates() as $update) {
+                        if (! is_array($update)) {
+                            $update = [$update];
+                        }
+
+                        $this->addPart(...$update);
+                    }
+                }
+            })
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                if ($form->editGroup()) {
+                    Notification::success(sprintf(
+                        t('Successfully updated contact group %s'),
+                        $form->getValue('group_name')
+                    ));
+                }
+
+                $this->closeModalAndRefreshRemainingViews($this->getRequest()->getUrl()->getAbsoluteUrl());
+            })
+            ->handleRequest($this->getServerRequest());
+
+        $this->addContent($form);
+        $this->setTitle(t('Edit Contact Group'));
+    }
+
+    protected function addContent(ValidHtml $content): self
+    {
+        if ($content instanceof BaseItemList) {
+            $this->content->getAttributes()->add('class', 'full-width');
+        }
+
+        return parent::addContent($content);
     }
 }
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -1,0 +1,28 @@
+<?php
+
+/* Icinga Notifications Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Notifications\Controllers;
+
+use Icinga\Module\Notifications\Common\Database;
+use ipl\Sql\Connection;
+use ipl\Web\Compat\CompatController;
+
+class ContactGroupController extends CompatController
+{
+    /** @var Connection $db */
+    private $db;
+
+    public function init()
+    {
+        $this->assertPermission('notifications/config/contact-groups');
+
+        $this->db = Database::get();
+    }
+
+    public function editAction(): void
+    {
+        // TODO: add edit logic
+        $groupId = $this->params->getRequired('id');
+    }
+}
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -77,8 +77,6 @@ class ContactGroupController extends CompatController
                         $form->getValue('group_name')
                     ));
                     $this->switchToSingleColumnLayout();
-                } elseif ($form->hasBeenCancelled()) {
-                    $this->closeModalAndRefreshRelatedView($this->getRequest()->getUrl()->getAbsoluteUrl());
                 } elseif (! $form->hasBeenSubmitted()) {
                     foreach ($form->getPartUpdates() as $update) {
                         if (! is_array($update)) {
--- a/application/controllers/ContactGroupController.php
+++ b/application/controllers/ContactGroupController.php
@@ -50,7 +50,7 @@ class ContactGroupController extends CompatController
         $this->addContent(
             (new ButtonLink(
                 Text::create(t('Edit Contact Group')),
-                Links::contactGroupEdit($groupId),
+                Links::contactGroupEdit($groupId)->with(['showCompact' => true, '_disableLayout' => 1]),
                 'edit',
                 ['class' => 'add-new-component']
             ))->openInModal()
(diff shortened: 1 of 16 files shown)


application/controllers/ContactGroupController.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Icinga\Module\Notifications\Controllers;
+
+use Icinga\Module\Notifications\Common\Database;
+use ipl\Sql\Connection;
+use ipl\Web\Compat\CompatController;
+
+class ContactGroupController extends CompatController
+{
+    /** @var Connection $db */
+    private $db;
```
sukhwinder33445, line 14: As `Database::get()` returns a singleton instance, you can use the method directly.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+use ipl\Web\Compat\CompatController;
+use ipl\Web\Compat\SearchControls;
+use ipl\Web\Widget\ButtonLink;
+
+class ContactGroupsController extends CompatController
+{
+    use SearchControls;
+
+    /** @var Connection */
+    private $db;
```
sukhwinder33445, line 28: Please remove as described above.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+
+    public function init()
+    {
+        $this->assertPermission('notifications/config/contact-groups');
+        $this->db = Database::get();
+    }
+
+    public function getTabs()
+    {
+        if ($this->getRequest()->getActionName() === 'index') {
```
sukhwinder33445, line 38: Unwrap please.

library/Notifications/Widget/MemberSuggestions.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        $this->setSearchTerm($requestData['term']['label']);
+        $this->setOriginalValue($requestData['term']['search']);
+        $this->excludeTerms($requestData['exclude'] ?? []);
+
+        return $this;
+    }
+
+    protected function assemble()
```
sukhwinder33445, line 75: Add return type, please.

library/Notifications/Widget/MemberSuggestions.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Load suggestions as requested by the client
+     *
+     * @param ServerRequestInterface $request
+     *
+     * @return $this
+     */
+    public function forRequest(ServerRequestInterface $request)
```
sukhwinder33445, line 57: Add return type, please.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+            $this->sendMultipartUpdate();
+        }
+
+        $this->setAutorefreshInterval(30);
+
+        $this->setTitle($this->translate('Contact Groups'));
+        $this->getTabs()->activate('contact-groups');
+    }
+
+    protected function addAction(): void
```
sukhwinder33445, line 123: Please make these methods `public`.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+                }
+            })
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                $groupIdentifier = $form->addGroup();
+                if ($groupIdentifier) {
+                    Notification::success(t('New group has been successfully added'));
+                    $this->closeModalAndRefreshRemainingViews($this->getRequest()->getUrl()->getAbsoluteUrl());
+                } else {
+                    Notification::error(t('New group could not be added'));
+                    // close the modal and refresh col1
```
sukhwinder33445, line 147: Please remove this comment. The function already makes it obvious.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+                ->addAttributes(['class' => 'add-new-component'])
+        );
+
+        $this->addContent(new ContactGroupList($groups));
+
+        if (! $searchBar->hasBeenSubmitted() && $searchBar->hasBeenSent()) {
+            $this->sendMultipartUpdate();
+        }
+
+        $this->setAutorefreshInterval(30);
```
sukhwinder33445, line 117: Please remove it.

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,192 @@
...
+
+        if (! array_key_exists('group', $data)) {
+            return null;
+        } elseif (
+            ! array_key_exists('group_name', $data['group'])
+            || ! array_key_exists('group_members', $data['group'])
+        ) {
+            return null;
+        }
+
```
sukhwinder33445, line 160: You can simplify it:
```suggestion
if (! isset($data['group']['group_name'], $data['group']['group_members'])) {
            return null;
}
```

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+                    }
+                }
+            })
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                $groupIdentifier = $form->addGroup();
+                if ($groupIdentifier) {
+                    Notification::success(t('New group has been successfully added'));
+                    $this->closeModalAndRefreshRemainingViews($this->getRequest()->getUrl()->getAbsoluteUrl());
+                } else {
+                    Notification::error(t('New group could not be added'));
```
sukhwinder33445, line 146: Use `Failed to create new contact group` instead, please.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+
+                        $this->addPart(...$update);
+                    }
+                }
+            })
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                $groupIdentifier = $form->addGroup();
+                if ($groupIdentifier) {
+                    Notification::success(t('New group has been successfully added'));
+                    $this->closeModalAndRefreshRemainingViews($this->getRequest()->getUrl()->getAbsoluteUrl());
```
sukhwinder33445, line 144: Please move it outside the conditions.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * Add attribute 'class' => 'full-width' if the content is an instance of BaseItemList
+     *
+     * @param ValidHtml $content
+     *
+     * @return ContactGroupsController
+     */
+    protected function addContent(ValidHtml $content): ContactGroupsController
```
sukhwinder33445, line 169: Please use self as return type.

library/Notifications/Widget/MemberSuggestions.php
```diff
@@ -0,0 +1,135 @@
...
+    /** @var string */
+    protected $searchTerm;
+
+    /** @var string */
+    protected $originalValue;
+
+    /** @var string[] */
+    protected $excludeTerms = [];
+
+    public function setSearchTerm(string $term): self
```
raviks789, line 29: This method is not used outside the class. Is there a reason, why this method is public?

library/Notifications/Widget/MemberSuggestions.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        $this->searchTerm = $term;
+
+        return $this;
+    }
+
+    public function setOriginalValue(string $term): self
```
raviks789, line 36: Same here.

library/Notifications/Widget/MemberSuggestions.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        $this->originalValue = $term;
+
+        return $this;
+    }
+
+    public function excludeTerms(array $terms): self
```
raviks789, line 43: Same here

raviks789 (changes requested):
IMO, you need to either disable the `Delete` button for contacts belonging to a contact group or show error notification why the contact cannot be deleted. Currently, you get the sql error, with error trace.

Edit: You could also check if the contact belongs to an escalation. Same goes for contact groups.

application/controllers/ContactGroupController.php
```diff
@@ -0,0 +1,125 @@
...
+        $this->addContent($form);
+    }
+
+    /**
+     * Add attribute 'class' => 'full-width' if the content is an instance of BaseItemList
+     *
+     * @param ValidHtml $content
+     *
+     * @return $this
+     */
```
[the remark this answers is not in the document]
sukhwinder33445: Yes, the PHPDoc is not required.

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,284 @@
...
+        if ($group === null) {
+            $this->httpNotFound(t('Contact group not found'));
+        }
+
+        $groupMembers = [];
+        foreach ($group->contact->columns('id') as $contact) {
+            $groupMembers[] = $contact->id;
+        }
+
+        $this->storedValues = [
```
nilmerg, line 180: Don't store this temporarily, as discussed. Load the stored data in the same transaction as when applying changes.

nilmerg (changes requested):

library/Notifications/Widget/ItemList/ContactListItem.php
```diff
@@ -43,7 +43,8 @@ protected function assembleVisual(BaseHtmlElement $visual): void
     protected function assembleFooter(BaseHtmlElement $footer): void
     {
         $contactIcons = new HtmlElement('div', Attributes::create(['class' => 'contact-icons']));
-        if ($this->item->has_email) {
+
+        if (! empty($this->item->has_email)) {
```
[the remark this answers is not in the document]
nilmerg: @sukhwinder33445 told me. Though, the isset that was previously there would have made this way more clear. Now, either restore the isset or explain in a comment why this check is required.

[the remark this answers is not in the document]
ncosta-ic: @nilmerg I didn't touch that class as you can see through my commits: https://github.com/Icinga/icinga-notifications-web/pull/188/files/d490cd52975ded7891f3e7ec06503fc124616d8d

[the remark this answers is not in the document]
nilmerg: Oh. My bad. But it's @sukhwinder33445's turn anyway ;)

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,284 @@
...
+                        'contactgroup_id' => $this->contactgroupId,
+                        'contact_id'      => $contactId
+                    ]
+                );
+            }
+
+            $isUpdated = true;
+        }
+
+        return $isUpdated;
```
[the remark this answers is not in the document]
sukhwinder33445: This is only used to display the success message when something has been updated.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,185 @@
...
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                $groupIdentifier = $form->addGroup();
+                if ($groupIdentifier) {
+                    Notification::success(t('New group has been successfully added'));
+                    $this->sendExtraUpdates(['#col1']);
+                    $this->getResponse()->setHeader('X-Icinga-Container', 'col2');
+                    $this->redirectNow(Links::contactGroup($groupIdentifier));
+                } else {
+                    Notification::error(t('Failed to add new contact group'));
+                    $this->switchToSingleColumnLayout();
```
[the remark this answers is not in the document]
sukhwinder33445: Yes, the form can be kept open.

But there is another problem now, when the form stays open, it contains the controls and shows a small bar. I think that controls should not be displayed in the modal and requires a css fix.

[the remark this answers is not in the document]
nilmerg: No, that's dependent on the `showCompact` and `disableLayout` params on the form's action

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,284 @@
...
+                'label'    => $this->translate('Members'),
+                'required' => true
+            ]
+        ))
+            ->setVerticalTermDirection()
+            ->setSuggestionUrl(Links::contactGroupsSuggestMember())
+            ->on(TermInput::ON_ENRICH, $callValidation)
+            ->on(TermInput::ON_ADD, $callValidation)
+            ->on(TermInput::ON_SAVE, $callValidation)
+            ->on(TermInput::ON_PASTE, $callValidation);
```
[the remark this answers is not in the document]
nilmerg: > php > var_dump(is_callable([new \DateTime(), 'getTimestamp']));
> bool(true)
> php >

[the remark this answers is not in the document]
nilmerg: Well. **That** is a reason. And probably why I wouldn't (and never did) use a method here.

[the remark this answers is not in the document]
sukhwinder33445: So you think I should leave it as it is?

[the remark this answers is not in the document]
ncosta-ic: > Yes, but it doesn't work with protected methods. We can use `Closure::fromCallable([$this, 'validateTerms'])` which returns a closure, but I don't think it's much better than what we're already doing.

A closure seems wrong as you are supposed to pass it an object of a callable type (see [Events.php#L31-34](https://github.com/Icinga/ipl-stdlib/blob/main/src/Events.php#L31-L34))

As for the issue that apparently arises when using a protected method, the [php documentation](https://www.php.net/manual/en/language.types.callable.php#language.types.callable.passing) states the following:
> ... Accessing protected and private methods from within a class is allowed.

Did you test it? According to the PHP documentation a `array($this, 'methodName')` or the shorter form `[$this, 'methodname']` should work :thinking:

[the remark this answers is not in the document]
sukhwinder33445: Yes, I have tested it. And it might work for the same class, but here we pass it to the `TermInput` instance.

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,300 @@
...
+                [
+                    'contactgroup_id' => $groupIdentifier,
+                    'contact_id'      => $contactId
+                ]
+            );
+        }
+
+        $this->db->commitTransaction();
+
+        return $groupIdentifier;
```
nilmerg, line 203: This cannot be null.

library/Notifications/Common/Links.php
nilmerg, line 1: The `showCompact` and `_disableLayout` are only required when using the links as modal targets. As such they should only be added if required. This means that adding them here is wrong. So either don't add those methods here or remove the default parameters.

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,184 @@
...
+            ->on(Form::ON_SUCCESS, function (ContactGroupForm $form) {
+                $groupIdentifier = $form->addGroup();
+                if ($groupIdentifier) {
+                    Notification::success(t('New group has been successfully added'));
+                    $this->sendExtraUpdates(['#col1']);
+                    $this->getResponse()->setHeader('X-Icinga-Container', 'col2');
+                    $this->redirectNow(Links::contactGroup($groupIdentifier));
+                } else {
+                    Notification::error(t('Failed to add new contact group'));
+                }
```
nilmerg, line 120: This is impossible to occur. Remove the conditional block entirely, it's not needed as the form implementation catches errors by default.

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,300 @@
...
+
+class ContactGroupForm extends CompatForm
+{
+    use CsrfCounterMeasure;
+
+    /** @var Connection */
+    private $db;
+
+    /** @var TermInput */
+    private $termInput;
```
nilmerg, line 29: obsolete

raviks789 (dismissed):

application/controllers/ContactGroupsController.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        $members = new MemberSuggestions();
+        $members->forRequest($this->getServerRequest());
+
+        $this->getDocument()->addHtml($members);
+    }
+
+    public function getTabs(): Tabs
+    {
+        if ($this->getRequest()->getActionName() === 'index') {
```
nilmerg, line 138: Why is this necessary? All other actions don't show tabs, so I don't see a reason for this

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,300 @@
...
+        $groupMembers = [];
+        foreach ($group->contact->columns('id') as $contact) {
+            $groupMembers[] = $contact->id;
+        }
+
+        return [
+            'group_name'   => $group->name,
+            'group_members' => implode(',', $groupMembers)
+        ];
+    }
```
nilmerg, line 299: Find the mistake:

image

nilmerg (changes requested):
style checks are failing

application/forms/ContactGroupForm.php
```diff
@@ -0,0 +1,299 @@
...
+        $values = $this->getValues();
+
+        $this->db->beginTransaction();
+
+        $storedValues = $this->fetchDbValues();
+
+        if ($values['group_name'] !== $storedValues['group_name']) {
+            $this->db->update(
+                'contactgroup',
+                ['name' => $values['group_name']],
```
nilmerg, line 220: > SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type bigint: "" CONTEXT: unnamed portal parameter $1 = ''

nilmerg (changes requested):
* [ ] Editing a previously empty group fails
* [ ] The control lines don't align: 
image

nilmerg (changes requested):
Adding a group without members doesn't work

nilmerg (approved):

PR #191, opened 2024-05-27
ContactController: Fix the retreival of contact full name during deletion

### LINKED ISSUE #190: Deleting a contact causes "Array to string conversion" error
Once a contact is deleted you get the error as shown in the screenshot instead of deleted message.

Screenshot 2024-05-27 at 10 25 45

raviks789:
fix #190

### DIFF
--- a/application/controllers/ContactController.php
+++ b/application/controllers/ContactController.php
@@ -54,7 +54,7 @@ class ContactController extends CompatController
                 $contactElement = $form->getElement('contact');
                 Notification::success(sprintf(
                     t('Deleted contact "%s" successfully'),
-                    $contactElement->getValue()
+                    $contactElement->getValue('full_name')
                 ));
 
                 $this->redirectNow('__CLOSE__');


application/controllers/ContactController.php
```diff
@@ -50,11 +50,9 @@ public function indexAction()
                 $this->redirectNow('__CLOSE__');
             })->on(ContactForm::ON_REMOVE, function (ContactForm $form) {
                 $form->removeContact();
-                /** @var FieldsetElement $contactElement */
-                $contactElement = $form->getElement('contact');
                 Notification::success(sprintf(
                     t('Deleted contact "%s" successfully'),
-                    $contactElement->getValue()
+                    $form->getValue('contact[full_name]')
```
sukhwinder33445, line 55: This cannot work.

sukhwinder33445: `$form->getValue('contact')['full_name']` should be used instead.

And please also add the change for `ON_SUCCESS`.

application/controllers/ContactController.php
```diff
@@ -40,21 +40,17 @@ public function indexAction()
             ->populate($contact)
             ->on(ContactForm::ON_SUCCESS, function (ContactForm $form) {
                 $form->addOrUpdateContact();
-                /** @var FieldsetElement $contactElement */
-                $contactElement = $form->getElement('contact');
                 Notification::success(sprintf(
                     t('Contact "%s" has successfully been saved'),
-                    $contactElement->getValue('full_name')
+                    $form->getValue('contact')['full_name']
```
nilmerg, line 45: There's absolutely no question that the previous case in the remove handler was wrong, but why is this changed?

My understanding is, and phpstan seems to agree, that this is now even more relying on implementation details than before. Yes, the full name is required, so the key always exists. But there's no way to express, even in the future, that the `contact` is a fieldset and has a field called `full_name` which holds a string.

I'd rather keep this as it was, and apply the same handling in the remove handler.

sukhwinder33445: That was my suggestion so that both pieces of code look the same. But yes, I can understand your point.

sukhwinder33445 (approved):

PR #213, opened 2024-06-20
Fix active list item handling

### LINKED ISSUE #212: Fix active list item handling
1. The active state of the list item is lost on page refresh 
2. When the object link in the list item is clicked, the list item does not change to active state

raviks789:
fix #212


library/Notifications/Widget/ItemList/ContactGroupList.php
```diff
@@ -14,4 +15,10 @@ protected function getItemClass(): string
     {
         return ContactGroupListItem::class;
     }
+
+    protected function init(): void
+    {
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/contact-group'));
```
ncosta-ic, line 22: I would not reintroduce a static url path and instead use the already existing reference in `Icinga\Module\Notifications\Common\Links` at line `89`.
This minifies errors for the case where the route definition could get changed in the future. It also helps with maintainability.
You also need to allow null values for use-cases where only the url reference is needed without any parameters, just like in this case.

A modification of that function could look like the following:
```php
/* add the ? modifier to the type definition, check for the existence of the id param, return the url without the id in case of its absence */
public static function contactGroup(?int $id): Url
{
    if ($id === null) {
        return Url::fromPath('notifications/contact-group');
    }

    return Url::fromPath('notifications/contact-group', ['id' => $id]);
}
```

library/Notifications/Widget/ItemList/ContactList.php
```diff
@@ -14,4 +15,10 @@ protected function getItemClass(): string
     {
         return ContactListItem::class;
     }
+
+    protected function init(): void
+    {
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/contact'));
```
ncosta-ic, line 22: Same change as for `ContactGroupList`.
Remember to introduce null values to `Links::contact()` as well.

library/Notifications/Widget/ItemList/EventList.php
```diff
@@ -37,6 +39,9 @@ protected function init(): void
         $this->on(self::ON_ASSEMBLED, function () {
             ObjectsRendererHook::load();
         });
+
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/event'));
```
ncosta-ic, line 44: Same change as for `ContactGroupList`.
Remember to introduce null values to `Links::event()` as well.

library/Notifications/Widget/ItemList/EventRuleList.php
```diff
@@ -17,4 +18,10 @@ protected function getItemClass(): string
     {
         return EventRuleListItem::class;
     }
+
+    protected function init(): void
+    {
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/event-rule'));
```
ncosta-ic, line 25: Same change as for `ContactGroupList`.
Remember to introduce null values to `Links::eventRule()` as well.

library/Notifications/Widget/ItemList/IncidentList.php
```diff
@@ -24,6 +25,9 @@ protected function init(): void
         $this->on(self::ON_ASSEMBLED, function () {
             ObjectsRendererHook::load();
         });
+
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/incident'));
```
ncosta-ic, line 30: Same change as for `ContactGroupList`.
Remember to introduce null values to `Links::incident()` as well.

library/Notifications/Widget/ItemList/ScheduleList.php
```diff
@@ -14,4 +15,10 @@ protected function getItemClass(): string
     {
         return ScheduleListItem::class;
     }
+
+    protected function init(): void
+    {
+        $this->getAttributes()
+            ->set('data-icinga-detail-url', (string) Url::fromPath('notifications/schedule'));
```
ncosta-ic, line 22: Same change as for `ContactGroupList`.
Remember to introduce null values to `Links::schedule()` as well.

ncosta-ic (changes requested):
We should stay consistent and use `Icinga\Module\Notifications\Common\Links` for path references to strengthen the maintainability.

ncosta-ic (approved):
Looks good now! :)

ncosta-ic (dismissed):

library/Notifications/Widget/ItemList/ContactGroupListItem.php
```diff
@@ -23,7 +25,9 @@ class ContactGroupListItem extends BaseListItem
 
     protected function init(): void
     {
-        $this->getAttributes()->set('data-action-item', true);
+        $this->getAttributes()
+            ->set('data-action-item', true)
+            ->set('data-icinga-detail-filter', QueryString::render(Filter::equal('id', $this->item->id)));
```
nilmerg, line 30: > We should stay consistent and use Icinga\Module\Notifications\Common\Links for path references to strengthen the maintainability.

I'd extend this to query param references:

`Links::contactGroup($this->item->id)->getQueryString()`

nilmerg (changes requested):

nilmerg:
Requires a rebase.

PR #230, opened 2024-07-15
Do not store null for channel options

nilmerg:


oxzi (approved):
I have tested this PR together with the current version of <https://github.com/Icinga/icinga-notifications/pull/206> and they worked perfectly together. Thanks!

PR #231, opened 2024-07-16
ObjectSuggestions: Fix exotic columns match

sukhwinder33445:
These columns should only be shown if explicitly given by user. Previously, columns with a similar suffix were also found, e.g. the search for `id` also showed `uuid` as search suggestion.

### DIFF
--- a/library/Notifications/Web/Control/SearchBar/ObjectSuggestions.php
+++ b/library/Notifications/Web/Control/SearchBar/ObjectSuggestions.php
@@ -232,9 +232,8 @@ class ObjectSuggestions extends Suggestions
     protected function matchSuggestion($path, $label, $searchTerm)
     {
         if (preg_match('/[_.](id)$/', $path)) {
-            // Only suggest exotic columns if the user knows about them
-            $trimmedSearch = trim($searchTerm, ' *');
-            return substr($path, -strlen($trimmedSearch)) === $trimmedSearch;
+            // Only suggest exotic columns if the user knows the full column path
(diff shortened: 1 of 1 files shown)


library/Notifications/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -231,10 +231,9 @@ protected function queryTags(Model $model, string $searchTerm): Query
 
     protected function matchSuggestion($path, $label, $searchTerm)
     {
-        if (preg_match('/[_.](id)$/', $path)) {
+        if (preg_match('/[_.](id)$/', $path, $matches)) {
             // Only suggest exotic columns if the user knows about them
-            $trimmedSearch = trim($searchTerm, ' *');
-            return substr($path, -strlen($trimmedSearch)) === $trimmedSearch;
+            return $matches[1] === trim($searchTerm, ' *');
```
nilmerg, line 236: This makes it too easy to find such columns. The previous intent was to only show these if the user knows their exact path.

The previous condition however was indeed incorrect. I've attempted to fix this myself earlier and came up with this instead:

```
return substr($path, strrpos($path, '.') + 1) === $trimmedSearch;
```

Though, this was only a quick fix. Please check if this matches the original intent better.

nilmerg (changes requested):

library/Notifications/Web/Control/SearchBar/ObjectSuggestions.php
sukhwinder33445: This works fine.

nilmerg (approved):

PR #237, opened 2024-07-18
js: Fix operator precedence issues

nilmerg:


public/js/notifications.js
```diff
@@ -8,25 +8,25 @@ const VERSION = 1;
...
 
         return;
     }
 
-    if (! navigator.serviceWorker) {
+    if (! ('serviceWorker' in navigator)) {
         console.error(LOG_PREFIX + "this browser does not support the 'Service Worker API' in the current context");
 
         return;
     }
```
nilmerg, line 27: This seems obsolete now, since `! ('ServiceWorker' in self)` is fixed and matches.

ncosta-ic (approved):
Seems fine. I also double-checked the conditions on my Firefox instance (both, http and https) and the checks return the correct values.

PR #241, opened 2024-07-19
Fix hook error handling

nilmerg:


yhabteab (approved):
Works fine now, but the action lists are no longer clickable without Icinga DB Web.

nilmerg:
That's a known limitation, and already being worked on: https://github.com/Icinga/ipl-web/pull/225

PR #244, opened 2024-07-22
Don't limit browser_session.user_agent to 4096 chars

Al2Klimov:
* This isn't necessary anymore due to https://github.com/Icinga/icinga-notifications/pull/239

### DIFF
--- a/library/Notifications/Daemon/Server.php
+++ b/library/Notifications/Daemon/Server.php
@@ -406,8 +406,7 @@ class Server
 
             if ($browserSession !== null) {
                 if (isset($headers['User-Agent'][0])) {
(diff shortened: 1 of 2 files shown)


julianbrost:
> This isn't necessary anymore

IMHO that doesn't mean that this has to be removed. Having a reasonable limit still sounds like a good idea to me to prevent the database from filling up with too much garbage. Removing the hard-coded limit in the schema just means that Icinga Notifications Web is now able to more freely adjust that limit.

Al2Klimov:
Theoretically a UA could be 7.9KB, given that Apache 2 allows 8KB headers.

nilmerg (approved):

PR #248, opened 2024-07-24
FilterRenderer: Add missing return type to ensure compatibility with the parent class

sukhwinder33445:


nilmerg:
Didn't saw this, and pushed to main directly :sweat_smile:

PR #286, opened 2025-01-29
IncidentDetail: Don't render `Object Tags` section if no tags are available

### LINKED ISSUE #270: Don't render the `Object Tags` section if there's none
Displaying the `Object Tags` section in the incident detail view without extra tags looks odd, and I suggest not to display it at all or render an empty state widget with a proper description.

Bildschirmfoto 2024-08-16 um 15 27 48

sukhwinder33445:
resolves #270

### DIFF
--- a/library/Notifications/Widget/Detail/IncidentDetail.php
+++ b/library/Notifications/Widget/Detail/IncidentDetail.php
@@ -92,7 +92,7 @@ class IncidentDetail extends BaseHtmlElement
         ];
     }
 
-    protected function createObjectTag()
+    protected function createObjectTag(): array
     {
         $tags = [];
         foreach ($this->incident->object->object_extra_tag as $extraTag) {
@@ -104,7 +104,7 @@ class IncidentDetail extends BaseHtmlElement
         }
 
         return [
-            Html::tag('h2', t('Object Tags')),
+            new HtmlElement('h2', null, new Text(t('Object Tags'))),
             (new Table())
                 ->addHtml(...$tags)
                 ->addAttributes(['class' => 'object-tags-table'])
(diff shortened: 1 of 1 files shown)


library/Notifications/Widget/Detail/IncidentDetail.php
```diff
@@ -92,17 +92,22 @@ protected function createSource()
...
     {
-        $objectTags = (new Table())->addAttributes(['class' => 'object-tags-table']);
-
+        $tags = [];
         foreach ($this->incident->object->object_extra_tag as $extraTag) {
-            $objectTags->addHtml(Table::row([$extraTag->tag, $extraTag->value]));
+            $tags[] = Table::row([$extraTag->tag, $extraTag->value]);
+        }
+
+        if (empty($tags)) {
```
ncosta-ic, line 102: There's no need for the `empty()` statement, as this checks for variable allocation *and* falsey variable content.

A simple `if (! $tags)` would be sufficient, because you guarantee its existence on line `97`.
Or is there any reason other than legibility as to why you've opted for this?

library/Notifications/Widget/Detail/IncidentDetail.php
```diff
@@ -92,17 +92,22 @@ protected function createSource()
...
+            return $tags;
         }
 
         return [
-            Html::tag('h2', t('Object Tags')),
-            $objectTags
+            new HtmlElement('h2', null,  new Text(t('Object Tags'))),
```
ncosta-ic, line 107: There's a double space after `null,`, please reformat

ncosta-ic (changes requested):

ncosta-ic (approved):
Looks good to me

PR #289, opened 2025-01-30
ChannelForm: Type bool (checkbox) option must be stored as bool in db

### LINKED ISSUE #267: Channel Plugin Config Option of type Boolean is stored as a string in the database
After creating a channel plugin with the following shortened `config_attrs` value in the `available_channel_type` table, the stored `config` in the `channels` table looks like this.

-  `available_channel_type.config_attrs`:
```
[
  {
    "name": "tls_insecure",
    "type": "bool",
    "label": {
      "de_DE": "Keine TLS-Verifizierung",
      "en_US": "No TLS Verification"
    },
    "help": {
      "de_DE": "Führe keine TLS-Verifizierung durch. Dies vermag unsicher zu sein.",
      "en_US": "Skip TLS verification. This might be insecure."
    },
    "default": false,
    "min": null,
    "max": null
  }
]
 ```
- `config.channels`:
```
{
  "tls_insecure": "n"
}
```

The boolean value will be stored as a string of either `"y"` or `"n"` instead of an expected JSON boolean.

### PR DESCRIPTION
**Pull request by sukhwinder33445**
fixes #267

### DIFF
--- a/application/forms/ChannelForm.php
+++ b/application/forms/ChannelForm.php
@@ -334,6 +334,11 @@ class ChannelForm extends CompatForm
             'label' => $this->fromCurrentLocale($elementConfig['label'])
         ];
 
+        if ($elementConfig['type'] === 'bool') {
+            $options['checkedValue'] = 'checked';
+            $options['uncheckedValue'] = 'unchecked';
+        }
+
         if (isset($elementConfig['help'])) {
             $options['description'] = $this->fromCurrentLocale($elementConfig['help']);
         }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: application/forms/ChannelForm.php [hunk 1]
```diff
@@ -398,6 +403,10 @@ protected function fromCurrentLocale(array $localeMap): ?string
     private function filterConfig(array $config): array
     {
         foreach ($config as $key => $value) {
+            if (in_array($value, ['checked', 'unchecked'])) {
```
ncosta-ic, line 406: One could argue that you should always use in_array's `strict` flag, since the value that is getting checked for always corresponds to a string type. It would not cause any issues in this instance, as there's no possibility for the value to get checked against `null, true, false`, but it's considered best-practice nonetheless.
[An example where `in_array()` would return bogus](https://onlinephp.io/c/9e49adb8-6417-4b54-b6d6-da6718244008)

PR #290, opened 2025-01-30
ChannelForm: Show message intead of stacktrace when no channel type found

### LINKED ISSUE #225: Module Channel Configuration: No Available Channels results in Stack Trace for Add Channel
An empty `available_channel_type` table results in a stack trace when one tries to add a new channel. This might happen if the daemon was not started yet.

To reproduce, start with a "clean" database or optionally delete all entries from `available_channel_type`. Start your clicking journey in Icinga Web from the configuration gear in the sidebar, go to _Modules_, select the _notifications_ module, go to the _Channels_ tab and _Add Channel_.

```
Undefined array key ""

#0 /usr/share/icingaweb2-modules/notifications/application/forms/ChannelForm.php(103): Icinga\Application\ApplicationBootstrap->Icinga\Application\{closure}(Integer, String, String, Integer)
#1 /usr/local/src/ipl-html/src/HtmlDocument.php(344): Icinga\Module\Notifications\Forms\ChannelForm->assemble()
#2 /usr/local/src/ipl-html/src/Form.php(210): ipl\Html\HtmlDocument->ensureAssembled()
#3 /usr/share/icingaweb2-modules/notifications/application/controllers/ChannelsController.php(117): ipl\Html\Form->handleRequest(Object(GuzzleHttp\Psr7\ServerRequest))
#4 /usr/share/icinga-php/vendor/vendor/shardj/zf1-future/library/Zend/Controller/Action.php(522): Icinga\Module\Notifications\Controllers\ChannelsController->addAction()
#5 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch(String)
#6 /usr/share/icinga-php/vendor/vendor/shardj/zf1-future/library/Zend/Controller/Front.php(954): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#7 /icingaweb2/library/Icinga/Application/Web.php(294): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#8 /icingaweb2/library/Icinga/Application/webrouter.php(105): Icinga\Application\Web->dispatch()
#9 /icingaweb2/public/index.php(4): require_once(String)
#10 {main}
```

sukhwinder33445:
fixes #225 

### Requires:
- https://github.com/Icinga/icingaweb2/pull/5370
- https://github.com/Icinga/ipl-web/pull/284

### DIFF
--- a/application/controllers/ChannelsController.php
+++ b/application/controllers/ChannelsController.php
@@ -15,7 +15,6 @@ use Icinga\Module\Notifications\Widget\ItemList\ObjectList;
 use Icinga\Web\Notification;
 use Icinga\Web\Widget\Tab;
 use Icinga\Web\Widget\Tabs;
-use ipl\Sql\Connection;
 use ipl\Sql\Expression;
 use ipl\Stdlib\Filter;
 use ipl\Web\Compat\CompatController;
@@ -30,22 +29,17 @@ class ChannelsController extends CompatController
 {
     use SearchControls;
 
-    /** @var Connection */
-    private $db;
-
     /** @var Filter\Rule Filter from query string parameters */
     private $filter;
 
-    public function init()
+    public function init(): void
     {
         $this->assertPermission('config/modules');
-
-        $this->db = Database::get();
     }
 
-    public function indexAction()
+    public function indexAction(): void
     {
-        $channels = Channel::on($this->db);
+        $channels = Channel::on(Database::get());
         $this->mergeTabs($this->Module()->getConfigTabs());
         $this->getTabs()->activate('channels');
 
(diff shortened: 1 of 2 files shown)


ncosta-ic:
@sukhwinder33445 I feel like it looks kinda weird with the current implementation, as it still shows the name input, even though there's no submit button or the like.

What if you moved the result check to the top of the assemble method? I think this would correlate with our philosophy of having early breakouts if required data is not provided to a method.

It currently looks like this:
image

And adding the check and early return to the top of the method would result in this representation:
image

The displayed message is also missing a dot (`.`) at the end of the second sentence.

Keep in mind that I only wondered about the reasoning for this exact implementation, this isn't really a suggestion. But if you do feel like it would improve the UI, feel free to change it. Will do a proper review after your comment on this topic :)

nilmerg:
Our actual philosophy on this topic is not allowing access to a form at all. i.e. to disable the button that leads to it and adding a title to explain why.

nilmerg:
…or to hide the button but still explain why if there is no contextual hint.

PR #292, opened 2025-02-06
Schedules: Add timescale and time indicator

### LINKED ISSUE #283: Clarify temporal direction of the timeline
## Goal 
 
* Add a timescale at the bottom, below the result
* Show an indicator for the current time

## Implementation Hint

The parent issue includes mockups how both should look like.

Note that the day mode has enough width to show a more detailed timescale.

sukhwinder33445:
resolves #283

### DIFF
--- a/library/Notifications/Widget/TimeGrid/Timescale.php
+++ b/library/Notifications/Widget/TimeGrid/Timescale.php
@@ -1,0 +1,87 @@
+<?php
+
+/* Icinga Notifications Web | (c) 2025 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Notifications\Widget\TimeGrid;
+
+use DateTime;
+use IntlDateFormatter;
+use ipl\Html\Attributes;
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\I18n\Translation;
+use ipl\Web\Style;
+use Locale;
+
+/**
+ * Creates a localized timescale for the TimeGrid
+ */
+class Timescale extends BaseHtmlElement
+{
+    use Translation;
+
+    protected $tag = 'div';
+
+    protected $defaultAttributes = ['class' => 'timescale'];
+
+    /** @var int The number of days shown */
+    protected $days;
+
+    /** @var Style */
+    protected $style;
+
+    /**
+     * Create a new Timescale
+     *
+     * @param int $days
+     * @param Style $style
+     */
+    public function __construct(int $days, Style $style)
+    {
+        $this->days = $days;
+        $this->style = $style;
+    }
+
+    public function assemble(): void
+    {
+        if ($this->days === 1) {
+            $timestampPerDay = 12;
+        } elseif ($this->days <= 7) {
+            $timestampPerDay = 2;
+        } else {
+            $timestampPerDay = 1;
+        }
+
+        $this->style->addFor($this, ['--timestampsPerDay' => $timestampPerDay * 2]); // *2 for .ticks
+
+        $dateFormatter = new IntlDateFormatter(
+            Locale::getDefault(),
+            IntlDateFormatter::NONE,
+            IntlDateFormatter::SHORT
+        );
+
+        $timeIntervals = 24 / $timestampPerDay;
+
+        $time = new DateTime();
+        $dayTimestamps = [];
+        for ($i = 0; $i < $timestampPerDay; $i++) {
+            // am-pm is separated by non-breaking whitespace
+            $parts = preg_split('/\s/u', $dateFormatter->format($time->setTime($i * $timeIntervals, 0)));
+
+            $stamp = [new HtmlElement('span', null, new Text($parts[0]))];
+            if (isset($parts[1])) {
+                $stamp[] = new HtmlElement('span', null, new Text($parts[1]));
+            }
+
+            $dayTimestamps[] = new HtmlElement('span', new Attributes(['class' => 'timestamp']), ...$stamp);
+            $dayTimestamps[] = new HtmlElement('span', new Attributes(['class' => 'ticks']));
+        }
+
+        $allTimestamps = array_merge(...array_fill(0, $this->days, $dayTimestamps));
+        // clone is required because $allTimestamps contains references of same object
+        $allTimestamps[] = (clone $allTimestamps[0])->addAttributes(['class' => 'midnight']); // extra stamp of 12AM
+
+        $this->addHtml(...$allTimestamps);
+    }
+}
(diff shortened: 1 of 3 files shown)


library/Notifications/Widget/Timeline.php
```diff
@@ -312,6 +316,42 @@ protected function assemble()
...
+            );
+
+            $now = new DateTime();
+            $currentTime = new HtmlElement(
+                'div',
+                new Attributes(['class' => 'time-hand']),
+                new HtmlElement(
+                    'div',
+                    new Attributes(['class' => 'now', 'title' => $dateFormatter->format($now)]),
+                    Text::create('now')
```
ncosta-ic, line 333: This is missing a translation

ncosta-ic (changes requested):
Additional note: Your PR introduces a vertical scrollbar. This is probably caused by your translate rules. Please recheck your code and fix it.

ncosta-ic:
@sukhwinder33445 I also noticed that something's off in Firefox. While the timeline and grid render correctly, the timescale's off. It's especially noticeable when using the monthly view.
I can't test it on Safari, but you might want to double-check that one as well.
difference_chrome_firefox

library/Notifications/Widget/TimeGrid/Timescale.php
```diff
@@ -0,0 +1,96 @@
...
+            IntlDateFormatter::SHORT
+        );
+
+        $timeIntervals = 24 / $timestampPerDay;
+
+        $time = new DateTime();
+        $dayTimestamps = [];
+        for ($i = 0; $i < $timestampPerDay; $i++) {
+            // am-pm is separated by non-breaking whitespace
+            $parts = preg_split('/\s/u', $dateFormatter->format($time->setTime($i * $timeIntervals, 0)));
```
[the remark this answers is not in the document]
sukhwinder33445: `explode` does not work here. I have already tried your suggestions.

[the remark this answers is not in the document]
ncosta-ic: That one goes on me. I looked at the wrong place. The formats are shipped in https://unicode.org/Public/cldr/42/ and in version 42, a lot of locales changed their ASCII space to a narrow non-breaking space (`\u20F2`)[^1].
It does work with a `\u{20F2}`, but this is not fool-proof against all locales, albeit the ones that I've checked all used this separator for the (`... a` date format).

This means your solutions would be more future-proof, in case they swap it onto another whitespace character.

[^1]: https://icu.unicode.org/download/72#h.u2dtz3f7ik9a , "In many formatting patterns, ASCII spaces are replaced with Unicode spaces (e.g., a "thin space")."

flourish86 (changes requested):
Looks fine so far.

There're one thing I noticed, though.

The cursor is inapporpriate in this case, since there's nothing happening on click.
Frame 110

The scale labels still look cluttered especially in month and 2 weeks mode. But we should postpone the refinement to another time.

public/css/timeline.less
```diff
@@ -134,6 +209,18 @@
...
+  .timescale .timestamp {
+    color: @gray-semilight;
+    background: @body-bg-color;
+  }
+
+  .clock .now {
+    background-color: @gray-light;
+    font-size: 0.75em;
+    color: red;
+    user-select: none;
```
flourish86, line 222: I found, that it won't work on Safari and a Quick Look [here](https://caniuse.com/user-select-none) shows, that you probably need the prefixes for all browsers to make this work.

So I'd suggest to add
```
-webkit-user-select: none;
-moz-user-select: none;
```
and optionally 
```
-ms-user-select: none;
```

sukhwinder33445: Ahmm, fortunately we have a mixin for that.

sukhwinder33445: Safari disable the selecting, but still shows the `caret` cursor :facepalm:.

flourish86 (approved):

ncosta-ic (approved):
LGTM

public/css/timeline.less
```diff
@@ -65,9 +70,86 @@
...
+
+      span:nth-last-of-type(2), // last .ticks before .midnight
+      .midnight {
+        grid-area: ~"1 / -2 / 1 / -1";
+      }
+    }
+
+    .clock {
+      display: grid;
+      grid-template-columns: repeat(var(--primaryColumns), 1fr);
```
nilmerg, line 129: Same orientation, same rules. (As e.g. the header)

```suggestion
      grid-template-columns: repeat(var(--primaryColumns), minmax(var(--minimumPrimaryColumnWidth), 1fr));
```

public/css/timeline.less
```diff
@@ -65,9 +70,86 @@
...
+      grid-area: ~"3 / 2 / 4 / 3";
+      border-top: 1px solid transparent; // left not required, otherwise the .time-hand is not aligned properly
+
+      .current-day {
+        display: grid;
+        grid-template-columns: repeat(var(--columnsPerStep), 1fr);
```
nilmerg, line 135:
```suggestion
        grid-template-columns: repeat(var(--columnsPerStep), minmax(var(--minimumStepColumnWidth), 1fr));
```

public/css/timeline.less
```diff
@@ -65,9 +70,86 @@
...
-      bottom: var(--stepRowHeight);
       right: 0;
       left: 0;
+      // -1 to exclude result row
+      top: ~"calc((var(--stepRowHeight) * calc(var(--primaryRows) - 1)) + var(--daysHeaderHeight))";
+    }
+
+    .timescale {
+      display: grid;
+      grid-template-columns: repeat(~"calc(var(--primaryColumns) * var(--timestampsPerDay))", minmax(0,1fr));
```
nilmerg, line 81:
```suggestion
      grid-template-columns: repeat(~"calc(var(--primaryColumns) * var(--timestampsPerDay))", minmax(calc(var(--minimumPrimaryColumnWidth) / var(--timestampsPerDay)),1fr));
```

public/css/timeline.less
```diff
@@ -65,9 +70,86 @@
...
+          left: 50%;
+        }
+
+        > span:last-child {
+          opacity: 0.5;
+        }
+      }
+
+      span:nth-last-of-type(2), // last .ticks before .midnight
+      .midnight {
```
nilmerg, line 122: You should apply such rules as strict as possible. At the moment, this bleeds on to the children of each timestamp as well, albeit with no effect.

nilmerg (commented):
I have difficulties to read the scale's values, let alone seeing the ticks!

At the moment though, this is not the only problem. All modes but the day mode do not benefit of this at all. The chronological orientation is already visible in the header. Though, it only shows the differing days. The scale just swaps this to hours but only shows midday or midnight, which just doesn't add anything useful.

The original intention to clarify the direction of the timeline is not properly addressed here. The scale does not add a visual component but still relies on the ability of the user to comprehend the textual representation.

If the modes still cannot benefit of the scale once its values are readable (larger) because it prevents more steps to be shown, we should not show it at all. Only the day mode may still benefit, so let's keep it there.

For the other modes, a visual representation is needed. Please discuss this with @flourish86.

nilmerg:
> The original intention to clarify the direction of the timeline is not properly addressed here. The scale does not add a visual component but still relies on the ability of the user to comprehend the textual representation.

We've concluded with the result, that the font size must be scaled up. Nothing else changes.

PR #295, opened 2025-02-12
RotationConfigForm: Add new rotation at the bottom of the list

### LINKED ISSUE #278: Additional rotations should be appended to the list at the bottom
Adding a new rotation to a schedule inserts it at the top of the list right now.

## Goal

It should be appended at the bottom instead. This is not about changing the location of the button, though!

## Implementation Hints

Rotations have a priority assigned to them. There should already be a way to preserve these and only re-index only what needs to. Though, adding a new rotation and keeping it at the bottom then has the *expected* side-effect that all rotations of this schedule need to be re-indexed.

sukhwinder33445:
resolves #278

### DIFF
--- a/application/forms/RotationConfigForm.php
+++ b/application/forms/RotationConfigForm.php
@@ -311,6 +311,20 @@ class RotationConfigForm extends CompatForm
         }
 
         $changedAt = time() * 1000;
+
+        $rotationsToMove = Rotation::on($this->db)
+            ->columns('id')
+            ->filter(Filter::equal('schedule_id', $this->scheduleId))
+            ->orderBy('priority', SORT_DESC);
+
+        foreach ($rotationsToMove as $rotation) {
+            $this->db->update(
+                'rotation',
+                ['priority' => new Expression('priority + 1'), 'changed_at' => $changedAt],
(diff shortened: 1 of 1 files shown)


application/forms/RotationConfigForm.php
```diff
@@ -311,6 +311,19 @@ private function createRotation(int $priority): Generator
         }
 
         $changedAt = time() * 1000;
+
+        $RotationsToMove = Rotation::on($this->db)
```
ncosta-ic, line 315: Please start lowercased for camelCased variable names

ncosta-ic (changes requested):
This whole PR does not work properly at all. Just add multiple rotations with the same "First Handoff" and you'll get unique constraint violations.
It also completely broke my existing entries when I deleted one of the newly added rotations (through the UI).

It looks like one of the problems lies in the fact, that your newly added `foreach` loop updates all entries on that specific schedule and upps their priority by one.
When incrementing the rotation priorities, you have to start with the biggest priority and work your way down, as there can't exist two rotations with the same first handoff dates and the same priority.

ncosta-ic (approved):
Looks and works fine now :)

PR #296, opened 2025-02-12
Make rotation title clickable and editable

### LINKED ISSUE #279: Rotation titles should be clickable and allow to edit them
At the moment a rotation can only be edited by clicking on a occurrence in the timeline.

## Goal

In addition to that, a button with the cog icon should be shown in the side bar. Possibly besides the drag handle.

## Implementation Hints

The interaction area of the button somewhat collides with that of the drag handle. A good compromise is required for both.

Also, remember that a title might be long enough to occupy most of the cell. Because of this, the interaction area of the drag handle and button might be static and relatively small. (Variable interaction areas are somewhat irritating.)

The mockups in the parent issue include this change.

sukhwinder33445:
resolves #279

### DIFF
--- a/public/css/timeline.less
+++ b/public/css/timeline.less
@@ -80,13 +80,14 @@
   .rotation-name {
     font-size: 1.25em;
     font-weight: bold;
-    cursor: grab;
 
     [data-drag-initiator] {
       color: @gray-light;
-    }
-    &:hover [data-drag-initiator] {
-      color: @text-color;
+
+      &:hover {
+        cursor: grab;
+        color: @list-item-title-hover-color;
+      }
     }
 
     // :not(:empty) is only necessary to increase specificity
--- a/public/css/timeline.less
+++ b/public/css/timeline.less
@@ -28,6 +28,20 @@
         margin-left: .5em;
         .text-ellipsis();
       }
+
+      > a {
+        display: flex;
+        width: 100%;
+        align-items: baseline;
+        margin-left: .5em;
+
+        span {
+          width: 0;
+          flex: 1 0 auto;
+          margin-right: 0.25em;
+          .text-ellipsis();
+        }
+      }
     }
 
     .overlay .entry {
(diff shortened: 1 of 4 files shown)


public/js/schedule.js
```diff
@@ -41,7 +41,8 @@
             Sortable.create(sideBar, {
                 scroll: true,
                 direction: 'vertical',
-                draggable: '.rotation-name'
+                draggable: '.rotation-name',
+                handle: ".rotation-name > i[data-drag-initiator]",
```
ncosta-ic, line 45: There's an unnecessary decimal point at the end

ncosta-ic (changes requested):

ncosta-ic (approved):

public/css/timeline.less
```diff
@@ -28,6 +28,20 @@
...
+        width: 100%;
+        align-items: baseline;
+        margin-left: .5em;
+
+        span {
+          width: 0;
+          flex: 1 0 auto;
+          margin-right: 0.25em;
+          .text-ellipsis();
+        }
```
nilmerg, line 43: The `> span` rule above was the basis for this I suppose. Why is it still there though? The timeline now always renders a link in the sidebar.

public/css/timeline.less
```diff
@@ -97,6 +112,22 @@
...
+        display: none;
+      }
+
+      &:hover {
+        color: @list-item-title-hover-color;
+
+        i {
+          display: inline-block;
+        }
+      }
```
nilmerg, line 129: The fact that `i` uses `inline-block` by default, is only the case if it is also an `.icon`. This isn't addressed here.

I'd suggest to do this instead:

```
&:not(:hover) .icon {
  display: none;
}
```

This way, it resets to the default display value without the need to repeat it here. And it doesn't rely on the fact that an `ipl\Web\Icon` uses `i` as tag.

nilmerg:
@sukhwinder33445

PR #297, opened 2025-02-13
SchedulesController::getTabs(): Add missing `baseTagrget` to tabs

### LINKED ISSUE #282: Inconsistent Column Behavior
I noticed, that the two column layout behaves inconsistent when a Schedule Detail View is opened in the second column.

But we might also reconsider how columns behave, when Schedule or Event Rule Detail are opened, since the editors take a lot of horizontal space.

https://github.com/user-attachments/assets/fc3d0108-b874-4cea-9e58-9a23935a014f

sukhwinder33445:
fixes #282


ncosta-ic (approved):
Looks good to me as it respects https://github.com/Icinga/icinga-notifications-web/issues/282#issuecomment-2606685043

PR #302, opened 2025-02-27
Improve rotation mode thumbnails

### LINKED ISSUE #291: Improve Rotation Mode Thumbnails
User seem to struggle to understand what the different rotation modes mean. They can't derive how the rotation modes work from the preview thumbnails.

* users expect the timeline of the placeholder entries in the preview thumbnails are vertical
* users interpret placeholders for days of the week in the headers look like whole day entries

sukhwinder33445:
resolves #291


flourish86 (commented):
The thumbnails by themselves look great so far. I'm not sure, if this is for a different issue, but the line-breaks are now a bit weird, which makes it look a little messy.

I'd consider to adjust the width of the `li`s.

Something like this should do the trick:

```
.rotation-mode ul label {
   width: auto
}

.rotation-mode ul li {
    flex: 1 1 auto;
    width: 0;
}

.rotation-mode li:not(:last-child) {
    margin-right: 1em;
}
```

flourish86 (approved):
Looks good!

PR #305, opened 2025-03-06
Getting started with schedules

### LINKED ISSUE #280: Help users getting started with schedules
Users seem to struggle to…

* …notice that there are multiple rotations possible
* …notice where to add the first rotation

## Goal

As long as there are no rotations, the timeline should not be visible. Instead, the view makes it clear what can be done at first.

@flourish86 will provide a mockup for this.

sukhwinder33445:
resolves #280

### DIFF
--- a/public/js/schedule.js
+++ b/public/js/schedule.js
@@ -20,6 +20,11 @@
             this.on('mouseleave', '#notifications-schedule .entry', this.onEntryLeave, this);
         }
 
+        /**
+         * Make the sidebar sortable and add drag&drop support.
+         *
+         * @param event The event object.
+         */
         onRendered(event)
         {
             if (event.target !== event.currentTarget) {
@@ -46,6 +51,11 @@
             });
         }
 
+        /**
+         * Handle drop event on the sidebar.
+         *
+         * @param event The event object.
+         */
         onDrop(event)
         {
             event = event.originalEvent;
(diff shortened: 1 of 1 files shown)


flourish86 (changes requested):
This looks basically good already. I found some smaller issues, though.

1) The layout of the empty state notice is broken when becoming narrow. I'd probably opt for a flex box solution.

Screen Shot 2025-03-17 at 15 34 07

2) I find the timestamps for the scale pretty distracting, especially when the columns width becomes really narrow. We had this offline already and I see the points you made about the locale stuff. But can we probably try to decrease the line-height and fade the AM/PMs?

Something like this.

```
.timescale .timestamp {
   line-height: 1;
}

.timescale .timestamp > span:last-child {
    opacity: 0.5;
}
```

3) Can you adjust the background of the empty-state button container? It's a little too dark. I noticed, that I used a custom value. It's #d9d9d9 in the mockup, but with 25% opacity. Probably @gray-lighter would be close, but also with 25%. I think you can use `fade(@gray-light, 25%)`.

4) Same for the "Add another …" Button. And the row header should be the same colors, please.
Screenshot 2025-03-17 at 16 14 52

5) Not sure if it's addressed in a different issue, but the row header should be the same as the body row area, when highlighted.
Frame 36

public/css/timeline.less
```diff
@@ -210,6 +260,25 @@
     }
   }
 
+  .btn-container {
+    background-color: mix(@gray-light, @body-bg-color, 25%);
+    margin-left: ~"calc(var(--sidebarWidth) * -1)";
```
flourish86, line 265: You added the negative margin here. To compensate for this you can add the same amount of padding-left and then the content will be centered.

`padding-left: ~"calc(var(--sidebarWidth))"; // short for ~"calc(var(--sidebarWidth) * -1) * -1"`

flourish86 (changes requested):
The "Add another rotation" version looks great.

In the empty state version, the content is not centered, though.

public/css/timeline.less
flourish86: Sorry, my fault. That suggestion doesn't make sense in the broader context. Please add the negativer margin only for `time-grid:not(.empty) and leave the padding-left out.

flourish86 (approved):

nilmerg:
Sorry, this touched so much irrelevant code parts I rewrote it from scratch in #359

Except the hover/highlight stuff, please open a separate PR for this.

Issue #226, opened 2024-07-11
Incremental config update limitations


sukhwinder33445:
- [ ] Event rule
- Removing an event rule escalation causes the `changed_at` flag to be updated for all other escalations.
- Adding/removing/updating a recipient causes the escalation and all its recipients to have a new `changed_at`.
- [ ] The `changed_at` column
- It should be updated at a central point (Database.php) to avoid future errors.
- It should be precise to milliseconds in order to avoid a data race (daemon side), e.g. if two users update the data at the same time.

nilmerg:
> It should be precise to milliseconds in order to avoid a data race (daemon side), e.g. if two users update the data at the same time.

If two users update the same configuration at the same time, one should get a conflict and need to re-submit. This is no reason to increase precision on the timestamp values.

nilmerg:
> > It should be precise to milliseconds in order to avoid a data race (daemon side), e.g. if two users update the data at the same time.
> 
> If two users update the same configuration at the same time, one should get a conflict and need to re-submit. This is no reason to increase precision on the timestamp values.

Though, in case it's not possible to differentiate between new and updated entries (which is the case), it is...

PR #306, opened 2025-03-10
Don't cut off milliseconds from current-time-in-milliseconds

Al2Klimov:
* Re: https://github.com/Icinga/icinga-notifications-web/issues/226#issuecomment-2695054673

The more precisely we write our timestamps, the less race conditions happen to the Go daemon's incremental config update mechanism if two users change something during the same second.

php > var_dump((int) (new DateTime())->format("Uv"));
int(1741607923054)
php > var_dump(time() * 1000);
int(1741607928000)
php >

### DIFF
--- a/application/forms/ChannelForm.php
+++ b/application/forms/ChannelForm.php
@@ -4,6 +4,7 @@
 
 namespace Icinga\Module\Notifications\Forms;
 
+use DateTime;
 use Icinga\Exception\Http\HttpNotFoundException;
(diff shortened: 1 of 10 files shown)


oxzi (commented):
@sukhwinder33445: We talked about this on July 11 and I am generally in favor of this PR. However, I cannot estimate the impact of this change.

sukhwinder33445 (approved):
Works fine and has no negative impact.

PR #308, opened 2025-03-17
RotationConfigForm: Don't increase priority when editing a rotation

sukhwinder33445:
https://github.com/Icinga/icinga-notifications-web/pull/295 missed this change.

### DIFF
--- a/application/forms/RotationConfigForm.php
+++ b/application/forms/RotationConfigForm.php
@@ -280,16 +280,19 @@ class RotationConfigForm extends CompatForm
     /**
      * Insert a new rotation in the database
      *
-     * @param int $priority The priority of the rotation
+     * @param ?int $priority The priority of the rotation to add.
+     *
+     * If no priority is specified, the priority of all existing schedule rotations is increased by 1, and the
+     * new rotation is added with the lowest priority (zero).
      *
      * @return Generator<int, DateTime> The first handoff of the rotation, as value
      */
-    private function createRotation(int $priority): Generator
+    private function createRotation(int $priority = null): Generator
     {
         $data = $this->getValues();
         $data['options'] = Json::encode($data['options']);
         $data['schedule_id'] = $this->scheduleId;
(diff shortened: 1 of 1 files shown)


application/forms/RotationConfigForm.php
```diff
@@ -280,16 +280,16 @@ public function loadRotation(int $rotationId): self
     /**
      * Insert a new rotation in the database
      *
-     * @param int $priority The priority of the rotation
+     * @param ?int $priority The priority of the rotation when editing an existing one
```
ncosta-ic, line 283: This sounds wrong. We are documenting the `createRotation` method here. It should be phrased like the following:
```txt
The priority of the rotation to add. If no priority gets passed, all existing rotations of the schedule get their priority increased and the new rotation is getting added with the lowest priority (zero).
```

ncosta-ic (changes requested):

ncosta-ic (approved):
Looks good to me :+1:

application/forms/RotationConfigForm.php
```diff
@@ -312,17 +315,19 @@ private function createRotation(int $priority): Generator
...
-            ->filter(Filter::equal('schedule_id', $this->scheduleId))
-            ->orderBy('priority', SORT_DESC);
-
-        foreach ($rotationsToMove as $rotation) {
-            $this->db->update(
-                'rotation',
-                ['priority' => new Expression('priority + 1'), 'changed_at' => $changedAt],
-                ['id = ?' => $rotation->id]
-            );
+        if ($priority === null) {
```
nilmerg, line 318: Honestly, the need for this fix should have told you that the location of this piece of code is incorrect.

The only time the priority is null now is when a rotation is created from scratch. And the need to relocate a rotation is also not a new thing. Removing one, reordered them for example. Which is why this method is a generator: To perform preparations.

And this here is such a preparation. Just put this inside `addRotation`.

PR #309, opened 2025-03-17
Timeline: Visualize future rotations

### LINKED ISSUE #214: Prevent creation of out of scope rotations
For Partial and Multi-day rotations, if the start day of the rotation is not within next month, then it is not possible to edit the rotation until the start day appears within the scope of next month

Please see the screen recording below.

https://github.com/Icinga/icinga-notifications-web/assets/33730024/97ae60dd-6063-46fb-9a6b-72ffc4d656af

sukhwinder33445:
fixes #214

### DIFF
--- a/library/Notifications/Widget/TimeGrid/BaseGrid.php
+++ b/library/Notifications/Widget/TimeGrid/BaseGrid.php
@@ -7,7 +7,6 @@ namespace Icinga\Module\Notifications\Widget\TimeGrid;
 use DateInterval;
 use DateTime;
 use Generator;
-use Icinga\Module\Notifications\Widget\Timeline\FutureEntry;
 use ipl\Html\Attributes;
 use ipl\Html\BaseHtmlElement;
 use ipl\Html\HtmlElement;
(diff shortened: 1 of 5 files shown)


flourish86 (changes requested):
When I create future entries, the rows are somehow shifted.

It only looks right, when only the first rotation has future entries and there are additional entries, with no future entries.

## Only Rotation with future entries
Broken layout
Screenshot 2025-03-24 at 14 33 12

## Second Rotation with future entries
Shifted row
Screenshot 2025-03-24 at 14 33 22

## First Rotation with future entries 
This looks fine
Screenshot 2025-03-24 at 14 37 00

flourish86:
As discussed offline, it already looks nice.

Would be great, if you could apply the following changes:

* width of the element to 3em
* change the border-color to `--base-disabled` (`#d0d3da) and the according color for the background
* align the icon to the right, with `padding-right: .25em`

Then it should look like this.
Screen Shot 2025-04-03 at 16 44 48
Screen Shot 2025-04-03 at 16 44 41

flourish86 (approved):

nilmerg (approved):
Image

Image

Issue #224, opened 2024-07-09
Add source form sometimes says the password is required even if it was filled in


julianbrost:
When opening the source form, filling all field in order, disabling certificate validation and then submitting the form, it shows a "This field is required." validation error even though all field were filled.

https://github.com/Icinga/icinga-notifications-web/assets/18552/e90254d3-3068-445e-9792-50946d7e11e9

nilmerg:
This is actually not wrong, just not well presented. The toggle automatically submits the form, causing it to be displayed again afterwards. Due to security concerns, we don't repopulate password fields in such a case. The problem here is that the dummy password is visible, which we usually show to indicate something was previously set. You'd see this as well by editing an existing source. Though, in case of creating a new source, the dummy password shouldn't be shown.

PR #310, opened 2025-03-19
Disable placeholders when rendering `SourceForm` for additions

ncosta-ic:
This PR resolves https://github.com/Icinga/icinga-notifications-web/issues/224 by disabling SourceForm's password replacement when being called through the add action.

Requires https://github.com/Icinga/ipl-html/pull/142


ncosta-ic:
This PR is obsolete as https://github.com/Icinga/ipl-html/pull/142 fixes the problem without the need for additional data attributes.

PR #311, opened 2025-03-24
Ensure new/updated rows having a changed_at greater than all previous ones

Al2Klimov:
The Go daemon's incremental config update mechanism regularily selects only config updates newer than the last known one.

This change ensures config rows having ascending changed_at, so nothing gets lost even if two users click during the same few ms. (E.g. CockroachDB allows nodes' clocks to differ by 7ms.)

closes https://github.com/Icinga/icinga-notifications/pull/288
closes https://github.com/Icinga/icinga-notifications/pull/237

## Tests

```
mysql> -- create in web
mysql> select changed_at from source;
ERROR 2013 (HY000): Lost connection to MySQL server during query
No connection. Trying to reconnect...
Connection id:    7
Current database: notifications

+---------------+
| changed_at    |
+---------------+
| 1742577440317 |
+---------------+
1 row in set (0,07 sec)

mysql> update source set changed_at=1742999000000;
Query OK, 1 row affected (0,01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000000 |
+---------------+
1 row in set (0,00 sec)

mysql> -- update in web
mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000001 |
+---------------+
1 row in set (0,00 sec)

mysql> -- insert in web again
mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000001 |
| 1742999000002 |
+---------------+
2 rows in set (0,00 sec)

mysql> -- create in web
mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000001 |
| 1742999000002 |
| 1742999000003 |
+---------------+
3 rows in set (0,01 sec)

mysql> -- rm * in web
mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000004 |
| 1742999000005 |
| 1742999000006 |
+---------------+
3 rows in set (0,00 sec)

mysql> -- create in web
mysql> select changed_at from source;
+---------------+
| changed_at    |
+---------------+
| 1742999000004 |
| 1742999000005 |
| 1742999000006 |
| 1742999000007 |
+---------------+
4 rows in set (0,01 sec)

mysql>
```

# 👍

### DIFF
--- a/library/Notifications/Common/Database.php
+++ b/library/Notifications/Common/Database.php
@@ -4,6 +4,7 @@
 
 namespace Icinga\Module\Notifications\Common;
 
+use DateTime;
 use Icinga\Application\Config as AppConfig;
 use Icinga\Data\ResourceFactory;
 use Icinga\Exception\ConfigurationError;
@@ -12,8 +13,10 @@ use ipl\Sql\Adapter\Pgsql;
 use ipl\Sql\Config as SqlConfig;
 use ipl\Sql\Connection;
 use ipl\Sql\Expression;
+use ipl\Sql\Insert;
 use ipl\Sql\QueryBuilder;
 use ipl\Sql\Select;
+use ipl\Sql\Update;
 use PDO;
 
 final class Database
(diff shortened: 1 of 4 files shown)


sukhwinder33445 (commented):
The transaction (rollback) is missing in some methods, please add (whenever `begin/CommitTransaction()` is called).
For example, in `ChannelForm::editChannel()`.

Al2Klimov:
But I didn't change ChannelForm#editChannel().

sukhwinder33445:
> But I didn't change ChannelForm#editChannel().

I see, these changes are not part of this PR.

library/Notifications/Common/Database.php
```diff
@@ -147,12 +157,65 @@ private static function getConnection(): Connection
...
+                    }
+                }
+
                 $select->where([$baseTableAlias . '.' . $condition => 'n']);
+            })
+            ->on(QueryBuilder::ON_ASSEMBLE_INSERT, function (Insert $insert) use ($db): void {
+                $columns = $insert->getColumns();
+                $pos = array_search('changed_at', $columns);
+
+                if (! $pos) {
```
sukhwinder33445, line 173: Check for `false` please, the returned value can be `0`.

library/Notifications/Common/Database.php
```diff
@@ -147,12 +157,65 @@ private static function getConnection(): Connection
...
     }
 
+    /**
+     * Return the next changed_at value for the given database and table
+     *
+     * @param Connection $db
+     * @param string $table
+     * @param mixed $nowUnixMilli
+     *
+     * @return mixed The given timestamp or 1 + the maximum changed_at value in the table, whichever is greater
```
sukhwinder33445, line 202: it's always an int, isn't it?

sukhwinder33445 (changes requested):

sukhwinder33445 (approved):

Al2Klimov:
@nilmerg Please consider the following idea of mine:

You said that the transaction method can't **unconditionally** re-try everything on 40001. What about an additional bool param which is set for creations of new stuff? Such **can** be unconditionally re-tried, right?

Al2Klimov:
@nilmerg You complained about two users editing stuff and one having to re-submit due to the 40001 error message. I suggested the transaction method to re-try the callback on 40001. And you claimed it's not always re-tryable which is wrong. Look, normally:

1. User A GETs the form
2. User B modifies the row
3. User C modifies the row
4. ...
5. User Y modifies the row
6. User A submits the form to modify the row
7. The tx callback takes the POST data and fetches the current row data as modified by user Y
8. The tx callback attempts to do its thing based on the above

Now, with serializable tx, re-trying the tx callback is just a jump back in time, but with more recent data. I mean, let's hypothetically continue the above enumeration:

* The tx callback fails with 40001 as user Z modified the row
* The tx callback is re-tried
* The tx callback takes the POST data and fetches the current row data as modified by user Z
* ...

Now what's the difference between the tx callback operating based on Y or Z data? Not much. So we can just re-try, w/o any caching.

Al2Klimov:
@nilmerg Thanks for explaining (offline) why we shouldn't retry even though we could technically. In a nutshell, it's because we want to inform a user if stuff he wants to edit changed in the meantime.

# But!

This already doesn't work for concurrent edits already COMMITted before any BEGIN/SELECT. Like,

1. GET form
2. Someone else edits successfully
3. Submit form (wins against 2)

Luckily both

* #311 and
* #329

ensure changed_at is always incrementing. So it could be used as version! Like,

1. Fetch it in a hidden form field during initial GET
2. On submit, compare it with the current changed_at
3. Mismatch? Reject! And show diff: submitted data vs current row

PR #323, opened 2025-04-17
ObjectsRendererHook: Always use `_next` as `data-base-target`

sukhwinder33445:
Otherwise, the link is opened in the same container (_self).


raviks789 (commented):
Pleas rebase onto main branch.

raviks789 (approved):

PR #324, opened 2025-04-22
Use 'addElement()' to add CSRF tokens and UID elements

### LINKED ISSUE #313: Forms: **Always** use `addElement()` for CSRF tokens
There are several places in the codebase where the CSRF token is added via `add()` instead of `addElement()`. This does not validate the CSRF token, as `add()` only adds HTML content and does not register any form elements.

Usages of the `FormUid` trait should also be checked, as there are also several `add()` calls for this. Having only briefly looked at the `FormUid` code, this should not be a problem at the moment as the implementation is not based on form specifics, e.g. validators but this could change in the future.

For our protection and other users, we could also think about a mechanism for form elements that prohibits them from being used in a form without being registered, which is the case at least with CSRF tokens. Either we check that after `Form::assemble()` and throw an exception if there is a CSRF token that is not registered, or we implement that in the CSRF token itself by throwing an exception in `assemble()` if a certain property is false, which would be set to true in `onRegistered()`. Or whatever else we can think of.

jrauh01:
Closes #313

### DIFF
--- a/application/forms/AddEscalationForm.php
+++ b/application/forms/AddEscalationForm.php
@@ -26,7 +26,7 @@ class AddEscalationForm extends Form
     protected function assemble()
     {
         $this->addElement($this->createCsrfCounterMeasure(Session::getSession()->getId()));
-        $this->add($this->createUidElement());
+        $this->addElement($this->createUidElement());
 
 
         $this->addElement(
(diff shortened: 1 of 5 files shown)


nilmerg (changes requested):
You've got a small typo there, 3 times:

counter**M**easure

nilmerg (approved):

PR #328, opened 2025-04-23
Force line-break for days-header

### LINKED ISSUE #314: days-header: Force line-break
Just a minor one: Can we force a line break for the day titles in the date-header so the appearance is consistent?

Image

jrauh01:
Closes #314

### DIFF
--- a/public/css/timeline.less
+++ b/public/css/timeline.less
@@ -227,6 +227,12 @@
   font-size: .75em;
 }
 
+#layout.twocols:not(.wide-layout) .days-header .column-title {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+}
+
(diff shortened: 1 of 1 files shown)


flourish86 (changes requested):
In detail view the break now works great.

I noticed one thing, though: Now, the headers in the schedule list break as well (which is good), but as a result the "today" element is (mis-)aligned to the top (see day header "today" in left column in the screenshot).

This would look more structured and be better scannable if it would be consistent with the other day names and be bottom aligned.

Screenshot 2025-04-23 at 11 48 36

Then while you're at it, another thing the contrast and alignment could be improved a little. Something like this should do the trick …

Screenshot 2025-04-23 at 11 50 51

… which makes it look like this

Screenshot 2025-04-23 at 11 51 00

flourish86 (approved):
Go!

PR #329, opened 2025-04-24
Revert "Ensure new/updated rows having a changed_at greater than all previous ones"

Al2Klimov:
This reverts commit 5ddc751f4e05ec9fecc851dd57bf314c983381c4.

As you requested, now completely different channels won't conflict (40001) w/ each other, as they don't SELECT a common MAX(changed_at) in a serializable tx.

But one and the same e.g channel would still conflict (40001) if updated by two users which is good.

### DIFF
--- a/library/Notifications/Common/Database.php
+++ b/library/Notifications/Common/Database.php
@@ -14,6 +14,7 @@ use ipl\Sql\Connection;
 use ipl\Sql\Expression;
 use ipl\Sql\QueryBuilder;
 use ipl\Sql\Select;
+use ipl\Sql\Update;
 use PDO;
 
 final class Database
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@julianbrost @oxzi @nilmerg But, honestly speaking, as I already said:

Web is a point&click thing, so generally the users are slow enough not to run into #311 (40001 error), ex. once or twice a year.

It is simply the right thing that a row, which is committed later, has a greater changed_at than previous ones. Especially if our incremental config update mechanism relies on it!

julianbrost:
I'm missing context to understand what you're doing here.

Al2Klimov:
The point is, our incremental config update mechanism relies on strongly sequential changed_at not to miss any updates. I ensured this consistency with this:

* https://github.com/Icinga/icinga-notifications-web/pull/311

And now @nilmerg says, like, _Oh no! Now, whenever two user edit distinct channels during the same(!) very few ms [which happens.. once a year? twice.. ? Who cares?], one of them gets an error message [saying that tx serialization failed or similar] and has to re-submit. This is bad!_

Al2Klimov:
Won't happen.

We'll keep the current behavior, but re-try based on per-row changed_at.

PR #331, opened 2025-04-24
Fix button margins

### LINKED ISSUE #330: Button margin missing
The left margins on the `New Schedule` and `New Event Rule` buttons are missing.

Image

Image

jrauh01:
Fix the missing button margins by adding the full-width class to the contents and set the button's classes to `.add-new-component`.

Fixes #330


jrauh01:
Will already be fixed in #315

PR #333, opened 2025-04-30
Handle config decoding errors from DB charset issues

### LINKED ISSUE #271: Missing information in docs: Database ressource Character Set needs to be utf8
This has cost me quite some time to figure out. When the `Character Set` of the database ressource is not set to `utf8` you only get the following error when creating a `Channel`: 
image

So please add this to the documentation.

jrauh01:
If the database's character set is not set correctly, the decoding of the config attributes may fail. For that case an error message will be rendered in the form. There will also be a docs entry to prevent that.

Closes #271

### DIFF
--- a/application/forms/ChannelForm.php
+++ b/application/forms/ChannelForm.php
@@ -6,14 +6,16 @@ namespace Icinga\Module\Notifications\Forms;
 
 use DateTime;
 use Icinga\Exception\Http\HttpNotFoundException;
-use Icinga\Module\Notifications\Model\Channel;
 use Icinga\Module\Notifications\Model\AvailableChannelType;
+use Icinga\Module\Notifications\Model\Channel;
 use Icinga\Module\Notifications\Model\Contact;
 use Icinga\Module\Notifications\Model\RuleEscalationRecipient;
 use Icinga\Web\Session;
+use ipl\Html\Attributes;
 use ipl\Html\Contract\FormSubmitElement;
 use ipl\Html\FormElement\BaseFormElement;
 use ipl\Html\FormElement\FieldsetElement;
+use ipl\Html\HtmlElement;
 use ipl\I18n\GettextTranslator;
 use ipl\I18n\StaticTranslator;
 use ipl\Sql\Connection;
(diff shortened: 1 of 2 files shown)


application/forms/ChannelForm.php
```diff
@@ -267,6 +269,22 @@ protected function createConfigElements(string $type, string $config): void
         $elementsConfig = json_decode($config, true);
 
         if (empty($elementsConfig)) {
+            $this->prependHtml(
+                HtmlElement::create(
+                    'ul',
+                    Attributes::create(['class' => 'errors']),
+                    HtmlElement::create(
+                        'li',
+                        Attributes::create(),
```
nilmerg, line 278: null, please

application/forms/ChannelForm.php
```diff
@@ -267,6 +269,22 @@ protected function createConfigElements(string $type, string $config): void
...
+                        sprintf(
+                            'Could not decode options for type \'%s\'.'
+                            . ' Check if your database\'s character set is set correctly.',
```
nilmerg, line 281: > …set is set…

Such a repetition hurts readability. Please avoid it.

nilmerg (changes requested):
Instead of documenting this, Alex's suggestion makes more sense. (To use the correct encoding automatically) `Icinga\Module\Notifications\Common\Database` can do that.

application/forms/ChannelForm.php
```diff
@@ -267,6 +269,22 @@ protected function createConfigElements(string $type, string $config): void
...
+            $this->prependHtml(
+                HtmlElement::create(
+                    'ul',
+                    Attributes::create(['class' => 'errors']),
+                    HtmlElement::create(
+                        'li',
+                        null,
+                        sprintf(
+                            'Could not decode options for type \'%s\'.'
+                            . ' Check if your database\'s character set is correctly configured.',
```
nilmerg, line 281: Mark this as translatable with `$this->translate()`

nilmerg (changes requested):
Please rebase

nilmerg (approved):

PR #340, opened 2025-06-23
ContactGroupsController: Show `add contact` link if no contact exists

### LINKED ISSUE #338: Contact Groups: Disable
In #276 we address the issue, when no Channels are available to select.

Then there's a similar situation for contact groups. When no contacts are created, it will be impossible to create a new Contact Group in the first place.

We should adapt the solution for this scenario:

* Add a hint within the empty state
* link to the contact creation tab accordingly
* Disable the "Add contact group" button

Users hit a dead end, when they want to create a a contact group, but no contacts are available to select

Image

sukhwinder33445:
resolves #338


sukhwinder33445:
Case : There is only one contact available, and there is also a contact group containing this contact.

When the user deletes the contact, should the contactgroup also be deleted? Because the contact group now has no members and the modal `Edit contact group` again shows no suggestion (in contactgroup detail view).

sukhwinder33445:
We can display a kind of wizard when the user activates the module for the first time to show these steps (channel, contact). Same as the setup wizard in Icingaweb.

flourish86:
I thought of something like this for the case when there's neither channels nor contacts configured.

And another thing: The empty state statement still reads "No contacts found.", even for contact groups, it should be "No contact groups found."

Frame 70

flourish86 (approved):

nilmerg (approved):

PR #359, opened 2025-10-06
Help users getting started with schedules

### LINKED ISSUE #280: Help users getting started with schedules

nilmerg:
resolves #280


flourish86 (approved):
lgtm!

PR #365, opened 2025-10-09
Final rotation config improvements

### LINKED ISSUE #274: Improve rotation configuration
Users seem to struggle to…

* …know what the title is used for
* …navigate the available options
* …understand the mode pictograms
* …understand what the different modes are used for
* …understand the meaning of *handoffs*
  * The fact that they configure a rotation doesn't seem to help to know what a *handoff* might be
  * The need to configure a handoff interval, even if only one member is chosen, is irritating
  * The effect of the *first* handoff is not clear
* …configure intervals of full days starting at 00:00 in the multi day mode
  * The end date's last element is always 11:30. To configure full 24 hours of a day, the next day's 00:00 must be chosen

## Goal

* Change *Title* to *Rotation Name*
* @flourish86 will provide suggestions/mockups for the remaining ones

nilmerg:
Changes according to: https://github.com/Icinga/icinga-notifications-web/issues/274#issuecomment-2620971839
Disabling the time period inputs is not possible, though.

resolves #274

### DIFF
--- a/application/forms/RotationConfigForm.php
+++ b/application/forms/RotationConfigForm.php
@@ -18,6 +18,7 @@ use Icinga\Util\Json;
 use Icinga\Web\Session;
 use ipl\Html\Attributes;
 use ipl\Html\DeferredText;
+use ipl\Html\FormDecoration\DescriptionDecorator;
 use ipl\Html\FormElement\FieldsetElement;
 use ipl\Html\HtmlDocument;
 use ipl\Html\HtmlElement;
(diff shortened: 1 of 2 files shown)


application/forms/RotationConfigForm.php
```diff
@@ -909,7 +922,8 @@ protected function assembleMultiDayOptions(FieldsetElement $options): DateTime
             'step' => 1,
             'min' => 1,
             'value' => 1,
-            'label' => $this->translate('Handoff every')
+            'label' => $this->translate('Handoff every'),
+            'description' => $this->translate('Have multiple rotation members take turns after this interval.')
```
sukhwinder33445: Please use `ipl-html` `DescriptionDecorator` instead.

sukhwinder33445 (approved):

PR #369, opened 2025-10-17
Proper timezone handling for schedules

### LINKED ISSUE #273: Improper time zone handling in schedules
## Problem

The re-designed schedules introduced with #177 don't provide proper handling of different time zones.

As long as rotation entries created in the same time zone are visible and the viewer has said time zone configured, all is fine.

Problems occur once entries created in different time zones are mixed or if the viewer has a different one configured.

## How to solve this

These data types are affected:

* Every `MillisecondTimestamp`: Those are already normalized to UTC upon persistence and back **to the viewer's time zone** upon retrieval
* `rotation.first_handoff`: This is only a date. A time zone can only be applied by combining it with the mode specific start time
* `timeperiod_entry.timezone` and `timeperiod_entry.rrule`

### Detail

- schedule creation MUST demand a timezone so we need `schedule.timezone`, though, it is just a setting for the UI.
  - Once set, this setting MUST be immutable.
  - `timeperiod_entry.timezone` is kept for now, but the UI MUST ensure to use the schedule's timezone upon creation.
    - Loading an entry SHOULD normalize it to the schedule's timezone, if it differs.
- Upon opening (viewing or editing) a schedule, its timezone is the base for all time based calculations and comparisons in the backend.
  - summer/winter/leap time support is out of scope right now, so don't consider this at all
  - The timezone used to display entries MUST be variable, but right now is always set to the schedule's timezone by default.
  - The frontend must clearly indicate the used timezone. The title seems to be good fit for now. (<title> (\<tz>))
  - Editing a rotation uses the same timezone. The form must clearly indicate the timezone as well.

```mermaid
---
config:
  layout: dagre
---
flowchart TB
    n8["Backend"] <--> n2["UI"]
    n1["Database"] <--> n8
    n5["schedule.timezone (variable)"] -.-> n2
    n6["timeperiod_entry.timezone"] -.-> n1
    n7["schedule.timezone"] -.-> n8
    n8@{ shape: procs}
    n2@{ shape: display}
    n1@{ shape: db}
    n5@{ shape: text}
    n6@{ shape: text}
    n7@{ shape: text}
```

### List

The list shows currently a preview of each schedule's result. With mixed timezones this will be a problem, since there's only a single header and this won't necessarily align with each schedule's timezone. We may introduce a separate header for each list item, but this introduces additional problems. The view becomes hard to read, as if separate headers differ, dates do not align and TODAY won't be the same column everywhere. Another option is to simply assume the user's timezone for the list, but then clicking on an item leads to a different result: The timelines don't match. So, let me propose to use the third option: Removing the preview. Sounds boring, but is maybe the least controversial solution…

jrauh01:
This PR introduces timezone support for schedules. It relies on the `schedule.timezone` column in the database added in https://github.com/Icinga/icinga-notifications/pull/344. 

For the following explanation, it is important to understand the difference between schedule and display timezone.
- **schedule timezone**: the timezone the schedule is created for and stored in the database.
- **display timezone**: the timezone the user want to display the schedule in.

### Schedule List

The timeline part has been removed completely, so that we don't have to create a header for each schedule which wouldn't be very well readable.

### Schedule Creation

The form to create a new schedule now has a suggestion input to choose the timezone the schedule will be created for. Once created, the timezone cannot be changed via the web UI, because the suggestion input won't be displayed for the `editAction()`. This is ensured with a bool `$showTimezoneSuggestionInput` in the `ScheduleForm` which defaults to `false`. The default of the suggestion input is the timezone configured in the the user preferences. The value of the suggestion input will be stored in the `schedule.timezone` column.

### ScheduleTimezoneStorage

The `ScheduleTimezoneStorage` is a util storage class which stores the schedule timezone globally. The schedule timezone is set in the `ScheduleController::indexAction()`.

### Schedule Detail View

The timeline start is in the display timezone and is passed to the `ScheduleDatail` in the `ScheduleController::indexAction()`. The `rotation.actual_handoff` and `timeperiod_entry.start_time` are stored as unix timestamps, so the entries are displayed correctly in the timeline grid. Additionally all date formatters are constructed with the display timezone, so all titles are displayed correctly for the user's choice.

The user changes the display timezone via a dropdown control called `TimezonePicker`. **This may be change in the future.**

### Rotation Config Form

All configurations in the form are done in the **schedule timezone**, no matter what display timezone is chosen. The `RotationConfigForm` is constructed with both the **display** and **schedule timezone**. If the display timezone differs following will happen:
- A warning is displayed to make the user aware that they configure the rotation in the **schedule timezone**.
- The first handoff hint is complemented by an additional hint that shows the first handoff in the **display timezone**

The dropdowns to select times did show 'Next Day' hints earlier. This logic has been replaced by an own option group for next day elements.

Requires https://github.com/Icinga/icinga-notifications/pull/344
Requires https://github.com/Icinga/ipl-scheduler/pull/55
Fixes #273

### DIFF
--- a/application/controllers/ScheduleController.php
+++ b/application/controllers/ScheduleController.php
@@ -16,6 +16,7 @@ use Icinga\Module\Notifications\Web\Control\TimezonePicker;
 use Icinga\Module\Notifications\Widget\Detail\ScheduleDetail;
 use Icinga\Module\Notifications\Widget\RecipientSuggestions;
 use Icinga\Module\Notifications\Widget\TimezoneWarning;
+use Icinga\Web\Session;
 use ipl\Html\Form;
 use ipl\Html\Html;
 use ipl\Stdlib\Filter;
@@ -66,7 +67,10 @@ class ScheduleController extends CompatController
         $this->addContent(new ScheduleDetail(
             $schedule,
             $scheduleControls,
-            new DateTime('today', new DateTimeZone($timezonePicker->getDisplayTimezone()))
+            new DateTime('today', new DateTimeZone(
+                $timezonePicker->getValue(TimezonePicker::DEFAULT_TIMEZONE_PARAM)
+                    ?? $this->getDisplayTimezoneFromSession($schedule->timezone)
+            )),
         ));
     }
 
@@ -121,7 +125,7 @@ class ScheduleController extends CompatController
     {
         $scheduleId = (int) $this->params->getRequired('schedule');
         $scheduleTimezone = $this->getScheduleTimezone($scheduleId);
-        $displayTimezone = (new TimezonePicker($scheduleTimezone))->getDisplayTimezone();
+        $displayTimezone = $this->getDisplayTimezoneFromSession($scheduleTimezone);
         $this->setTitle($this->translate('Add Rotation'));
 
         if ($displayTimezone !== $scheduleTimezone) {
@@ -160,7 +164,7 @@ class ScheduleController extends CompatController
         $id = (int) $this->params->getRequired('id');
         $scheduleId = (int) $this->params->getRequired('schedule');
         $scheduleTimezone = $this->getScheduleTimezone($scheduleId);
-        $displayTimezone = (new TimezonePicker($scheduleTimezone))->getDisplayTimezone();
+        $displayTimezone = $this->getDisplayTimezoneFromSession($scheduleTimezone);
         $this->setTitle($this->translate('Edit Rotation'));
 
         if ($displayTimezone !== $scheduleTimezone) {
@@ -254,17 +258,45 @@ class ScheduleController extends CompatController
      */
     protected function createTimezonePicker(string $scheduleTimezone, int $scheduleId): TimezonePicker
     {
-        return (new TimezonePicker($scheduleTimezone))
+        $defaultTimezoneParam = TimezonePicker::DEFAULT_TIMEZONE_PARAM;
+        return (new TimezonePicker())
             ->populate([
-                TimezonePicker::DEFAULT_TIMEZONE_PARAM => $this->params->get(TimezonePicker::DEFAULT_TIMEZONE_PARAM)
+                $defaultTimezoneParam => $this->params->get($defaultTimezoneParam)
+                    ?? $this->getDisplayTimezoneFromSession($scheduleTimezone)
             ])
-            ->on(TimezonePicker::ON_SUBMIT, function (TimezonePicker $timezonePicker) use ($scheduleId) {
-                $this->redirectNow(
-                    Links::schedule($scheduleId)->with([
-                        TimezonePicker::DEFAULT_TIMEZONE_PARAM => $timezonePicker->getDisplayTimezone()
-                    ])
-                );
-            })
+            ->on(
+                TimezonePicker::ON_SUBMIT,
+                function (TimezonePicker $timezonePicker) use ($defaultTimezoneParam, $scheduleId, $scheduleTimezone) {
+                    $this->writeDisplayTimezoneToSession($timezonePicker->getValue($defaultTimezoneParam));
+                    $this->redirectNow(Links::schedule($scheduleId)->with([
+                        $defaultTimezoneParam => $timezonePicker->getValue($defaultTimezoneParam)
+                    ]));
+                }
+            )
             ->handleRequest($this->getServerRequest());
+    }
+
+    /**
+     * Get the display timezone from the session
+     *
+     * @param string|null $defaultTimezone
+     *
+     * @return string
+     */
+    protected function getDisplayTimezoneFromSession(?string $defaultTimezone = null): string
+    {
+        return Session::getSession()->getNamespace('notifications')->get('schedule.display_timezone', $defaultTimezone);
+    }
+
+    /**
+     * Write the display timezone to the session
+     *
+     * @param string $displayTimezone
+     *
+     * @return void
+     */
+    protected function writeDisplayTimezoneToSession(string $displayTimezone): void
+    {
+        Session::getSession()->getNamespace('notifications')->set('schedule.display_timezone', $displayTimezone);
     }
 }
--- a/library/Notifications/Web/Control/TimezonePicker.php
+++ b/library/Notifications/Web/Control/TimezonePicker.php
@@ -6,7 +6,6 @@ namespace Icinga\Module\Notifications\Web\Control;
 
 use DateTime;
 use DateTimeZone;
-use Icinga\Web\Session;
 use IntlTimeZone;
 use ipl\Html\Form;
 use ipl\I18n\Translation;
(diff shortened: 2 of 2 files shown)


public/css/schedule.less
```diff
@@ -67,6 +67,25 @@
...
 
+.timezone-warning {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  column-gap: 1em;
+
+  width: fit-content;
+  margin: 0 auto 1em auto;
+  padding: .5em 1em;
```
nilmerg: you have a design rule, so please move the padding there, as padding generally is design related, without it, the box would look differently after all

public/css/schedule.less
```diff
@@ -67,6 +67,25 @@
...
+
+  i.icon:before {
```
nilmerg: please use double colon syntax for `::before` and `::after`

library/Notifications/Widget/Timeline.php
```diff
@@ -306,7 +307,11 @@ protected function assembleSidebarEntry(Rotation $rotation): BaseHtmlElement
         $entry = new HtmlElement('div', Attributes::create(['class' => 'rotation-name']));
 
         $form = new MoveRotationForm();
-        $form->setAction(Links::moveRotation()->getAbsoluteUrl());
+        $form->setAction(
+            Links::moveRotation()
+                ->with(['display_timezone' => ScheduleDateTimeFactory::getDisplayTimezone()->getName()])
+                ->getAbsoluteUrl()
```
nilmerg: The `rotationAdd` and `rotationSettings` handle this themselves, so I think `moveRotation` should do the same.

library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -57,7 +58,12 @@ protected function assembleContainer(BaseHtmlElement $container): void
             $dateType = \IntlDateFormatter::SHORT;
         }
 
-        $formatter = new \IntlDateFormatter(\Locale::getDefault(), $dateType, $timeType);
+        $formatter = new \IntlDateFormatter(
+            \Locale::getDefault(),
+            $dateType,
+            $timeType,
+            ScheduleDateTimeFactory::getDisplayTimezone()
```
nilmerg: Shouldn't it suffice to use the timezone of either the entry's start or end?

application/controllers/ScheduleController.php
```diff
@@ -51,10 +54,23 @@ public function indexAction(): void
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
             ->populate(['mode' => $this->params->get('mode')])
             ->on(Form::ON_SUCCESS, function (ScheduleDetail\Controls $controls) use ($id) {
-                $this->redirectNow(Links::schedule($id)->with(['mode' => $controls->getMode()]));
+                $redirectUrl = Links::schedule($id)->with(['mode' => $controls->getMode()]);
+                $requestUrl = Url::fromRequest();
+                if ($requestUrl->getParam('mode') !== $controls->getValue('mode')) {
```
nilmerg: Why…? If you wanted to avoid the redirect although nothing changed, then this is the wrong place, and PR.

sukhwinder33445:
Some timezones from `DateTimeZone::listIdentifiers()` are not supported by `IntlDateFormatter`.
The method `IntlTimeZone::createEnumeration()` provides valid, supported time zones. Please use this instead. 

Please make sure the daemon also supports these time zones, or filter out unsupported ones to avoid runtime errors.

application/controllers/ScheduleController.php
```diff
@@ -92,6 +106,7 @@ public function addAction(): void
     {
         $this->setTitle($this->translate('New Schedule'));
         $form = (new ScheduleForm(Database::get()))
+            ->setShowTimezoneSuggestionInput()
             ->setAction($this->getRequest()->getUrl()->getAbsoluteUrl())
```
nilmerg: It's now required to add `->setParam('showCompact')` here, otherwise tabs are visible inside the modal in case of an invalid tz.

library/Notifications/Widget/Timeline/EntryFlyout.php
```diff
@@ -251,9 +291,24 @@ protected function generateAndSetRotationInfo(): static
 
         $noneType = \IntlDateFormatter::NONE;
         $shortType = \IntlDateFormatter::SHORT;
-        $timeFormatter = new \IntlDateFormatter(\Locale::getDefault(), $noneType, $shortType);
-        $dateFormatter = new \IntlDateFormatter(\Locale::getDefault(), $shortType, $noneType);
-        $firstHandoff = $dateFormatter->format(DateTime::createFromFormat('Y-m-d', $this->firstHandoff));
+        $scheduleTimezone = new DateTimeZone(ScheduleTimezoneStorage::getScheduleTimezone());
+        $displayTimezone = new DateTimeZone((new TimezonePicker($scheduleTimezone->getName()))->getDisplayTimezone());
```
nilmerg: You don't need `ScheduleTimezoneStorage::getScheduleTimezone()`, when an `Timeline\Entry` is created, it's done by the rotation which knows its own tz very well, so just teach such entries about their own tz. It might be a good idea to change `EntryFlyout::withActiveMember()` to `EntryFlyout::forEntry()`.

Similarly, the display tz should be a mandatory requirement for a flyout. Make it a constructor argument. The timeline, which is responsible to prepare and construct flyouts, knows the display tz very well and can pass it on.

public/css/form.less
```diff
@@ -84,6 +85,15 @@
   }
 }
 
+.timezone-picker {
+  display: inline;
```
nilmerg: Apply CSS classes `icinga-form, inline, icinga-controls` to the form and this is not required. The extra div around the select is also not required this way.

public/css/form.less
```diff
@@ -84,6 +85,15 @@
...
+  margin-left: 1em;
```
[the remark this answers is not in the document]
jrauh01: You mean the `margin-left` should be moved there?

application/forms/ScheduleForm.php
```diff
@@ -176,6 +199,40 @@ protected function assemble(): void
             'placeholder'   => $this->translate('e.g. working hours, on call, etc ...')
         ]);
 
+        if ($this->showTimezoneSuggestionInput) {
+            $this->addElement(
+                'suggestion',
+                'timezone',
+                [
+                    'suggestionsUrl' => Url::fromPath('notifications/suggest/timezone', [
+                        'showCompact' => true, '_disableLayout' => 1
```
nilmerg: Break once, break always.

library/Notifications/Widget/Timeline/EntryFlyout.php
nilmerg: Pattern: Fr-Mo, 17:00-09:00
Schedule tz: Europe/Berlin
Display tz: Australia/Melbourne

image

nilmerg (changes requested):

library/Notifications/Widget/Timeline/EntryFlyout.php
```diff
@@ -43,21 +44,29 @@ class EntryFlyout extends BaseHtmlElement
     /** @var ?ValidHtml Information about name and mode of the rotation */
     protected ?ValidHtml $nameInfo = null;
 
+    /** @var DateTimeZone The display timezone */
+    protected DateTimeZone $displayTimezone;
+
+    public function __construct(DateTimeZone $displayTimezone)
```
nilmerg: missing doc

public/css/form.less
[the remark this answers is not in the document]
nilmerg: ping

PR #370, opened 2025-10-20
Fix/event rule without escalations

### LINKED ISSUE #251: Event Rule breaks if I delete a Contact used in an escalation
Hoi,

while testing some stuff I noticed that, when I create an Event Rule with two escalations, first a Contact, then a Schedule - save it - and afterwards delete the Contact in the Contacts overview, the Event Rule cant be edited anymore cause of: `Undefined array key 1`

screenshot_2024-07-24-155533

Regards,
Markus

BastianLedererIcinga:
Expand load function of EventRuleConfigForm to handle the case of a rule with no escalations, so the UI will look as when creating a new rule.
This was the before the change:
Screenshot from 2025-10-20 13-14-19

This is how it looks now:
Screenshot from 2025-10-20 13-14-47

fix #251

### DIFF
--- a/test/php/application/forms/EventRuleConfigFormTest.php
+++ b/test/php/application/forms/EventRuleConfigFormTest.php
@@ -184,16 +184,10 @@ class EventRuleConfigFormTest extends TestCase
         $ruleEscalationMock->expects($this->once())
             ->method('orderBy')
             ->with('position', 'asc')
-            ->willReturn($ruleEscalationMock);
+            ->willReturnSelf();
 
-        $queryResultMock = $this->createMock(ResultSet::class);
-        $queryResultMock->expects($this->once())
-            ->method('hasResult')
-            ->willReturn(true);
-
-        $queryResultMock->expects($this->exactly(2))
-            ->method('current')
-            ->willReturnOnConsecutiveCalls(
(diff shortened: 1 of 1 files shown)


application/forms/EventRuleConfigForm.php
```diff
@@ -244,14 +244,19 @@ public function createExternalSubmitButtons(): array
...
-        $this->populate([
+        $fields = [
             'id'                => $rule->id,
             'name'              => $rule->name,
-            'object_filter'     => $rule->object_filter,
-            'escalations'       => EventRuleConfigElements\Escalations::prepare(
+            'object_filter'     => $rule->object_filter
+        ];
+
+        if ($rule->rule_escalation->count() > 0) {
```
nilmerg: You can note this down for the future, as it applies generally to all languages and frameworks:

> Don't count if you're not interested in the actual number of items.

In this case, you want to know whether there are any results or none. Our framework allows this by using this:

```php
$escalations = $query->execute();
if (! $escalations->hasResult()) {
```

Though, this will run the query explicitly. Don't forget to use the result in subsequent expressions, as otherwise the query is ran again and the advantage of using `hasResult` is gone.

nilmerg (approved):

PR #373, opened 2025-10-23
Source bound event rules

### LINKED ISSUE #354: Source bound event rules
We plan to establish a mandatory and strong connection between sources and event rules. This is a breaking change.

When creating a new event rule, the user must not only choose a name, but also to which source the rule belongs to. If there are no sources configured yet, the button must be disabled and the user pointed to the source configuration, just like it's been done as part of other recent changes.

Regarding compatibility with existing rules configured before this change, we'll have to think about this eventually, but there's no need for that right now.

The filter capabilities of event rules are severely limited at the moment. To enhance this and to provide enough flexibility for source integrations, the latter are now required. In order to configure sources or filters in an event rule, a source integration is mandatory. The integration is now expected to provide the necessary controls to configure the filter. If no integration is available, neither a source nor event rules can be configured.

A source integration (e.g. Icinga DB Web) must provide an implementation for the following hook:

```php
<?php

/* Icinga Notifications Web | (c) 2025 Icinga GmbH | GPLv2 */

namespace Icinga\Module\Notifications\Hook\V1;

use ipl\Html\Contract\Form;
use ipl\Web\Widget\Icon;

interface SourceHook
{
    /**
     * Get the type of source this integration is responsible for
     *
     * @return string
     */
    public function getSourceType(): string;

    /**
     * Get the label of the source this integration is responsible for
     *
     * @return string
     */
    public function getSourceLabel(): string;

    /**
     * Get the icon of the source this integration is responsible for
     *
     * @return Icon
     */
    public function getSourceIcon(): Icon;

    /**
     * Get available filter targets for event rules
     *
     * If this returns an array of length 1, {@see getRuleFilterEditor()} will immediately be called with the item's
     * key. If the returned array contains multiple values, a simple select box will be rendered first, and the user
     * has to choose an item from the list.
     *
     * @param int $sourceId The ID of the source the rule belongs to
     *
     * @return array<string, string|array<string, string>> target => label | optgroup => (target => label)
     */
    public function getRuleFilterTargets(int $sourceId): array;

    /**
     * Get an editor for the given filter
     *
     * The returned form MUST NOT have a request associated with it and MUST NOT navigate away upon submission.
     * The action of the form is overridden in any case.
     *
     * @param string $filter A filter template or a filter previously serialized by {@see serializeRuleFilter()}
     *
     * @return Form
     */
    public function getRuleFilterEditor(string $filter): Form;

    /**
     * Serialize the filter of the given editor
     *
     * The returned string is stored as-is in the database. The source MUST be able to deserialize it.
     * Upon editing by a user, {@see getRuleFilterEditor()} will be called with the serialized filter.
     *
     * @param Form $editor
     *
     * @return string
     */
    public function serializeRuleFilter(Form $editor): string;
}

```

This is done as usual in `run.php` as follows:

```php
$this->provideHook('Notifications/v1/Source');
```

There can only be a single integration be responsible for a given source. When loading hooks of the aforementioned type, we have to make sure `SourceHook::getSourceType()` matches with the event rule's associated source.

Since #159, the `EventRuleController` and `EventRuleConfigForm` are already somewhat prepared for this.

`EventRuleConfigForm::createObjectFilterControls()` needs to be adjusted so that no text element is shown if there is a filter defined. Instead, only the button with the ⚙️ icon is shown, similar to the button if no filter is defined.

`SourceHook::getRuleFilterTargets()` and `SourceHook::getRuleFilterEditor()` are supposed to be used by `EventRuleController::searchEditorAction()` as part of a two step mini wizard: The user has to choose a target first, if there are multiple, then the editor appears instead. If a filter is being edited, the first step is of course not required. To change the target selection, it is required to clear the filter and re-open the editor.

Upon submission of the editor, `SourceHook::serializeRuleFilter()` is used to write the new changes to the session.

nilmerg:
resolves #354

### DIFF
--- a/doc/01-About.md
+++ b/doc/01-About.md
@@ -6,11 +6,11 @@
 
     At the moment, we don't provide any support for this module.
 
-Icinga Notifications is a set of components that processes received events from miscellaneous sources, manages
-incidents and forwards notifications to predefined contacts, consisting of:
+Icinga Notifications is a set of components that processes received events from various sources, manages
+incidents and forwards notifications to predefined contacts. The components are:
 
 * [Icinga Notifications](https://github.com/Icinga/icinga-notifications), which receives events and sends notifications.
-* Icinga Notifications Web, which provides graphical configuration.
+* Icinga Notifications Web, which lets users configure Icinga Notifications and manage incidents.
 
 Icinga 2 itself and other sources propagate state updates and other events to [Icinga Notifications](https://github.com/Icinga/icinga-notifications).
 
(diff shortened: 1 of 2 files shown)


application/forms/DeleteSourceForm.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public function removeSource(Connection $db): void
+    {
+        $evtRuleConfigForm = (new ReflectionClass(EventRuleConfigForm::class))
+            ->newInstanceWithoutConstructor();
+        $rules = Rule::on($db)
+            ->columns('id')
+            ->filter(Filter::equal('source_id', $this->source->id));
+        foreach ($rules as $rule) {
+            $evtRuleConfigForm->removeRule($db, $rule);
```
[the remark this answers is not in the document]
nilmerg: True. Granted.

sukhwinder33445 (commented):
There are some minor ui issues In smaller browser window with two col layout:
Screenshot 2025-10-24 at 16 28 09

The model/form does not apply any padding:
Screenshot 2025-10-24 at 16 28 37

I think we should use the filter icon instead of the cog, it makes the purpose clearer.
Screenshot 2025-10-24 at 16 29 51

application/forms/DeleteSourceForm.php
```diff
@@ -0,0 +1,119 @@
...
+     * Get this source's name
+     *
+     * @return string
+     */
+    public function getSourceName(): string
+    {
+        return $this->source->name;
+    }
+
+    protected function assemble()
```
sukhwinder33445: Please add return type.

public/css/form.less
```diff
@@ -23,6 +23,12 @@
   fieldset p.description:last-child { // fieldset > .control-group:last-of-type unset the last element's bottom margin
     margin-top: 1em;
   }
```
sukhwinder33445: This can be removed now, as the style below covers this case as well

PR #374, opened 2025-10-24
Feature/improved rotation tooltips

### LINKED ISSUE #364: Users seem to struggle with how to see details about rotation entries
The current tooltips on hover are not final. Rich popups that appear earlier/faster should improve this.

@flourish86 has provided suggestions/mockups:

Image

Image

BastianLedererIcinga:
Adds tooltips that are shown when hovering an entry. resolves #364 
Screenshot from 2025-10-24 13-10-40

### DIFF
--- a/library/Notifications/Widget/Timeline/Rotation.php
+++ b/library/Notifications/Widget/Timeline/Rotation.php
@@ -120,7 +120,6 @@ class Rotation
                     Filter::greaterThanOrEqual('until_time', $after) // Or one which isn't over yet
                 )
             ));
-
         foreach ($entries as $timeperiodEntry) {
             if ($timeperiodEntry->member->contact->id !== null) {
                 $member = new Member($timeperiodEntry->member->contact->full_name);
--- a/library/Notifications/Widget/Timeline/EntryFlyout.php
+++ b/library/Notifications/Widget/Timeline/EntryFlyout.php
@@ -48,9 +48,9 @@ class EntryFlyout extends BaseHtmlElement
      *
      * @param Member $member
      *
-     * @return self
+     * @return static
      */
-    public function withActiveMember(Member $member): self
+    public function withActiveMember(Member $member): static
     {
         if (! isset($this->timeInfo)) {
             $this->generateAndSetRotationInfo();
@@ -234,7 +234,7 @@ class EntryFlyout extends BaseHtmlElement
      *
      * @return $this
      */
-    public function generateAndSetRotationInfo(): static
+    protected function generateAndSetRotationInfo(): static
     {
         $this->setTag('div');
         $this->setAttribute('class', 'rotation-info');


library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -9,12 +9,14 @@
...
 use ipl\Html\HtmlElement;
 use ipl\Html\Text;
+use ipl\Html\ValidHtml;
 use ipl\Web\Widget\Icon;
 
 class Entry extends TimeGrid\Entry
 {
     /** @var Member */
     protected $member;
+    protected $flyoutContent;
```
nilmerg: Please add:

* a type hint (`?ValidHtml`)
* a docstring
* an empty line before it

library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -33,6 +35,21 @@ public function getColor(int $transparency): string
         return TimeGrid\Util::calculateEntryColor($this->getMember()->getName(), $transparency);
     }
 
+    /** Sets Content of a Flyout that is shown when the Entry is hovered
+     * @param ValidHtml $content
+     * @return $this
+     */
```
nilmerg: * Make always a line break after `/**`
* Please insert empty lines after the subject…
* …and after the parameter list

e.g.

```
/**
 * subject
 *
 * @param …
 *
 * @return …
 */
```

library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -16,7 +16,9 @@ class Entry extends TimeGrid\Entry
 {
     /** @var Member */
     protected $member;
-    protected $flyoutContent;
+
+    /**@var ?ValidHtml Content of the flyoutmenu that is shown when the entry is hovered*/
```
nilmerg: A space after `/**` and before `*/` please

public/css/timeline.less
```diff
@@ -60,6 +64,72 @@
...
+            display: flex;
+          }
+
+          &:before {
+            content: '';
+            position: absolute;
+            display: inline-block;
+            width: 1em;
+            height: 1em;
+            left: 0.5em;
```
nilmerg: I know I've wrote that the placement should not depend on the entry's width, but of course, except when the entry is too narrow. Now the flyout appears so far to the left or right that it looks weird, at least to me. Though, I just guess you used .5em as the offset due to the narrow entries. What about flagging entries that fall below a given threshold and only use a dynamic offset (e.g. 50%) for them? For all others a fixed offset, say 1.5-2em.

The backend knows how large the grid is, so it can detect narrow entries just fine I guess and apply an additional CSS class.

library/Notifications/Widget/Timeline.php
```diff
@@ -249,6 +250,8 @@ public function getEntries(): Traversable
                     $resultEntry = (new Entry($entry->getId()))
                         ->setStart($start)
                         ->setEnd($end)
+                        ->setFlyoutContent($entry->getFlyoutContent())
+                        ->calculateAndSetWidthClass($this->getGrid())
```
nilmerg: What I noticed, is that on narrow entries the flyout does not highlight who is on duty in the current rotation. That's a lost opportunity as the flyout got enough space to even show long names which get ellipsized in the UI.

Also, the flyout is still rendered in the overview of schedules. The timeline previews there will be gone with another change as previously pointed out, but not the minimal mode support of the timeline. As such, the flyout should not be rendered at all in the list right now, i.e. also not prepared by the backend.

In the end, the detection how wide or narrow an entry is, is also only required if the flyout is visible. So please make this dependent on the view mode as well, or preferably, make the entire process a single step. So once the timeline asks for a flyout, or to apply one to an entry, it asks the rotation to do so.

BastianLedererIcinga: I added an additional line for the member on duty, and therefore decided to skip generating the list of members in cases where only one exists. The result looks like this:
Screenshot from 2025-10-31 09-50-11

But I'm unsure whether it's better this way, or if it's better to show only the name without the "On duty:" comment

library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -50,23 +51,23 @@ public function getColor(int $transparency): string
...
-     * @param ValidHtml $content
+     * @param HtmlDocument $content
      *
      * @return static
      */
-    public function setFlyoutContent(ValidHtml $content): static
+    public function setFlyoutContent(HtmlDocument $content): static
     {
-        $this->flyoutContent = $content;
+        $this->flyoutContent = clone $content;
```
nilmerg: With the new widget, consider to avoid cloning it here again. Think about a factory method, since the current member is only known later on anyway. Assembling an entry is already responsible to inject the flyout content, so I'd suggest to add a method like `withCurrentMember(Member $member): self` and inject its result instead. (Still inside `assembleContainer`)

library/Notifications/Widget/Timeline/Entry.php
```diff
@@ -19,8 +17,8 @@ class Entry extends TimeGrid\Entry
     /** @var Member */
     protected $member;
 
-    /** @var ?HtmlDocument Content of the flyoutmenu that is shown when the entry is hovered */
-    protected ?HtmlDocument $flyoutContent;
+    /** @var ?EntryFlyout Content of the flyoutmenu that is shown when the entry is hovered */
+    protected ?EntryFlyout $flyoutContent;
```
nilmerg: If it's nullable type, it must be initialized explicitly with null, as of PHP 8.4

library/Notifications/Widget/Timeline/Rotation.php
```diff
@@ -80,200 +73,25 @@ public function getPriority(): int
...
-        } else {
-            $startText = $this->translate('Started on %s');
-        }
+        $flyout = new EntryFlyout();
+        $flyout->setMode($this->model->mode)
+            ->setRotationMembers($rotationMembers)
+            ->setRotationOptions($this->model->options)
+            ->setRotationName($this->model->name)
+            ->setFirstHandoff($this->model->first_handoff)
+            ->generateAndSetRotationInfo();
```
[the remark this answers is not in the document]
BastianLedererIcinga: Running FlyoutEntry::generateAndSetRotationInfo() in the FlyoutEntry::assemble() has the disadvantage that it will be run once per entry, though once per rotation is enough.
But you are probably right about it being displaced in Rotation.php. I moved it to FlyoutEntry::withActiveMember(), since that is already required before assembling the FlyoutEntry, this way we can keep running FlyoutEntry::generateAndSetRotationInfo() only once per rotation.

library/Notifications/Widget/Timeline/Rotation.php
```diff
@@ -96,6 +120,7 @@ public function fetchTimeperiodEntries(DateTime $after, DateTime $until): Genera
                     Filter::greaterThanOrEqual('until_time', $after) // Or one which isn't over yet
                 )
             ));
+
```
nilmerg: please refrain from changing style of code unrelated to your change

nilmerg (changes requested):

PR #375, opened 2025-10-24
Cleanup access control

### LINKED ISSUE #322: Roadmap: Improve view contact-permissions for restricted users
I would like to put some kind of permission on the roadmap similar to the `no-monitoring/contacts` in the monitoring module to allow restricted icingaweb2-users to see notification-contact-names but not the target (e.g. Pager or Email field).
Image

In older versions restricted users were allowed to see the contact-name, but when they opened the view in `/icingaweb2/monitoring/show/contact?contact_name=` they received a "no permission" response, which imho is a good option from a GDPR-perspective. 
Starting with [icingaweb2-v2.8.0](https://github.com/Icinga/icingaweb2/issues/3973) (I believe) the behavior in the module changed.

Image

nilmerg:
resolves #322

### DIFF
--- a/configuration.php
+++ b/configuration.php
@@ -74,10 +74,10 @@ $this->providePermission(
     $this->translate('Allow to modify configuration via API')
 );
 
-$this->provideRestriction(
-    'notifications/filter/objects',
-    $this->translate('Restrict access to the objects that match the filter')
-);
+//$this->provideRestriction(
+//    'notifications/filter/objects',
+//    $this->translate('Restrict access to the objects that match the filter')
+//);
 
 $this->provideConfigTab(
     'database',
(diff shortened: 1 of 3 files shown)


library/Notifications/Common/ConfigurationTabs.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/* Icinga Notifications Web | (c) 2023 Icinga GmbH | GPLv2 */
```
sukhwinder33445: 2025.

application/controllers/SchedulesController.php
```diff
@@ -17,15 +18,20 @@
...
 
 class SchedulesController extends CompatController
 {
+    use ConfigurationTabs;
     use SearchControls;
 
     /** @var Filter\Rule Filter from query string parameters */
     private $filter;
 
+    public function init()
```
sukhwinder33445: Please add return type.

sukhwinder33445 (changes requested):

sukhwinder33445 (approved):

PR #379, opened 2025-10-30
Use new decorators in forms

sukhwinder33445:
Requires:

-  https://github.com/Icinga/ipl-web/pull/327

### DIFF
--- a/application/forms/ContactGroupForm.php
+++ b/application/forms/ContactGroupForm.php
@@ -29,10 +29,10 @@ class ContactGroupForm extends CompatForm
     use CsrfCounterMeasure;
 
     /** @var Connection */
-    private $db;
+    private Connection $db;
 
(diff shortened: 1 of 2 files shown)


library/Notifications/Web/Form/ContactForm.php
```diff
@@ -98,49 +94,46 @@ protected function assemble()
...
-                new HtmlElement(
-                    'p',
-                    new Attributes(['class' => 'description']),
-                    new Text($this->translate(
-                        'Use this to associate actions in the UI, such as incident management, with this contact.'
-                        . ' To successfully receive desktop notifications, this is also required.'
-                    ))
-                )
-            );
+        $usernameEl = $this->createElement(
```
nilmerg: Register the element loader in the fieldset instead please

Issue #371, opened 2025-10-21
SQL Error when attempting to remove escalation


BastianLedererIcinga:
When trying to delete any but the lowest escalation of an Event Rule, pressing the Save button will cause an SQL Error to be dispalyed.

In this Example
<img width="2279" height="909" alt="Image" src="https://github.com/user-attachments/assets/23e00e51-34a2-48ef-8691-f48bb8c628ad" />

trying to delete the escalation with Contact b causes this result
<img width="2279" height="909" alt="Image" src="https://github.com/user-attachments/assets/0a873d27-3472-4e58-a956-e78ada77c27a" />

PR #389, opened 2025-11-04
delete escalations before updating positions

BastianLedererIcinga:
fix: #371 
Before updating escalation positions, any escalations removed by the user must be marked as deleted in the database. This prevents duplicate (rule_id, position) entries in the rule_escalation table, which previously caused an SQL error.

### DIFF
--- a/application/forms/EventRuleConfigForm.php
+++ b/application/forms/EventRuleConfigForm.php
@@ -311,9 +311,9 @@ class EventRuleConfigForm extends CompatForm
         $escalations = $this->getElement('escalations')->getEscalations();
         $remainingDbEscalations = [];
         $escalationConfigs = [];
-        foreach ($escalations as $escalation) {
+        foreach ($escalations as $index => $escalation) {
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):

application/forms/EventRuleConfigForm.php
```diff
@@ -309,14 +309,18 @@ public function storeInDatabase(Connection $db, ?Rule $previousRule): int
...
-            function ($escalation) {
-                return $escalation->getEscalation()['id'];
-            },
-            $escalations
-        );
+        $remainingDbEscalations = [];
+        $escalationConfigs = [];
+        foreach ($escalations as $escalation) {
+            $config = $escalation->getEscalation();
+            $escalationConfigs[] = $config;
```
nilmerg: it's just slight safe-guard, but since you address the entries in `$escalationConfigs` by the position of an escalation in `$escalations` below, you should also use the same position when inserting them here.

Issue #366, opened 2025-10-13
ScheduleForm: Shift display bug


flourish86:
I found this as a reproducible glitchy behavior.

To reproduce:

* Create a Multiday Rotation, whose end day is the same as "today"
* Set Rotation Start to "Today"
* Add Rotation 

In the resulting view, there's an additional shift rendered, that does not make sense.
<img width="1281" height="1026" alt="Image" src="https://github.com/user-attachments/assets/6b1fd160-09b6-40d8-bdae-7f872aad305a" />

Here's the configuration for the Rotation

<img width="719" height="629" alt="Image" src="https://github.com/user-attachments/assets/a1a6baa3-aafc-42fc-959f-29ac908c0b8e" />

PR #393, opened 2025-11-05
fix handling of multi day rotation entries that end the same day they started

BastianLedererIcinga:
fix: #366
Rotations that have their start on the same day as their to-day, will now have their first entries created correctly.
Screenshot from 2025-11-05 14-37-19

### DIFF
--- a/application/forms/RotationConfigForm.php
+++ b/application/forms/RotationConfigForm.php
@@ -1431,7 +1431,8 @@ class RotationConfigForm extends CompatForm
                 if (
                     $fromDay < $toDay && ($firstHandoffDay < $fromDay || $firstHandoffDay > $toDay)
                     || $toDay < $fromDay && ($firstHandoffDay < $fromDay && $firstHandoffDay > $toDay)
-                    || $firstHandoffDay === $toDay && $toDay !== $fromDay && $options["to_at"] === "00:00"
+                    || $firstHandoffDay === $toDay && $toDay !== $fromDay
+                    && $firstHandoff >= $this->parseDateAndTime($this->getValue('first_handoff'), $options['to_at'])
                 ) {
                     // Normalize the first handoff to the first day of the shift in case it's outside the range
                     $firstHandoff->add(new DateInterval(sprintf(


application/forms/RotationConfigForm.php
```diff
@@ -1416,6 +1416,8 @@ private function yieldRecurrenceRules(int $count): Generator
                 $fromDay = (int) $options['from_day'];
                 $toDay = (int) $options['to_day'];
                 $interval = (int) $options['interval'];
+                $fromAt = new DateTime($options['from_at']);
+                $toAt = new DateTime($options['to_at']);
```
nilmerg: In light of the current development regarding timezones, I'd love to see a solution that doesn't involve datetime objects, unless you really need them. As of right now, I think numerical comparisons might suffice.

application/forms/RotationConfigForm.php
```diff
@@ -1431,6 +1431,7 @@ private function yieldRecurrenceRules(int $count): Generator
                 if (
                     $fromDay < $toDay && ($firstHandoffDay < $fromDay || $firstHandoffDay > $toDay)
                     || $toDay < $fromDay && ($firstHandoffDay < $fromDay && $firstHandoffDay > $toDay)
+                    || $firstHandoffDay === $toDay && $options["to_at"] === "00:00"
```
nilmerg: That's the wrong fix as discussed. The previous error, the entry without width, must be fixed so that the entry has a width, and does not disappear.

application/forms/RotationConfigForm.php
```diff
@@ -1431,6 +1431,7 @@ private function yieldRecurrenceRules(int $count): Generator
                 if (
                     $fromDay < $toDay && ($firstHandoffDay < $fromDay || $firstHandoffDay > $toDay)
                     || $toDay < $fromDay && ($firstHandoffDay < $fromDay && $firstHandoffDay > $toDay)
+                    || $firstHandoffDay === $toDay && $toDay !== $fromDay && $options["to_at"] === "00:00"
```
nilmerg:
```suggestion
                    || $firstHandoffDay === $toDay && (
                        $toDay !== $fromDay
                        && $firstHandoff >= $this->parseDateAndTime($this->getValue('first_handoff'), $options['to_at'])
                    )
```

nilmerg (changes requested):

nilmerg (approved):

Issue #397, opened 2025-11-06
Enhance timeline flyout behavior


nilmerg:
At the moment the new and shiny flyouts appear faster than the default titles previously used. While this was one of the desired advantages, it now hinders access to other content they cover. If the user's goal was to open the flyout, that's fine. It's not when the goal was something entirely different. Such as the desire to push the button to add a rotation, or to click on an entry close to others.

The problem is the vicinity an entry is in to other entries or the add rotation button. The flyout appears immediately upon entering the hover area of an entry. If said entry is small or happens to start near something else, that something else is difficult to reach due to the flyout.

To solve this, the flyout must not appear immediately. But it should still appear somewhat faster than the default titles.

The common approach with JS is to delay and cancel the flyout if required. To do so, wrap the instruction to show the flyout in a `setTimeout()` call. Start this upon entering an entry's hover area. `setTimeout` returns an id, use it in conjunction with `clearTimeout` to cancel it. Do so in the `mouseleave` listener.

PR #400, opened 2025-11-10
add a delay before showing the flyout

BastianLedererIcinga:
resolves: #397

### DIFF
--- a/public/js/schedule.js
+++ b/public/js/schedule.js
@@ -14,11 +14,12 @@
         {
             super(icinga);
 
+            this.activeTimeout = null;
+
             this.on('rendered', '#main > .container', this.onRendered, this);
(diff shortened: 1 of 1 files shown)


public/js/schedule.js
```diff
@@ -123,7 +127,11 @@
             relatedEntries.forEach(element => element.classList.remove('highlighted'));
 
             if (tooltip) {
-                tooltip.classList.remove('is-left', 'is-bottom');
+                if (this.activeTimeout) {
+                    clearTimeout(this.activeTimeout);
+                }
+
+                tooltip.classList.remove('is-left', 'is-bottom', 'entry-is-hovered');
```
nilmerg: can be in the else case, right? I mean, the flyout does not show up if the timeout is cancelled.

Issue #398, opened 2025-11-07
Source: Configurable Username


oxzi:
As proposed in Icinga/icinga-notifications#227, we want to add usernames for sources to enable a more user-friendly authentication for the Icinga Notifications API. At the moment, the username is basically the source ID in the `source-${ID}` format.

With the proposed change and implementation in Icinga/icinga-notifications#338, each source gets its own unique username.

For the web part, the source UI needs another "listener user" text form, populating the `listener_user` column of the `source` table. Details to the schema are available in the diff of Icinga/icinga-notifications#338.

PR #401, opened 2025-11-11
Add input field for listener_username

BastianLedererIcinga:
resolves: #398

### DIFF
--- a/application/forms/SourceForm.php
+++ b/application/forms/SourceForm.php
@@ -77,7 +77,7 @@ class SourceForm extends CompatForm
             'listener_username',
             [
                 'required' => true,
-                'label' => $this->translate('Listener Username'),
+                'label' => $this->translate('Username'),
                 'validators' => [new CallbackValidator(
                     function ($value, CallbackValidator $validator) {
                         // Username must be unique


nilmerg (changes requested):

application/forms/SourceForm.php
```diff
@@ -66,23 +66,35 @@ protected function assemble(): void
...
+            'listener_username',
+            [
+                'required' => true,
+                'label' => $this->translate('Listener Username'),
+                'validators' => [new CallbackValidator(
+                    function ($value, CallbackValidator $validator) {
+                        // Deleted rows are included to avoid integrity constraint violations
+                        $source = Source::on($this->db)
+                            ->filter(Filter::equal('listener_username', $value))
+                            ->filter(Filter::any(Filter::equal('deleted', 'y'), Filter::equal('deleted', 'n')))
```
nilmerg: Huh, I wasn't aware of that (https://github.com/Icinga/icinga-notifications/pull/338#issuecomment-3520965617) and I think we should set `listener_username` to `NULL` upon deletion.

nilmerg (changes requested):

nilmerg (approved):

PR #404, opened 2025-11-13
schedule: Improve contrast of scale tick marks

### LINKED ISSUE #336: Schedule Detail: Improve contrast of scale tick marks
One thing that we missed in #283 is the contrast of the tick marks.

In the usability test, users clearly couldn't recognize the tick marks, without hinting at them and scaling the viewport.

Making them better recognizable should improve the connection between the time scale and the time labels and therefore improve the recognizability of the temporal direction.

nilmerg:
image

fixes #336


flourish86:
I assume this work should make them prominent enough. The missing leftmost tick doesn't make a difference, I guess.

It's not a big deal, but one thing I noticed is that there's a gap between the ticks and the grid.

Screenshot 2025-11-13 at 17 08 33

This could be solved with `margin-top: -1px` on the `.timescale` element, but I'm not sure, if you find this too hacky:

Screenshot 2025-11-13 at 17 11 38

flourish86 (approved):

PR #413, opened 2025-11-14
Calculate the days since latest handoff correctly

### LINKED ISSUE #412: End of older entries are cut off
All screenshots were taken **on 2025-11-14 at 14:50**. All of the rotations started one or two weeks in the past.

## Bug description

All of the rotations should have entries that end on the current day. But none of them are visible.

Image

Multi:

Image

Partial:

Image

24/7:

Image

## Expected behavior

Image

jrauh01:
## Problem

If the **last handoff ends today** and the **next handoff did already happen today**, the days since the latest handoff are `0`. That is correct, because the latest handoff happened today. Thus we get the current handoff as the first one. What we want is the handoff before.

## Solution

We now calculate the days since the latest handoff that did not happen **today**.

Checks:
- Check if there was an handoff before by checking if `days since the first handoff > 0`
- Check if `days since first handoff % $interval === 0`

If so set `$daysSinceLatestHandoff` to `$interval`.

fixes #412

### DIFF
--- a/library/Notifications/Widget/Timeline/Rotation.php
+++ b/library/Notifications/Widget/Timeline/Rotation.php
@@ -151,7 +151,12 @@ class Rotation
                 );
 
                 if ($timeperiodEntry->start_time < $before) {
-                    $daysSinceLatestHandoff = $timeperiodEntry->start_time->diff($before)->days % $interval;
+                    $daysSinceFirstHandoff = $timeperiodEntry->start_time->diff($before)->days;
+                    $daysSinceLatestHandoff = $daysSinceFirstHandoff % $interval;
+                    if ($daysSinceFirstHandoff > 0 && $daysSinceLatestHandoff === 0) {
+                        $daysSinceLatestHandoff = $interval;
+                    }
+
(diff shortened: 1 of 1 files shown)


library/Notifications/Widget/Timeline/Rotation.php
```diff
@@ -151,7 +151,11 @@ public function fetchTimeperiodEntries(DateTime $after, DateTime $until): Genera
                 );
 
                 if ($timeperiodEntry->start_time < $before) {
-                    $daysSinceLatestHandoff = $timeperiodEntry->start_time->diff($before)->days % $interval;
+                    $daysSinceFirstHandoff = $timeperiodEntry->start_time->diff($before)->days;
+                    $daysSinceLatestHandoff = $daysSinceFirstHandoff % $interval;
+                    if ($daysSinceFirstHandoff > 0 && $daysSinceLatestHandoff === 0) {
+                        $daysSinceLatestHandoff = $interval;
+                    }
```
nilmerg: empty line after this ;)

nilmerg (approved):

PR #414, opened 2025-11-14
Fix config tab parameter handling

nilmerg:
Interacting with controls like the search bar, must be persisted to the current tab. To achieve this, the tab must be activated before multipart updates are transmitted.

In case of reaching a configuration endpoint directly (following a link or reloading the entire page) the current tab's url must also reflect the current search query parameters, plus sort etc.

### DIFF
--- a/application/controllers/ContactsController.php
+++ b/application/controllers/ContactsController.php
@@ -127,8 +127,6 @@ class ContactsController extends CompatController
         if (! $searchBar->hasBeenSubmitted() && $searchBar->hasBeenSent()) {
             $this->sendMultipartUpdate();
         }
-
-        $this->setAutorefreshInterval(30);
     }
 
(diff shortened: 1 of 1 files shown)


application/controllers/ContactsController.php
```diff
@@ -126,9 +129,6 @@ public function indexAction()
         }
 
         $this->setAutorefreshInterval(30);
```
sukhwinder33445: Remove this please.

application/controllers/SchedulesController.php
```diff
@@ -34,6 +34,8 @@ public function init(): void
 
     public function indexAction(): void
     {
+        $this->getTabs()->activate('schedules');
```
sukhwinder33445: Please add missing `setTitle()`.

PR #423, opened 2025-11-17
Timeline: Make sure the header overlaps everything again

### LINKED ISSUE #399: Things are cut off when schedule is scrollable
## Description

If the schedule has enough rotations to be scrollable, things at the top are merging.

Image

Additionally the flyouts that open upward are cut off if they are at the upper border.

Image

## To reproduce

Add rotations till the schedule gets scrollable. Now scroll down, so the rotations are under the header. For the flyout problem you may need to add a few more rotations and scroll down further. Then hover over a rotation at the top.

## Expected behavior

The header should either scroll up or not be transparent. The flyouts that would be cut off should open downward instead of opening upward.

nilmerg:
Didn't succeed while trying to fix the flyout positioning. The problem is that since the header is a sticky element, it's not influencing the containers offsetHeight/scrollHeight and so there's no way to know for sure where the constraints of the box really are. But the workaround for the user should be straightforward anyway: Just scroll the content down again.

fixes #399

### DIFF
--- a/library/Notifications/Widget/TimeGrid/DaysHeader.php
+++ b/library/Notifications/Widget/TimeGrid/DaysHeader.php
@@ -63,6 +63,8 @@ class DaysHeader extends BaseHtmlElement
             $displayTimezone
         );
 
+        $this->addHtml(new HtmlElement('span', Attributes::create(['class' => 'spacer'])));
+
         for ($i = 0; $i < $this->days; $i++) {
             if ($time == $today) {
                 $title = [new HtmlElement(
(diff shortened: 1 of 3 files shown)


flourish86 (changes requested):
Great, that we noticed this. 🤩

One thing: I'd put a separator, so it's better distinguishable from the content when scrolled.

before
Screen Shot 2025-11-18 at 10 10 24

after
Screen Shot 2025-11-18 at 10 16 09

Should be easy
Screenshot 2025-11-18 at 10 20 35

PR #460, opened 2026-04-02
Ensure forward compatibility with ipl by adding return types

BastianLedererIcinga:
Ensure compatibility with the upcoming strict type additions in:
https://github.com/Icinga/ipl-orm/pull/158
https://github.com/Icinga/ipl-stdlib/pull/69
https://github.com/Icinga/ipl-html/pull/187

### DIFF
--- a/library/Notifications/Model/Behavior/IdTagAggregator.php
+++ b/library/Notifications/Model/Behavior/IdTagAggregator.php
@@ -28,8 +28,7 @@ class IdTagAggregator extends PropertyBehavior implements RewriteColumnBehavior,
         parent::__construct(['id_tags']);
     }
 
-    /** @return $this */
-    public function setQuery(Query $query)
+    public function setQuery(Query $query): static
     {
         $this->query = $query;
 
--- a/library/Notifications/Model/Behavior/IdTagAggregator.php
+++ b/library/Notifications/Model/Behavior/IdTagAggregator.php
@@ -99,7 +99,8 @@ class IdTagAggregator extends PropertyBehavior implements RewriteColumnBehavior,
     {
     }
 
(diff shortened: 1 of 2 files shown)


library/Notifications/Model/Behavior/ObjectTags.php
sukhwinder33445, line 30: The type is missing for $relation param.

library/Notifications/Model/Behavior/IdTagAggregator.php
```diff
@@ -100,7 +101,8 @@ public function rewriteColumnDefinition(ColumnDefinition $def, string $relation)
     {
     }
 
-    public function rewriteCondition(Filter\Condition $condition, $relation = null)
+    public function rewriteCondition(Filter\Condition $condition, $relation = null): null
```
sukhwinder33445, line 104: Add type for $relation please.

sukhwinder33445 (changes requested):

library/Notifications/Model/Behavior/IdTagAggregator.php
BastianLedererIcinga: That would break compatibility with existing ipl versions.

library/Notifications/Model/Behavior/ObjectTags.php
sukhwinder33445: Ahh, this is required to ensure backwards compatibility.

library/Notifications/Model/Behavior/IdTagAggregator.php
sukhwinder33445: Same

sukhwinder33445 (approved):

Issue #461, opened 2026-03-02
Let sources automate their initial setup and authentication


nilmerg:
At the moment, there are a few hurdles to overcome when activating notifications in a source. First, you have to create the source in Icinga Notifications and define access data (user and password). Next, the source itself must be configured so that it can authenticate itself with this access data in Icinga Notifications.

We want to give tightly integrated sources the option to configure themselves or to assist the user with domain specific configuration. Setting up the source's notification feature and its configuration in Icinga Notifications must be possible as a seamless step.

nilmerg:
:tada:

PR #469, opened 2026-04-28
Handle rule filter serialization

### LINKED ISSUE #466: Handle rule filter rendering and serialization on our own again
With the advent of *Attribute Negotiation* in Icinga Notifications, the current rule filter definition has to change again.

Goals:

* The filter target selection is obsolete and must be dropped
  * The modal therefore immediately shows a filter editor
* Source integrations are still expected to set up said editor
  * Their responsibilities are the suggestion, enrichment and validation
* Serialization of filters is equally done for all sources so that Icinga Notifications understands it
  * The format to use must still be chosen in conjunction with whoever implements https://github.com/Icinga/icinga-notifications/issues/406

BastianLedererIcinga:
resolve #466 

Introduce `RuleSerializer` to create Json from a configured filter.

Introduce `Icinga\Module\Notifications\Hook\V2` which no longer requires `serializeRuleFilter()` and `getRuleFilterTargets()` as they are now obsolete.

Adjust `EventRuleController` to skip target selection use the new `RuleSerializer`.

require https://github.com/Icinga/ipl-web/pull/380
require https://github.com/Icinga/ipl-web/pull/383
require https://github.com/Icinga/ipl-stdlib/pull/75
require https://github.com/Icinga/ipl-stdlib/pull/76

### DIFF
--- a/application/controllers/EventRuleController.php
+++ b/application/controllers/EventRuleController.php
@@ -7,6 +7,7 @@ namespace Icinga\Module\Notifications\Controllers;
 
 use Icinga\Application\Hook;
 use Icinga\Application\Logger;
+use Icinga\Exception\ConfigurationError;
 use Icinga\Exception\Http\HttpNotFoundException;
 use Icinga\Module\Notifications\Common\Auth;
 use Icinga\Module\Notifications\Common\Database;
@@ -14,20 +15,25 @@ use Icinga\Module\Notifications\Common\Links;
 use Icinga\Module\Notifications\Forms\EventRuleConfigElements\NotificationConfigProvider;
 use Icinga\Module\Notifications\Forms\EventRuleConfigForm;
 use Icinga\Module\Notifications\Forms\EventRuleForm;
-use Icinga\Module\Notifications\Hook\V1\SourceHook;
+use Icinga\Module\Notifications\Hook\V2\SourceHook;
 use Icinga\Module\Notifications\Model\Rule;
 use Icinga\Module\Notifications\Model\Source;
+use Icinga\Module\Notifications\Util\RuleParser;
+use Icinga\Module\Notifications\Util\RuleSerializer;
 use Icinga\Module\Notifications\Web\Control\SearchBar\ExtraTagSuggestions;
 use Icinga\Web\Notification;
 use Icinga\Web\Session;
 use ipl\Html\Contract\Form;
 use ipl\Html\Html;
 use ipl\Stdlib\Filter;
+use ipl\Stdlib\Filter\Condition;
 use ipl\Web\Compat\CompatController;
-use ipl\Web\Compat\CompatForm;
+use ipl\Web\Control\SearchBar\SearchException;
+use ipl\Web\Control\SearchEditor;
 use ipl\Web\Url;
 use ipl\Web\Widget\Icon;
 use ipl\Web\Widget\Link;
+use JsonException;
 use Psr\Http\Message\ServerRequestInterface;
 use Throwable;
 
@@ -195,7 +201,71 @@ class EventRuleController extends CompatController
     {
         $ruleId = (int) $this->params->getRequired('id');
         $filter = $this->params->get('object_filter', $this->session->get('object_filter'));
+        $hook = $this->resolveSourceHook($ruleId);
 
+        if ($filter) {
+            try {
+                $parsedFilter = (new RuleParser())->parseJson($filter);
+                $applyLabels = function (Filter\Rule $rule) use ($hook, &$applyLabels): void {
+                    if ($rule instanceof Filter\Chain) {
+                        foreach ($rule as $child) {
+                            $applyLabels($child);
+                        }
+                    } else {
+                        /** @var Condition $rule */
+                        $hook->enrichCondition($rule);
+                    }
+                };
+
+                $applyLabels($parsedFilter);
+            } catch (JsonException $e) {
+                Logger::error('Failed to parse rule filter configuration: %s (Error: %s)', $filter, $e);
+                throw new ConfigurationError($this->translate(
+                    'Failed to parse rule filter configuration. Please contact your system administrator.'
+                ));
+            }
+        }
+
+        $editor = (new SearchEditor())
+            ->setFilter($parsedFilter ?? new Filter\All())
+            ->setSuggestionUrl(
+                Url::fromPath(
+                    'notifications/event-rule/suggest',
+                    ['id' => $ruleId, '_disableLayout' => true, 'showCompact' => true]
+                )
+            )
+            ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl())
+            ->setMetadataFields(['customvarid'])
+            ->on(
+                SearchEditor::ON_VALIDATE_COLUMN,
+                function (Condition $condition) use ($hook) {
+                    if (! $hook->isValidCondition($condition)) {
+                        throw new SearchException($this->translate('Is not a valid column'));
+                    }
+
+                    $condition->metaData()->set('jsonPath', $hook->getJsonPath($condition));
+                }
+            )
+            ->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $hook) {
+                $this->session->set('object_filter', (new RuleSerializer($form->getFilter()))->getJson());
+                $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
+            })
+            ->handleRequest($this->getServerRequest());
+
+        $this->getDocument()->addHtml($editor);
+
+        $this->setTitle($this->translate('Adjust Filter'));
+    }
+
+    public function suggestAction(): void
+    {
+        $hook = $this->resolveSourceHook((int) $this->params->getRequired('id'));
+        $suggestions = $hook->getSuggestions();
+        $this->getDocument()->addHtml($suggestions->forRequest($this->getServerRequest()));
+    }
+
+    protected function resolveSourceHook(int $ruleId): SourceHook
+    {
         $source = null;
         if ($ruleId !== -1) {
             $source = Rule::on(Database::get())
@@ -217,7 +287,7 @@ class EventRuleController extends CompatController
         }
 
         $hook = null;
-        foreach (Hook::all('Notifications/v1/Source') as $h) {
+        foreach (Hook::all('Notifications/v2/Source') as $h) {
             /** @var SourceHook $h */
             try {
                 if ($h->getSourceType() === $source->type) {
@@ -237,51 +307,7 @@ class EventRuleController extends CompatController
             ), $source->type));
         }
 
-        if (! $filter) {
-            $targets = $hook->getRuleFilterTargets($source->id);
-            if (count($targets) === 1 && ! is_array(reset($targets))) {
-                $filter = key($targets);
-            } else {
-                $target = null;
-                $form = (new CompatForm())
-                    ->applyDefaultElementDecorators()
-                    ->setAction(Url::fromRequest()->getAbsoluteUrl())
-                    ->addElement('select', 'target', [
-                        'required' => true,
-                        'label' => $this->translate('Filter Target'),
-                        'options' => ['' => ' - ' . $this->translate('Please choose') . ' - '] + $targets,
-                        'disabledOptions' => ['']
-                    ])
-                    ->addElement('submit', 'btn_submit', [
-                        // translators: shown on a submit button to proceed to the next step of a form wizard
-                        'label' => $this->translate('Next')
-                    ])
-                    ->on(Form::ON_SUBMIT, function (CompatForm $form) use (&$target) {
-                        $target = $form->getValue('target');
-                    })
-                    ->handleRequest($this->getServerRequest());
-
-                if ($target !== null) {
-                    $filter = $target;
-                } else {
-                    $this->addContent($form);
-                }
-            }
-        }
-
-        if ($filter) {
-            $form = $hook->getRuleFilterEditor($filter)
-                ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl())
-                ->on(Form::ON_SUBMIT, function (Form $form) use ($ruleId, $hook) {
-                    $this->session->set('object_filter', $hook->serializeRuleFilter($form));
-                    $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
-                })
-                ->handleRequest($this->getServerRequest());
-
-            $this->getDocument()->addHtml($form);
-        }
-
-        $this->setTitle($this->translate('Adjust Filter'));
+        return $hook;
     }
 
     public function editAction(): void
--- a/application/controllers/EventRuleController.php
+++ b/application/controllers/EventRuleController.php
@@ -235,7 +235,7 @@ class EventRuleController extends CompatController
                 )
             )
             ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl())
-            ->setMetadataFields(['customvarid'])
+            ->setMetadataFields($hook->getMetadataKeys())
             ->on(
                 SearchEditor::ON_VALIDATE_COLUMN,
                 function (Condition $condition) use ($hook) {
@@ -247,7 +247,10 @@ class EventRuleController extends CompatController
                 }
             )
             ->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $hook) {
-                $this->session->set('object_filter', (new RuleSerializer($form->getFilter()))->getJson());
+                $this->session->set(
+                    'object_filter',
+                    (new RuleSerializer($form->getFilter(), $hook->getMetadataKeys()))->getJson()
+                );
                 $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
             })
             ->handleRequest($this->getServerRequest());
--- a/application/controllers/EventRuleController.php
+++ b/application/controllers/EventRuleController.php
@@ -5,29 +5,37 @@
 
 namespace Icinga\Module\Notifications\Controllers;
 
-use Icinga\Application\Hook;
 use Icinga\Application\Logger;
+use Icinga\Exception\ConfigurationError;
 use Icinga\Exception\Http\HttpNotFoundException;
 use Icinga\Module\Notifications\Common\Auth;
 use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Common\Links;
+use Icinga\Module\Notifications\Common\SourceHookLocator;
 use Icinga\Module\Notifications\Forms\EventRuleConfigElements\NotificationConfigProvider;
 use Icinga\Module\Notifications\Forms\EventRuleConfigForm;
 use Icinga\Module\Notifications\Forms\EventRuleForm;
-use Icinga\Module\Notifications\Hook\V1\SourceHook;
+use Icinga\Module\Notifications\Hook\V2\SourceHook;
 use Icinga\Module\Notifications\Model\Rule;
 use Icinga\Module\Notifications\Model\Source;
+use Icinga\Module\Notifications\Util\RuleSerializer;
 use Icinga\Module\Notifications\Web\Control\SearchBar\ExtraTagSuggestions;
 use Icinga\Web\Notification;
 use Icinga\Web\Session;
 use ipl\Html\Contract\Form;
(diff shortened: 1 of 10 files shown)


library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return array{op: string, column: string, value: mixed}
+     */
+    protected function serializeCondition(Filter\Condition $condition): array
+    {
+        return [
+            'op' => match (true) {
+                $condition instanceof Filter\Unlike => '!~',
+                $condition instanceof Filter\Unequal => '!=',
+                $condition instanceof Filter\Like => '~',
+                $condition instanceof Filter\Equal => '=',
```
nilmerg: I noticed that Icinga Notifications will interpret the like/unlike expressions as regex internally. My [suggestion](https://github.com/Icinga/icinga-notifications/pull/418#issuecomment-4419842201) there is to remove the distinction between like/equal and unlike/unequal, as this was only the case in Web because our query string representation needed a way to express this to avoid the need to distinguish such based on whether an asterisk (*) is contained in the value or not. Now that we don't rely on this anymore here, let's just pre-process the value of a like/unlike as regex and serialize it as `regex` instead of `value` and use the non-like operator counterpart instead.

Pre-process in this case means, `foo*` will be prepared as `^foo.*$`.

library/Notifications/Hook/V2/SourceHook.php
```diff
@@ -0,0 +1,91 @@
...
+    public function enrichCondition(Condition $condition): void;
+
+    /**
+     * Get all jsonPaths for the condition's column
+     *
+     * @param Condition $condition
+     *
+     * @return array<string>
+     */
+    public function getJsonPaths(Condition $condition): array;
```
nilmerg: This forces integrations to implement a cache of sorts to prevent issuing unnecessary queries in case a database is required. Please pass over all conditions at once.

library/Notifications/Hook/V2/SourceHook.php
```diff
@@ -0,0 +1,91 @@
...
+    public function getColumnSuggestions(string $searchTerm): Traversable;
+
+    /**
+     * Get whether a relation label should be added to the suggestion
+     *
+     * @param string $column
+     *
+     * @return bool
+     */
+    public function shouldShowRelationFor(string $column): bool;
```
nilmerg: That's for the integration to decide. Remove this.

library/Notifications/Model/Source.php
```diff
@@ -103,7 +103,7 @@ public function getIcon(): Icon
         // Fallback, in case an integration is inactive or missing
         $icon = new Icon('share-nodes');
 
-        foreach (Hook::all('Notifications/v1/Source') as $hook) {
+        foreach (Hook::all('Notifications/v2/Source') as $hook) {
```
nilmerg: I think there's a way to get rid of this loop and `getSourceType()` in the interface. Right now there can only be a single hook anyway, and that's where `Hook::first()` comes into play.

Please try to incorporate the source type in the name of the hook for v2. In case of icinga, that'd be: `Notifications/v2/IcingaSource`

The hook can then be loaded this way:

```php
$hook = Hook::first(sprintf('Notifications/v2/%sSource', \ipl\Stdlib\Str::camel($this->type));
```

Please note that this fundamentally changes how sources are configured. This way the available integrations cannot be known beforehand (before an event came in) so the type field must be a free text field. This collides with #450 somewhat, but that's already in progress and based on this anyway.

Once #461 is a thing, the source configuration is mainly for generic sources as well. So it doesn't hurt either.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Serialize the filter as Json
+     *
+     * @return string
+     *
+     * @throws JsonEncodeException
+     */
+    public function getJson(): string
+    {
+        $result = ['qs' => (new Renderer($this->filter))->render()];
```
nilmerg: Please use `QueryString::render()`

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,121 @@
...
+        foreach ($chain as $rule) {
+            if ($rule instanceof Chain) {
+                $rules[] = $this->serializeChain($rule);
+            } else {
+                $rules[] = $this->serializeCondition($rule);
+            }
+        }
+
+        $result['rules'] = $rules;
+        return $result;
```
nilmerg, line 92: an empty line before `return` pls, always unless it's the sole statement in a block.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Get the preprocessed value of a condition
+     *
+     * Creates a regex for Like and Unlike rules
+     *
+     * @param string $value
+     *
+     * @return string
+     */
+    protected function preprocessValue(string $value): string
```
nilmerg: `asRegularExpression()` 
`toRegularExpression()` 
`createRegularExpression()` 
🤔

nilmerg (changes requested):
Didn't look at the controller or suggestions yet.

library/Notifications/Model/Source.php
BastianLedererIcinga: As discussed, a class alias for `SourceHook` is created at runtime, so that the validation of the hook by Icinga Web succeeds.

library/Notifications/Common/SourceHookLocator.php
```diff
@@ -0,0 +1,38 @@
...
+    public static function forType(string $type): ?SourceHook
+    {
+        $name = ucfirst(Str::camel($type));
+
+        $alias = 'Icinga\Module\Notifications\Hook\V2\\' . $name . 'SourceHook';
+        if (! interface_exists($alias)) {
+            class_alias(SourceHook::class, $alias);
+        }
+
+        return Hook::first('Notifications/v2/' . $name . 'Source');
```
nilmerg: Please make sure to use backslashes and always double them regardless of necessity.

application/controllers/EventRuleController.php
```diff
@@ -190,13 +195,70 @@ public function completeAction(): void
...
+                    'notifications/event-rule/suggest',
+                    ['id' => $ruleId, '_disableLayout' => true, 'showCompact' => true]
+                )
+            )
+            ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl())
+            ->on(
+                SearchEditor::ON_VALIDATE_COLUMN,
+                function (Condition $condition) use ($hook) {
+                    if (! $hook->isValidCondition($condition)) {
+                        throw new SearchException($this->translate('Is not a valid column'));
```
nilmerg: The hook may (or rather [will](https://github.com/Icinga/icingadb-web/pull/1369/changes/37d7a17af32060453800bede4e49ae75f4ebcf74#diff-2868454a27c87e7589e43d334ef65b0cad882431386688cacef2633e78764c8eR98)) want to provide a customized message. Let's change this to `assertValidCondition` and expect implementations to throw `SearchException`. But please catch any other error, log it, and throw a generic `SearchException` instead.

Our guideline with hooks is that they shouldn't break the view using them. So the view here is the modal and it should open and allow the user to view the filter and allow to reset it, even though the hook is broken.

application/forms/SourceForm.php
```diff
@@ -88,14 +67,12 @@ protected function assemble(): void
             ]
         );
         $this->addElement(
-            'select',
+            'text',
             'type',
             [
                 'required'          => true,
                 'label'             => $this->translate('Source Type'),
                 'class'             => 'autosubmit',
```
nilmerg: By now this should not automatically submit anymore, it was already questionable before this, but now given it's a text field definitely.

library/Notifications/Hook/V2/SourceHook.php
```diff
@@ -0,0 +1,75 @@
...
+     * @return void
+     */
+    public function enrichCondition(Condition $condition): void;
+
+    /**
+     * Get all JsonPaths for all given columns, keyed by the column
+     *
+     * @param string ...$columns
+     *
+     * @return array<array<string>>
```
nilmerg: *keyed by the column* is rather `array<string, string[]>`

library/Notifications/Hook/V2/SourceHook.php
```diff
@@ -0,0 +1,75 @@
...
+    public function getJsonPaths(string ...$columns): array;
+
+    /**
+     * Get suggestions for a value field
+     *
+     * @param string $column
+     * @param string $searchTerm
+     * @param Chain $searchFilter
+     *
+     * @return Traversable Values to be suggested as `search` => `label`
```
nilmerg: > `search` => `label`

That's not true. Take a look at `SearchSuggestions` to see what it expects.

library/Notifications/Hook/V2/SourceHook.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function getValueSuggestions(string $column, string $searchTerm, Chain $searchFilter): Traversable;
+
+    /**
+     * Get suggestions for a column field
+     *
+     * @param string $searchTerm
+     *
+     * @return Traversable Columns to be suggested as `search` => `label`
```
nilmerg: same here

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,127 @@
...
+            $condition instanceof Filter\LessThanOrEqual => '<=',
+        };
+
+        $value = $condition instanceof Filter\Like || $condition instanceof Filter\Unlike
+            ? ['regex' => $this->createRegularExpression($condition->getValue())]
+            : ['value' => $condition->getValue()];
+
+        return [
+            'op' => $op,
+            'attributes' => $this->jsonPaths[$condition->getColumn()],
```
nilmerg: Just in case, if a hook didn't provide a json path for a column, this throws a generic index error. We should throw a descriptive exception in this case.

library/Notifications/Web/Control/SearchEditor/RuleFilterSuggestions.php
nilmerg, line 1: as I told in person, this implementation is way too fragile to keep. `SearchSuggestions` should be adjusted in order to provide parsed request data instead of extending it.

nilmerg (changes requested):

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Serialize the filter as Json
+     *
+     * @return string
+     *
+     * @throws JsonEncodeException
+     */
+    public function getJson(): string
```
nilmerg: Please include `version` as key and set it to `2`. The rule controller then must verify to only try to evaluate such with this version. (throw otherwise) This is a first attempt to handle backwards compat, but only the part that there is a differentiation possible.

application/controllers/EventRuleController.php
```diff
@@ -190,13 +197,120 @@ public function completeAction(): void
...
+                throw new ConfigurationError($this->translate(
+                    'Failed to parse rule filter configuration. Please contact your system administrator.'
+                ));
+            }
+
+            $version = $parsedFilter['version'] ?? null;
+            if ($version !== RuleSerializer::VERSION) {
+                Logger::error(
+                    'Unsupported rule filter version: %s (expected %d)',
+                    var_export($version, true),
```
nilmerg: `var_export`?

application/controllers/EventRuleController.php
```diff
@@ -190,13 +197,120 @@ public function completeAction(): void
...
+                Logger::error('Failed to parse rule filter configuration: %s (Error: %s)', $filter, $e);
+                throw new ConfigurationError($this->translate(
+                    'Failed to parse rule filter configuration. Please contact your system administrator.'
+                ));
+            }
+
+            $version = $parsedFilter['version'] ?? null;
+            if ($version !== RuleSerializer::VERSION) {
+                Logger::error(
+                    'Unsupported rule filter version: %s (expected %d)',
```
nilmerg: Please put yourself into the administrator's position for a moment:

User:
> There's a rule that breaks the UI if I access the filter! It says I should contact you.

You look into the log:
> lots of messages

You ask:
> When was this, approximately?

User:
> 15min ago or so

You look into the log again and by chance you find:
> Unsupported rule filter version: 1 (expected 2)

How do you tell which rule is affected?

Bonus: How do you know that this is not under your control but related to the recent upgrade?

--

That's a slight exaggeration that shows how useless unhelpful log messages are. Please include the rule's id. The *bonus* is not required to solve, as this hopefully will be tackled in another issue implementing the actual migration step.

application/controllers/EventRuleController.php
```diff
@@ -190,13 +197,120 @@ public function completeAction(): void
...
+                /** @var Filter\Chain $searchFilter */
+                $searchFilter = QueryString::parse($requestData['searchFilter'] ?? '');
+                $provider = $hook->getValueSuggestions($column, $label, $searchFilter);
+            }
+        }
+
+        $suggestions = (new SearchSuggestions($provider))
+            ->setSearchTerm($label)
+            ->setOriginalSearchValue($requestData['term']['search'] ?? '')
+            ->setExcludeTerms($requestData['exclude'] ?? []);
```
nilmerg: The search editor has no exclude handling and won't have one either I suppose, so please drop this.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,143 @@
...
+use RuntimeException;
+
+class RuleSerializer
+{
+    public const VERSION = 2;
+
+    /** @var Filter\Condition|Filter\Chain */
+    protected Filter\Rule $filter;
+
+    /** @var array<array<string>> JSON paths keyed by column name */
```
nilmerg: that's still incorrect

nilmerg (changes requested):

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,143 @@
...
+            $condition instanceof Filter\GreaterThanOrEqual => '>=',
+            $condition instanceof Filter\LessThanOrEqual => '<=',
+        };
+
+        $value = $condition instanceof Filter\Like || $condition instanceof Filter\Unlike
+            ? ['regex' => $this->createRegularExpression($condition->getValue())]
+            : ['value' => $condition->getValue()];
+
+        $column = $condition->getColumn();
+        if (! array_key_exists($column, $this->jsonPaths)) {
```
nilmerg: Why `array_key_exists`? Do you want to accept `null` instead of an array? Please `isset(…) && is_array(…) && ! empty(…)`.

application/controllers/EventRuleController.php
BastianLedererIcinga: This might be a bit too much, the idea is that if some bug causes `version` to be an empty string, the message is still readable.

nilmerg: then quote the placeholder in the template :joy:

BastianLedererIcinga: True, I removed it

application/controllers/EventRuleController.php
```diff
@@ -190,13 +197,120 @@ public function completeAction(): void
...
+                $filter = $form->getFilter();
+
+                $this->session->set(
+                    'object_filter',
+                    (new RuleSerializer($filter, $hook->getJsonPaths(...$this->collectColumns($filter))))->getJson()
+                );
+                $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
+            });
+
+        $editor->getParser()->on(QueryString::ON_CONDITION, $hook->enrichCondition(...));
```
nilmerg: this may also throw, please guard it as well

nilmerg (changes requested):
please add a unit test for the source hook locator and the rule serializer.

application/controllers/EventRuleController.php
```diff
@@ -190,13 +197,120 @@ public function completeAction(): void
...
+
+        $type = $requestData['term']['type'] ?? null;
+        $label = $requestData['term']['label'] ?? '';
+
+        if ($type === 'column') {
+            $provider = $hook->getColumnSuggestions($label);
+        } else {
+            $column = $requestData['column'] ?? null;
+            if ($column === null || $column === SearchEditor::FAKE_COLUMN) {
+                $provider = new EmptyIterator();
```
nilmerg: Once https://github.com/Icinga/ipl-web/issues/381 is a thing, this should be handled like https://github.com/Icinga/ipl-web/blob/fde750c3a30a47370003a03687bf719edaaf8a22/src/Control/SearchBar/Suggestions.php#L348.

test/php/library/Notifications/Common/SourceHookLocatorTest.php
```diff
@@ -0,0 +1,32 @@
...
+use Icinga\Application\Icinga;
+use Icinga\Module\Notifications\Common\SourceHookLocator;
+use PHPUnit\Framework\TestCase;
+
+class SourceHookLocatorTest extends TestCase
+{
+    public function setUp(): void
+    {
+        Hook::register('Notifications\\V2\\TestSource', 'stub', TestSource::class, true);
+        Icinga::app()->getLoader()->registerNamespace(__NAMESPACE__, __DIR__);
```
nilmerg: This is only necessary because it's a separate file. I don't like the dependency on Icinga Web's autoloader here, as this is a unit test after all. The reason why it is required at all is because `TestSource::class` doesn't seem to load the class and is instead parsed at compile time to a simple string. (according to claude)

The better, albeit dirty, alternative is to declare the hook implementation in the same test's file. Just put it above the test class.

nilmerg (changes requested):

nilmerg (dismissed):
Fine for now. Let's see if we get https://github.com/Icinga/ipl-web/issues/381 through before merging.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -0,0 +1,148 @@
...
+            'version' => self::VERSION,
+            'qs'      => QueryString::render($this->filter),
+        ];
+        if ($this->filter instanceof Filter\Chain) {
+            $result['ast'] = $this->serializeChain($this->filter);
+        } else {
+            $result['ast'] = $this->serializeCondition($this->filter);
+        }
+
+        return Json::encode($result);
```
nilmerg, line 61: This should return null in case no filter has been configured. At the moment this writes `{"version":2,"qs":"","ast":{"op":"&","rules":[]}}` to the database and also causes the editor to not allow to *add* a filter, but always *adjust* the filter. (As it also always sees the superfluous data.)

application/controllers/EventRuleController.php
```diff
@@ -190,13 +196,133 @@ public function completeAction(): void
...
+                        ));
+                    }
+                }
+            )
+            ->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $hook) {
+                $filter = $form->getFilter();
+
+                $this->session->set(
+                    'object_filter',
+                    (new RuleSerializer($filter, $hook->getJsonPaths(...$this->collectColumns($filter))))->getJson()
```
nilmerg: This will benefit of https://github.com/Icinga/ipl-stdlib/pull/75, please use it

nilmerg (changes requested):

application/controllers/EventRuleController.php
nilmerg: I meant both additions ;)

```suggestion
                    (new RuleSerializer(
                        $filter,
                        $hook->getJsonPaths(...Seq::map($filter->yieldRules(), fn ($r) => $r->getColumn()))
                    ))->getJson()
```

BastianLedererIcinga: As discussed I added `Seq::unique` to your suggestion to deduplicate columns.

test/php/library/Notifications/Util/RuleSerializerTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Source hook did not provide a JSON path for column "absent"');
+
+        (new RuleSerializer($filter, []))->getJson();
+    }
+
+    public function testGetJsonReturnsNullForEmptyChain()
+    {
+        $result = json_decode((new RuleSerializer(Filter::all(), []))->getJson(), true);
```
nilmerg: Why json_decode? `getJson` returns `NULL` doesn't it?

nilmerg (approved):

Issue #463, opened 2026-04-10
Ability for source integrations to create, update and delete their Notifications configuration


nilmerg:
Tightly integrates sources will need a way to set up their own configuration for Icinga Notifications. This means they use a name of their choosing and a generated username (e.g. a UUID) to create it on behalf of the user. The password must be optional.

The username is the identifier for the configuration and sufficient to update or delete the configuration at a later time.

PR #471, opened 2026-05-07
Support configuration of event rules for generic sources

BastianLedererIcinga:
For event rules of  generic sources a simple `SearchEditor` without enrichment, validation and suggestions is created.

resolve #450 
require https://github.com/Icinga/ipl-web/pull/384
require https://github.com/Icinga/ipl-web/pull/358

### DIFF
--- a/application/controllers/EventRuleController.php
+++ b/application/controllers/EventRuleController.php
@@ -16,6 +16,7 @@ use Icinga\Module\Notifications\Common\SourceHookLocator;
 use Icinga\Module\Notifications\Forms\EventRuleConfigElements\NotificationConfigProvider;
 use Icinga\Module\Notifications\Forms\EventRuleConfigForm;
 use Icinga\Module\Notifications\Forms\EventRuleForm;
+use Icinga\Module\Notifications\Forms\SourceForm;
 use Icinga\Module\Notifications\Hook\V2\SourceHook;
 use Icinga\Module\Notifications\Model\Rule;
 use Icinga\Module\Notifications\Model\Source;
@@ -26,15 +27,19 @@ use Icinga\Web\Session;
 use ipl\Html\Attributes;
 use ipl\Html\Contract\Form;
 use ipl\Html\Html;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
 use ipl\Stdlib\Filter;
 use ipl\Stdlib\Filter\Condition;
 use ipl\Stdlib\Seq;
+use ipl\Web\Common\CalloutType;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Control\SearchBar\SearchException;
 use ipl\Web\Control\SearchEditor;
 use ipl\Web\Filter\QueryString;
 use ipl\Web\FormElement\SearchSuggestions;
 use ipl\Web\Url;
+use ipl\Web\Widget\Callout;
 use ipl\Web\Widget\Icon;
 use ipl\Web\Widget\Link;
 use JsonException;
@@ -233,14 +238,15 @@ class EventRuleController extends CompatController
 
         $editor = (new SearchEditor())
             ->setQueryString($parsedFilter['qs'] ?? '')
-            ->setSuggestionUrl(
+            ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl());
+
+        if ($hook !== null) {
+            $editor->setSuggestionUrl(
                 Url::fromPath(
                     'notifications/event-rule/suggest',
                     ['id' => $ruleId, '_disableLayout' => true, 'showCompact' => true]
                 )
-            )
-            ->setAction(Url::fromRequest()->with('object_filter', $filter)->getAbsoluteUrl())
-            ->on(
+            )->on(
                 SearchEditor::ON_VALIDATE_COLUMN,
                 function (Condition $condition) use ($hook) {
                     try {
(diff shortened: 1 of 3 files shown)


public/css/form.less
```diff
@@ -78,6 +78,10 @@
   }
 }
 
+.search-editor .hint{
```
sukhwinder33445: Please add the missing space after `.hint`

application/controllers/EventRuleController.php
```diff
@@ -351,7 +381,7 @@ protected function resolveSourceHook(int $ruleId): SourceHook
 
         $hook = SourceHookLocator::forType($source->type);
 
-        if ($hook === null) {
+        if ($hook === null && $source->type !== 'generic') {
```
sukhwinder33445: You can use `SourceForm::TYPE_GENERIC` instead, and we can move it to `Model/Source` later when implementing the integration for [Source](https://github.com/Icinga/icinga-notifications-web/issues/463).

application/controllers/EventRuleController.php
```diff
@@ -259,32 +262,59 @@ function (Condition $condition) use ($hook) {
...
+            );
+            $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
+        })->handleRequest($this->getServerRequest());
+
+        if ($hook === null) {
+            $editor->addHtml(
+                new HtmlElement(
+                    'div',
+                    Attributes::create(['class' => 'hint']),
+                    Text::create($this->translate('Columns must be valid JsonPath, no validation is available.'))
```
sukhwinder33445: I would suggest providing a bit more detail in the hint, as validation is not available for this source specifically and the user must ensure that the value is a valid `JSONPath`. 
For example:` Please make sure columns are valid JsonPath expressions, as no validation is available for this source.`

application/controllers/EventRuleController.php
```diff
@@ -259,32 +262,59 @@ function (Condition $condition) use ($hook) {
...
+                    $getJsonPaths($filter)
+                ))->getJson()
+            );
+            $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
+        })->handleRequest($this->getServerRequest());
+
+        if ($hook === null) {
+            $editor->addHtml(
+                new HtmlElement(
+                    'div',
```
sukhwinder33445: As this is only a hint, please use `p` instead of `div`.

sukhwinder33445 (dismissed):

application/controllers/EventRuleController.php
nilmerg: I'd rather suggest to use https://github.com/Icinga/ipl-web/pull/358 and add it to content prior the editor.

application/controllers/EventRuleController.php
```diff
@@ -259,32 +262,59 @@ function (Condition $condition) use ($hook) {
...
+
+        $editor->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $getJsonPaths) {
+            $filter = $form->getFilter();
+            $this->session->set(
+                'object_filter',
+                (new RuleSerializer(
+                    $filter,
+                    $getJsonPaths($filter)
+                ))->getJson()
+            );
```
nilmerg: Just a minor thing, but can't this be a single line now?

application/controllers/EventRuleController.php
nilmerg: True, but also let's point to the fact that the source must be able to understand these as well and must provide some form of documentation as it does not provide an integration after all.

application/controllers/EventRuleController.php
```diff
@@ -351,7 +381,7 @@ protected function resolveSourceHook(int $ruleId): SourceHook
 
         $hook = SourceHookLocator::forType($source->type);
```
nilmerg, line 386: I don't want to *accidentally* support `GenericSource` hooks. Return earlier with null please.

application/forms/SourceForm.php
```diff
@@ -53,8 +53,9 @@ protected function assemble(): void
             Text::create($this->translate(
                 'Sources are the most vital part of Icinga Notifications. They submit events that will be'
                 . ' processed to notify users about incidents. You can either configure sources that provide an'
-                . ' integration in Icinga Web or use the Generic type for sources that communicate directly with the'
-                . ' Icinga Notifications API. Refer to the source\'s documentation for the correct source type.'
+                . ' integration in Icinga Web or enter "generic" as the type for sources that communicate directly'
+                . ' with the Icinga Notifications API. Refer to the source\'s documentation for the correct source'
+                . ' type.'
```
nilmerg: > Refer to the source's documentation for the correct source type.

This misses a relation to the non-generic part earlier on. Right now one might think a non-integrated source also documents its type.

nilmerg (changes requested):

BastianLedererIcinga:
The failing tests are unrelated to this PR, they are caused by the API tests being incompatible with recent changes to the schema.

application/controllers/EventRuleController.php
```diff
@@ -259,32 +265,55 @@ function (Condition $condition) use ($hook) {
...
+                    CalloutType::Info,
+                    Text::create(
+                        $this->translate(
+                            'Please make sure columns are valid JSON paths, '
+                            . 'as no validation is available for this source. '
+                            . 'Refer to the source\'s documentation for available columns.'
+                        )
+                    )
+                )
+            );
```
nilmerg, line 316: This needs some margin.

Image

application/forms/SourceForm.php
nilmerg, line 1: I think we should change this to something more self-explainable. Right now the description what to enter as type is somewhat convoluted and not easy to grasp. I think of a radio element for example, the first and default option is *Generic* and the second is *Integrated*. Choosing the second reveals the text input we have right now, but it has a description under it that explains what to enter. It should also hint at the fact that integrated sources *usually* have their own configuration UI for notifications and it is recommended to use this instead.

nilmerg (changes requested):

application/forms/SourceForm.php
```diff
@@ -68,14 +70,44 @@ protected function assemble(): void
...
-            'text',
-            'type',
+            'select',
+            'category',
             [
-                'required'          => true,
-                'label'             => $this->translate('Source Type'),
+                'ignore'    => true,
+                'required'  => true,
+                'label'     => $this->translate('Source Category'),
```
nilmerg: The hint at the top talks about *source types*, please streamline this.

application/forms/SourceForm.php
```diff
@@ -68,14 +70,44 @@ protected function assemble(): void
...
+                'value'     => self::TYPE_GENERIC,
+                'class'     => 'autosubmit',
+                'options'   => [
+                    self::TYPE_GENERIC    => $this->translate('Generic'),
+                    self::TYPE_INTEGRATED => $this->translate('Integrated')
```
nilmerg: these translations should have a scope, to differentiate them from other usages of the same words. The scope can be a technical notation such as `'notifications.source.type'`.

application/forms/SourceForm.php
```diff
@@ -32,6 +32,9 @@ class SourceForm extends CompatForm
     /** @var string @var The generic source type */
     public const TYPE_GENERIC = 'generic';
 
+    /** @var string The type for sources with an integration */
+    public const TYPE_INTEGRATED = 'integrated';
```
nilmerg: If you insist on using a constant, please mark it private.

application/forms/SourceForm.php
```diff
@@ -68,14 +70,44 @@ protected function assemble(): void
...
+                'value'     => self::TYPE_GENERIC,
+                'class'     => 'autosubmit',
+                'options'   => [
+                    self::TYPE_GENERIC    => $this->translate('Generic'),
+                    self::TYPE_INTEGRATED => $this->translate('Integrated')
+                ]
             ]
         );
 
+        if ($this->getValue('category') === self::TYPE_INTEGRATED) {
```
nilmerg: We usually use `getPopulatedValue` instead in case of autosubmits that change the form structure.

application/forms/SourceForm.php
```diff
@@ -68,14 +70,44 @@ protected function assemble(): void
...
+                            . ' notifications, which is the recommended way to set them up.'
+                        )
+                    )
+                ));
+            $this->addElement(
+                'text',
+                'type',
+                [
+                    'required'      => true,
+                    'label'         => $this->translate('Source Type'),
```
nilmerg: Since this now collides with my other comment, please change this to *Source Identifier*.

application/controllers/EventRuleController.php
```diff
@@ -311,7 +311,9 @@ function (Condition $condition) use ($hook) {
                             . 'Refer to the source\'s documentation for available columns.'
                         )
                     )
-                )
+                ))
+                    ->setFitContent()
```
nilmerg: I just wondered why you need this… The callout will soon not require 100% width by default so this is not required anymore.

nilmerg (changes requested):

application/forms/SourceForm.php
```diff
@@ -67,14 +70,45 @@ protected function assemble(): void
...
-                'required'          => true,
-                'label'             => $this->translate('Source Type'),
+                'ignore'    => true,
+                'required'  => true,
+                'label'     => $this->translate('Source Type'),
+                'value'     => self::TYPE_GENERIC,
+                'class'     => 'autosubmit',
+                'options'   => [
+                    self::TYPE_GENERIC    => $this->translateInDomain('notifications.source.type', 'Generic'),
+                    self::TYPE_INTEGRATED => $this->translateInDomain('notifications.source.type', 'Integrated')
```
nilmerg: Jeesh, I've meant *context*, not *scope*, in my previous comment. There's a fitting parameter in `translate()` so I guess this mistake is my responsibility…

Domain's are quite a different thing and have the same effect, but on a broader scope. There's an *icinga* domain, for example. ;)

nilmerg (changes requested):

application/forms/SourceForm.php
```diff
@@ -67,14 +70,45 @@ protected function assemble(): void
...
         );
 
+        if ($this->getPopulatedValue('source_type') === self::TYPE_INTEGRATED) {
+            $this->addHtml(
+                new HtmlElement(
+                    'p',
+                    Attributes::create(['class' => 'description']),
+                    Text::create(
+                        $this->translate(
+                            'Enter the source type as stated in the integration\'s documentation.'
```
nilmerg: *source* **identifier**

nilmerg (changes requested):

public/css/form.less
```diff
@@ -78,6 +78,10 @@
   }
 }
 
+.generic-source-hint {
+  margin: 1em;
```
nilmerg:
```suggestion
  margin: 1em 1em 0 1em;
```

nilmerg (changes requested):

application/forms/SourceForm.php
nilmerg, line 123: The entire section is redundant now as the input to set the identifier is not auto-submitting the form and hence the hint never changes or at least not as initially desired. But since we will move the configuration to the sources now and recommend it, this won't be required anyway any longer.

nilmerg: i.e. the translated text can already include the *generic* hint to *consult the documentation*.

nilmerg (changes requested):

nilmerg (approved):

PR #472, opened 2026-05-11
Allow integrations to interact with Incidents

### LINKED ISSUE #220: Model `IncidentContact` is missing columns `contactgroup_id` and `schedule_id`

### LINKED ISSUE #464: Ability for integrations to interact with Incidents and fetch details about them
Integrations may want to integrate certain features of Icinga Notifications Web in their own UI. Such as:

* Manage/Unmanage
* Subscribe/Unsubscribe
* Who has been notified?
* Who is manager or subscriber?
* Is the incident muted?
* Is there an incident?

Since sources identify their objects based on tags, it's probably best to provide a way to access incidents based on them.

A quick idea floating around might be an `Incidents` object that expects an array of tags and yields objects that allow the operations mentioned above.

BastianLedererIcinga:
resolve #464
resolve #220 

Add the class `Incidents` which finds incidents that match a given array of tags independant of the source.

Each yielded `Incident` allows basic read/write operations.

The tests are written by claude.

### DIFF
--- a/library/Notifications/Integrations/Incident.php
+++ b/library/Notifications/Integrations/Incident.php
@@ -1,0 +1,240 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Notifications\Integrations;
+
+use DateTime;
+use Exception;
+use Generator;
+use Icinga\Module\Notifications\Common\Database;
+use Icinga\Module\Notifications\Model\Contact;
+use Icinga\Module\Notifications\Model\Incident as IncidentModel;
+use Icinga\Module\Notifications\Model\IncidentContact;
+use ipl\Stdlib\Filter;
+
+class Incident
+{
+    protected IncidentModel $incident;
+
+    public function __construct(IncidentModel $incident)
+    {
+        $this->incident = $incident;
+    }
+
+    /**
+     * Add the contact as manager of the incident
+     *
+     * @param Contact $contact
+     *
+     * @return $this
+     */
+    public function addManager(Contact $contact): static
+    {
+        $this->setRole($contact, 'manager');
+
+        return $this;
+    }
+
+    /**
+     * Change the contact's role from manager to subscriber
+     *
+     * This has no effect if the contact is not a manager of the incident
+     *
+     * @param Contact $contact
+     *
+     * @return $this
+     */
+    public function removeManager(Contact $contact): static
+    {
+        $existing = $this->fetchIncidentContact($contact);
+        if ($existing?->role !== 'manager') {
+            return $this;
+        }
+
+        $this->setRole($contact, 'subscriber', $existing);
+        return $this;
+    }
+
+    /**
+     * Add the contact as a subscriber
+     *
+     * This has no effect if the contact is already a subscriber or a manager
+     *
+     * @param Contact $contact
+     *
+     * @return $this
+     */
+    public function addSubscriber(Contact $contact): static
+    {
+        $existing = $this->fetchIncidentContact($contact);
+        if ($existing?->role === 'subscriber' || $existing?->role === 'manager') {
+            return $this;
+        }
+
+        $this->setRole($contact, 'subscriber', $existing);
+        return $this;
+    }
+
+    /**
+     * Remove the contact as a subscriber
+     *
+     * This will delete the incident_contact row. Has no effect if the contact is not a subscriber.
+     *
+     * @param Contact $contact
+     *
+     * @return $this
+     */
+    public function removeSubscriber(Contact $contact): static
+    {
+        $db = Database::get();
+        $existing = $this->fetchIncidentContact($contact);
+        if ($existing === null || $existing->role !== 'subscriber') {
+            return $this;
+        }
+
+        $db->beginTransaction();
+        try {
+            $db->delete('incident_contact', [
+                'incident_id = ?' => $this->incident->id,
+                'contact_id = ?'  => $contact->id,
+                'role = ?'        => 'subscriber'
+            ]);
+
+            $this->insertHistory($contact->id, 'subscriber', null);
+        } catch (Exception $e) {
+            $db->rollBackTransaction();
+            throw $e;
+        }
+
+        $db->commitTransaction();
+        return $this;
+    }
+
+    /**
+     * Get all managers of the incident
+     *
+     * @return Generator<int, IncidentContact>
+     */
+    public function getManagers(): Generator
+    {
+        yield from $this->fetchContactsByRole('manager');
+    }
+
+    /**
+     * Get all subscribers of the incident
+     *
+     * @return Generator<int, IncidentContact>
+     */
+    public function getSubscribers(): Generator
+    {
+        yield from $this->fetchContactsByRole('subscriber');
+    }
+
+    /**
+     * Get all contacts that were notified about this incident
+     *
+     * @return Generator<int, Contact>
+     */
+    public function getNotifiedContacts(): Generator
+    {
+        $query = Contact::on(Database::get())
+            ->filter(Filter::all(
+                Filter::equal('incident_history.incident_id', $this->incident->id),
+                Filter::equal('incident_history.type', 'notified')
+            ));
+        $query->getSelectBase()->distinct();
+
+        yield from $query;
+    }
+
+    public function isMuted(): bool
+    {
+        return $this->incident->mute_reason !== null;
+    }
+
+    protected function setRole(Contact $contact, string $role, ?IncidentContact $existing = null): void
+    {
+        $db = Database::get();
+        if ($existing === null) {
+            $existing = $this->fetchIncidentContact($contact);
+        }
+
+        $oldRole = $existing?->role;
+
+        if ($oldRole === $role) {
+            return;
+        }
+
+        $db->beginTransaction();
+        try {
+            if ($existing !== null) {
+                $db->update(
+                    'incident_contact',
+                    ['role' => $role],
+                    [
+                        'incident_id = ?' => $this->incident->id,
+                        'contact_id = ?'  => $contact->id
+                    ]
+                );
+            } else {
+                $db->insert('incident_contact', [
+                    'incident_id' => $this->incident->id,
+                    'contact_id'  => $contact->id,
+                    'role'        => $role
+                ]);
+            }
+
+            $this->insertHistory($contact->id, $oldRole, $role);
+        } catch (Exception $e) {
+            $db->rollBackTransaction();
+            throw $e;
+        }
+
+        $db->commitTransaction();
+    }
+
+    protected function fetchIncidentContact(Contact $contact): ?IncidentContact
+    {
+        /** @var ?IncidentContact $entry */
+        $entry = IncidentContact::on(Database::get())
+            ->filter(Filter::all(
+                Filter::equal('incident_id', $this->incident->id),
+                Filter::equal('contact_id', $contact->id)
+            ))
+            ->first();
+
+        return $entry;
+    }
+
+    /**
+     * Fetch all contacts that have the given role
+     *
+     * @return Generator<int, IncidentContact>
+     */
+    protected function fetchContactsByRole(string $role): Generator
+    {
+        yield from IncidentContact::on(Database::get())
+            ->filter(Filter::all(
+                Filter::equal('incident_id', $this->incident->id),
+                Filter::equal('role', $role)
+            ));
+    }
+
+    protected function insertHistory(int $contactId, ?string $oldRole, ?string $newRole): void
+    {
+        $now = new DateTime();
+        Database::get()->insert(
+            'incident_history',
+            [
+                'incident_id'        => $this->incident->id,
+                'contact_id'         => $contactId,
+                'type'               => 'recipient_role_changed',
+                'new_recipient_role' => $newRole,
+                'old_recipient_role' => $oldRole,
+                'time'               => (int) $now->format('Uv')
+            ]
+        );
+    }
+}
--- a/library/Notifications/Integrations/Incident.php
+++ b/library/Notifications/Integrations/Incident.php
@@ -1,0 +1,333 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Notifications\Integrations;
+
+use DateTime;
+use Generator;
+use Icinga\Module\Notifications\Common\EntityManager;
+use Icinga\Module\Notifications\Model\Contact;
+use Icinga\Module\Notifications\Model\Incident as IncidentModel;
+use Icinga\Module\Notifications\Model\IncidentContact;
+use Icinga\Module\Notifications\Model\IncidentHistory;
+use InvalidArgumentException;
+use ipl\Sql\Connection;
+use ipl\Stdlib\Filter;
+
+/**
+ * Manage an incident's recipients and read its state
+ */
+class Incident
+{
+    /** @var IncidentModel The managed incident */
+    private IncidentModel $incident;
+
+    /** @var Connection The database connection to use */
+    private Connection $db;
+
+    /**
+     * Create a new wrapper for the given model
+     *
+     * @param IncidentModel $incident
+     * @param Connection $db The connection to read and persist through
+     */
+    public function __construct(IncidentModel $incident, Connection $db)
+    {
+        $this->incident = $incident;
+        $this->db = $db;
+    }
+
+    /**
+     * Add the contact with the given username as manager
+     *
+     * Has no effect if the contact is already a manager.
+     *
+     * @param string $username
+     *
+     * @return $this
+     *
+     * @throws InvalidArgumentException If no contact with that username exists
+     */
+    public function addManager(string $username): static
+    {
+        $this->assignRole($username, 'manager', ['manager']);
+
+        return $this;
+    }
+
+    /**
+     * Add the contact with the given username as subscriber
+     *
+     * Has no effect if the contact is already a subscriber or a manager.
+     *
+     * @param string $username
+     *
+     * @return $this
+     *
+     * @throws InvalidArgumentException If no contact with that username exists
+     */
+    public function addSubscriber(string $username): static
+    {
+        $this->assignRole($username, 'subscriber', ['subscriber', 'manager']);
+
+        return $this;
+    }
+
+    /**
+     * Demote the manager with the given username to subscriber
+     *
+     * Has no effect if the contact is not a manager of the incident.
+     *
+     * @param string $username
+     *
+     * @return $this
+     *
+     * @throws InvalidArgumentException If no contact with that username exists
+     */
+    public function removeManager(string $username): static
+    {
+        $this->assignRole($username, 'subscriber', [null, 'recipient', 'subscriber']);
+
+        return $this;
+    }
+
+    /**
+     * Remove the subscriber with the given username
+     *
+     * Has no effect if the contact is not a subscriber.
+     *
+     * @param string $username
+     *
+     * @return $this
+     *
+     * @throws InvalidArgumentException If no contact with that username exists
+     */
+    public function removeSubscriber(string $username): static
+    {
+        $contact = $this->getContactByName($username);
+        $existing = $this->existingContact($contact->id);
+
+        if ($existing?->role !== 'subscriber') {
+            return $this;
+        }
+
+        (new EntityManager($this->db))->save($existing->delete());
+        $this->addRoleChangedHistory($contact->id, 'subscriber', null);
+
+        return $this;
+    }
+
+    /**
+     * Yield each active subscriber of the incident
+     *
+     * @return Generator<int, array{
+     *     name: string,
+     *     full_name: string,
+     *     role: 'manager'|'subscriber',
+     *     roleChangedAt: DateTime}>
+     */
+    public function getSubscribers(): Generator
+    {
+        foreach ($this->resolveRecipients(['manager', 'subscriber']) as $recipient) {
+            yield [
+                'name'          => $recipient['name'],
+                'full_name'     => $recipient['full_name'],
+                'role'          => $recipient['role'],
+                'roleChangedAt' => $recipient['roleChangedAt'],
+            ];
+        }
+    }
+
+    /**
+     * Yield each configured recipient of the incident
+     *
+     * @return Generator<int, array{
+     *     type: 'contact'|'contactgroup'|'schedule',
+     *     name: string,
+     *     full_name: ?string,
+     *     roleChangedAt: DateTime}>
+     */
+    public function getRecipients(): Generator
+    {
+        foreach ($this->resolveRecipients(['recipient']) as $recipient) {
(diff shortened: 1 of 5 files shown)


library/Notifications/Widget/Detail/IncidentDetail.php
```diff
@@ -189,7 +189,6 @@ protected function assemble(): void
             $this->createHistory(),
             $this->createRelatedObject(),
             $this->createSource(),
-            $this->createObjectTag(),
```
sukhwinder33445: This is already removed, please rebase.

nilmerg:
please rebase

nilmerg (changes requested):
I have a rather fundamental change request, quite similar to what I already told @sukhwinder33445 regarding #483.

The `Incidents` traversable yields `Model\Incident` right now and expects other models such as `Contact` as part of its interface. This poses a problem regarding the future compatibility with integrations for which this implementation is for.

An integration should not have to access internal abstractions such as the database directly, which the models actually represent. So this is right now only half the story, as you've introduced a stable abstraction to fetch incidents, but not one to interact with them.

The first problem is already visible when looking at how a manager is being added: The method requires a `Model\Contact`. How does the integration get access to such? The better alternative is to require a `string $username`, since that's how the UI identifies users and what our own UI integration uses to associate UI users with contacts.

So what I'm thinking of is an interface that demands only as less information as possible and one that is readily available for a source integration. This then should be represented by a new object, such as `Integrations\Incident`.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,354 @@
...
+    {
+        $history = new IncidentHistory();
+        $history->contact_id = $contact->id;
+        $history->type = 'recipient_role_changed';
+        $history->old_recipient_role = $oldRole;
+        $history->new_recipient_role = $newRole;
+        $history->time = new DateTime();
+
+        $this->pendingHistory[] = $history;
+        $this->incident->incident_history = $this->pendingHistory;
```
sukhwinder33445: The `pendingHistory` property is only used in this method, so it can be omitted. Why not read from and write to `incident_history` directly instead?

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,354 @@
...
+     * @return void
+     */
+    protected function addRoleChangedHistory(Contact $contact, ?string $oldRole, ?string $newRole): void
+    {
+        $history = new IncidentHistory();
+        $history->contact_id = $contact->id;
+        $history->type = 'recipient_role_changed';
+        $history->old_recipient_role = $oldRole;
+        $history->new_recipient_role = $newRole;
+        $history->time = new DateTime();
```
sukhwinder33445: You can pass these directly to the constructor as parameters.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,354 @@
...
+     * @param Contact $contact
+     * @param ?string $oldRole
+     * @param ?string $newRole
+     *
+     * @return void
+     */
+    protected function addRoleChangedHistory(Contact $contact, ?string $oldRole, ?string $newRole): void
+    {
+        $history = new IncidentHistory();
+        $history->contact_id = $contact->id;
```
sukhwinder33445: As this method only requires the `contact_id`, I would suggest passing only that as the first parameter.

BastianLedererIcinga:
As discussed with @nilmerg I changed `Incidents`, to no longer require the `source_id`, added a factory method that doesn't need the `Connection` as parameter and modified the getters of `Incident` to return both `username` and `full_name`.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,356 @@
...
+
+        $oldRole = $existing?->role;
+
+        if ($existing !== null) {
+            $existing->role = $role;
+        } else {
+            $entry = new IncidentContact();
+            $entry->contact_id = $contact->id;
+            $entry->role = $role;
+            $contacts[] = $entry;
```
sukhwinder33445: You can pass these directly to the constructor as parameters:
```suggestion
$contacts[] = new IncidentContact([
    'contact_id' => $contact->id,
    'role'       => $role
]);
```

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,356 @@
...
+
+        if ($contact === null) {
+            throw new InvalidArgumentException(sprintf('There is no contact with the username "%s"', $username));
+        }
+
+        return $contact;
+    }
+
+    /**
+     * Find the entry for the given contact among the supplied incident_contact entries
```
sukhwinder33445: Please use backticks for table name.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,356 @@
...
+     *
+     * Has no effect if the contact is already a subscriber or a manager.
+     *
+     * @param string $username
+     *
+     * @return $this
+     */
+    public function addSubscriber(string $username): static
+    {
+        $this->assignRole($this->getContactByName($username), 'subscriber', ['subscriber', 'manager']);
```
sukhwinder33445: I would suggest to let `assignRole` retrieve the contacts internally and simply pass the username here.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Find the entry for the given contact among the supplied incident_contact entries
+     *
+     * @param list<IncidentContact> $contacts
+     * @param Contact $contact
+     *
+     * @return ?IncidentContact
+     */
+    private function findIncidentContact(array $contacts, Contact $contact): ?IncidentContact
```
sukhwinder33445: Passing `contactId` as secound param is sufficient here.

sukhwinder33445 (changes requested):

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Append a `recipient_role_changed` entry to the incident's history relation
+     *
+     * @param Contact $contact
+     * @param ?string $oldRole
+     * @param ?string $newRole
+     *
+     * @return void
+     */
+    protected function addRoleChangedHistory(Contact $contact, ?string $oldRole, ?string $newRole): void
```
sukhwinder33445: This can be private as well.

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /** @var ?ResultSet The executed query result */
+    protected ?ResultSet $results = null;
+
+    /**
+     * Create new Incidents
+     *
+     * Matches the incidents of every object that carries each tag given with a value and lacks each tag
+     * given as null. Tags not listed are unconstrained. Matching is performed by tag alone, independent of the source.
+     *
```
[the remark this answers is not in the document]
BastianLedererIcinga: Requiring a `source_id` was a mistake, I adjusted the PR description to reflect the new implementation.

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: As discussed in person, reading and writing directly via `incident_history` makes things more complicated as they have to be, because it may be a `Closure` or an array depending on whether it has already been loaded or not.
I also don't see a benefit in loading the entire existing history from the db if we don't actually need it.

I did however move the assignment of `incident_history` to `Incident::save()` to clean things up, and made sure to reset `pendingHistory` when saving.

BastianLedererIcinga:
@sukhwinder33445 You pointed out to me in person that caching `Incidents::results` may be fragile.
I checked on it, and `hasIncident` did return false if the `Generator` of the `ResultSet` had already been exhausted.
This order of calls is unlikely to happen in practice, but I adjusted `hasIncident` so it doesn't depend on the generator's internal state.

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->tags = $tags;
+        $this->db = $db;
+    }
+
+    /**
+     * Create new Incidents for the given tags, reading through the default database connection
+     *
+     * @param array<string, ?string> $tags Required tags keyed by name, a null value requires the tag's absence
+     *
+     * @return self
```
sukhwinder33445: We always use `@return $this`.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,359 @@
...
+     *
+     * @param int $contactId
+     * @param ?string $oldRole
+     * @param ?string $newRole
+     *
+     * @return void
+     */
+    protected function addRoleChangedHistory(int $contactId, ?string $oldRole, ?string $newRole): void
+    {
+        $history = new IncidentHistory(
```
sukhwinder33445: Add value directly to `$this->pendingHistory[]`;

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,359 @@
...
+        }
+
+        $this->incident->incident_contact = $contacts;
+        $this->addRoleChangedHistory($contact->id, $oldRole, $role);
+
+        return $this;
+    }
+
+    /**
+     * Append a recipient_role_changed entry to the pending history, to be persisted on the next save()
```
sukhwinder33445: Add backticks to the table name please.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,359 @@
...
+        if ($existing !== null && in_array($existing->role, $noopRoles, true)) {
+            return $this;
+        }
+
+        $oldRole = $existing?->role;
+
+        if ($existing !== null) {
+            $existing->role = $role;
+        } else {
+            $entry = new IncidentContact(['contact_id' => $contact->id, 'role' => $role]);
```
sukhwinder33445: Var `$entry` can be ommited.

sukhwinder33445 (dismissed):

library/Notifications/Integrations/Incidents.php
BastianLedererIcinga: This is a static function.

nilmerg: Then `static` unless the class is `final` 😛

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public static function create(array $tags): self
```
nilmerg: hmm, let's see how a usage might look like:

```php
foreach (Incidents::create(["host" => "icinga2"]) as $incident) {
}
```

What about this instead?

```php
foreach (Incidents::find(["host" => "icinga2"]) as $incident) {
}
```

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Get whether the object has at least one incident
+     *
+     * @return bool
+     */
+    public function hasIncident(): bool
+    {
+        // hasResult() fails if the generator has already been exhausted
```
nilmerg: This begs the question why anyone would call it after this. This isn't new behavior and I never heard or seen someone complaining about it. It's always:

1. Perform query
2. Call `hasResult()`
3. Iterate

Why would it be interesting to know if the iteration did something if `hasResult()` returns `true`?

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+use ipl\Orm\Query;
+use ipl\Orm\ResultSet;
+use ipl\Sql\Connection;
+use ipl\Sql\Filter\In;
+use ipl\Sql\Filter\NotIn;
+use ipl\Sql\Select;
+use ipl\Stdlib\Seq;
+use IteratorAggregate;
+
+class Incidents implements IteratorAggregate
```
nilmerg: Please make it `Countable` as well.

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+{
+    /** @var array<string, ?string> Required tags keyed by name; a null value requires the tag's absence */
+    protected array $tags;
+
+    /** @var Connection The database connection to read from */
+    protected Connection $db;
+
+    /** @var ?ResultSet The executed query result */
+    protected ?ResultSet $results = null;
```
nilmerg: Mark this private

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+     * Yield an interaction wrapper for each of the object's incidents
+     *
+     * @return Generator<Incident>
+     */
+    public function getIterator(): Generator
+    {
+        yield from Seq::map($this->incidents(), fn($incident) => new Incident($incident, $this->db));
+    }
+
+    protected function incidents(): ResultSet
```
nilmerg: This and `buildQuery` should also be private

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+                ->columns('object_id')
+                ->from('object_id_tag')
+                ->where('tag = ?', $tag);
+
+            if ($value === null) {
+                $query->filter(new NotIn('incident.object_id', $objectIds));
+            } else {
+                $objectIds->where('value = ?', $value);
+                $query->filter(new In('incident.object_id', $objectIds));
+            }
```
nilmerg: Can't this be based on how we usually build such queries?

```suggestion
            if ($value === null) {
                $query->filter(
                    Filter::unlike("incident.object.tag.$tag", '*')
                );
            } else {
                $query->filter(
                    Filter::equal("incident.object.tag.$tag", $value)
                );
            }
```

library/Notifications/Integrations/Incidents.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Yield an interaction wrapper for each of the object's incidents
+     *
+     * @return Generator<Incident>
+     */
+    public function getIterator(): Generator
+    {
+        yield from Seq::map($this->incidents(), fn($incident) => new Incident($incident, $this->db));
```
nilmerg: Is there an advantage to a simple foreach besides being a one-liner?

library/Notifications/Integrations/Incident.php
sukhwinder33445: Please revert this change and set the properties one by one as before — this helps with static analysis. @nilmerg  pointed this out yesterday on my PR.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,367 @@
...
+     *
+     * @return $this
+     */
+    public function removeSubscriber(string $username): static
+    {
+        $contact = $this->getContactByName($username);
+        $contacts = $this->incidentContacts();
+        $existing = $this->findIncidentContact($contacts, $contact->id);
+
+        if ($existing === null || $existing->role !== 'subscriber') {
```
nilmerg: Why not `$existing?->role !== …` as well?

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,367 @@
...
+    {
+        return $this->incident->mute_reason !== null;
+    }
+
+    /**
+     * Persist the pending changes to the database
+     *
+     * @return $this
+     */
+    public function save(): static
```
nilmerg: I don't quite see the necessity to *save* a just changed contact's role. And that's the only kind of change possible. To me it would be better to just update the database immediately.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,367 @@
...
+
+    /**
+     * Yield the username and full name of each of the subscribers
+     *
+     * @return Generator<int, array{username: string, full_name: string}>
+     */
+    public function getSubscribers(): Generator
+    {
+        yield from $this->contactsWithRole('subscriber');
+    }
```
nilmerg: Given what Florian just noted in https://github.com/Icinga/icingadb-web/issues/1358#issuecomment-4682061584 this should be a single method

library/Notifications/Integrations/Incident.php
nilmerg, line 1: The way the database persist is implemented is rather complex to me, let's discuss in person how this can be simplified.

nilmerg (changes requested):

library/Notifications/Integrations/Incidents.php
BastianLedererIcinga: I just tried this out.
`['host' => 'icinga2', 'service' => null]`
Would result in a query that requires a `service` tag to be present and have its value equal `null`.

nilmerg: Well, this is happens in case the same functionality is re-used but updates to the base are ignored: https://github.com/Icinga/icingadb-web/issues/865

I've moved over the same fix to our implementation here and oh wonder, it works: https://github.com/Icinga/icinga-notifications-web/pull/497

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: As discussed I adjusted the write functions to save immediately.

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: True, looking at the mockup I also realized that schedules and contactgroups were not yet supported by my code.

library/Notifications/Integrations/Incidents.php
BastianLedererIcinga: `buildQuery` is now extended in a test case so I'd keep it protected.

test/php/library/Notifications/Integrations/IncidentsTest.php
```diff
@@ -0,0 +1,238 @@
...
+     * @param array<string, ?string> $tags
+     */
+    private function builtFilter(array $tags): Chain
+    {
+        $incidents = new class ($tags, $this->db) extends Incidents {
+            public function exposeBuiltFilter(): Chain
+            {
+                return $this->buildQuery()->getFilter();
+            }
+        };
```
nilmerg: This can also be done via reflection, at runtime. I'd prefer this as now the class `Incidents` has two very closely related methods with different visibility which isn't best practice as well. So please decide whether both methods are protected or private + reflection. I can't argue for one or another objectively, though.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,349 @@
...
+
+        if ($existing?->role !== 'manager') {
+            return $this;
+        }
+
+        $existing->role = 'subscriber';
+        (new EntityManager($this->db))->save($existing);
+        $this->addRoleChangedHistory($contact->id, 'manager', 'subscriber');
+
+        return $this;
```
nilmerg: Why is this not using `assignRole`?

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+     * Yield each active subscriber of the incident
+     *
+     * An active subscriber is a recipient whose role is either `manager` or `subscriber`.
+     *
+     * @return Generator<int, array{
+     *     type: 'contact'|'contactgroup'|'schedule',
```
nilmerg: A contact group or schedule cannot push *Subscribe* and hence is never an *active* subscriber 🤣

nilmerg: Which is also the reason why `full_name` cannot be null.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,349 @@
...
+            ];
+        }
+    }
+
+    /**
+     * Yield each configured recipient of the incident
+     *
+     * @return Generator<int, array{type: 'contact'|'contactgroup'|'schedule', name: string, full_name: ?string}>
+     */
+    public function getRecipients(): Generator
```
nilmerg, line 152: I also seen the mockups, but I believe it makes sense to also provide the role change date here. Escalations are triggered at different stages of an incident and thus at different times after all.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,349 @@
...
+        $entries = IncidentContact::on($this->db)
+            ->with(['contact', 'contactgroup', 'schedule'])
+            ->filter(Filter::all(
+                Filter::equal('incident_id', $this->incident->id),
+                Filter::equal('role', $roles)
+            ));
+
+        $recipients = [];
+        foreach ($entries as $entry) {
+            if ($entry->contact_id !== null && ! $entry->contact->deleted) {
```
nilmerg: Isn't it possible to filter deleted objects already during the query?

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,349 @@
...
+        $entry = IncidentHistory::on($this->db)
+            ->columns(['time'])
+            ->filter(Filter::all(
+                Filter::equal('incident_id', $this->incident->id),
+                Filter::equal('type', 'recipient_role_changed'),
+                Filter::equal($recipient['type'] . '_id', $recipient['id']),
+                Filter::equal('new_recipient_role', $recipient['role'])
+            ))
+            ->orderBy('time', SORT_DESC)
+            ->first();
```
nilmerg: Aaaaah 🏃‍♂️ 

I think it's time for `incident_contact.changed_at` 😅

nilmerg (changes requested):

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: I am leaving this for now, until we have decided what we consider an _active subscriber_, and whether to distinguish them from recipients at all here.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,335 @@
...
+                    Filter::equal('contact.deleted', false)
+                ),
+                Filter::any(
+                    Filter::unlike('contactgroup_id', '*'),
+                    Filter::equal('contactgroup.deleted', false)
+                ),
+                Filter::any(
+                    Filter::unlike('schedule_id', '*'),
+                    Filter::equal('schedule.deleted', false)
+                )
```
nilmerg: There's a small optimization you might be unaware of:

Any negated condition is being transformed to essentially: *It either doesn't exist or differs from …*

You can drop the id filters and only use three `deleted!=true`.

sukhwinder33445 (commented):
The changes I requested have been made.

library/Notifications/Integrations/Incident.php
```diff
@@ -0,0 +1,335 @@
...
+    /**
+     * Load the contact with the given username
+     *
+     * @param string $username
+     *
+     * @return Contact
+     *
+     * @throws InvalidArgumentException If no contact with that username exists
+     */
+    private function getContactByName(string $username): Contact
```
nilmerg, line 181: every related public method should also document that this throws. It's a private method after all and my expectation of a proper external interface is that it describes it's full behavior.

nilmerg (changes requested):

nilmerg (approved):

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: As discussed subscribers will be expected to always be contacts, so I removed the `type` key and `full_name` is no longer nullable.

nilmerg (dismissed):

nilmerg (approved):

PR #482, opened 2026-05-28
Add write support for models

BastianLedererIcinga:
To support modifying models and then persisting them to the DB, an `EntityManager` is added, which can save and delete any given `Model` and its modified relations.

A new `Model` class is added that serves as parent class for all models in the module, it tracks whether it already exists in the DB, and what columns have been modified.

To manage members of to-many relations `Collection` is added, which exposes `attach()`, `detach()` and `sync()`, and also supports modifying its members in place.

### DIFF
--- a/library/Notifications/Common/Collection.php
+++ b/library/Notifications/Common/Collection.php
@@ -1,0 +1,318 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Notifications\Common;
+
+use ArrayIterator;
+use Countable;
+use ipl\Orm\Query;
+use IteratorAggregate;
+use LogicException;
+use Traversable;
+
+/**
+ * Container for members of a to-many relation
+ *
+ * For a HasMany relation, {@see self::attach()} and {@see self::detach()} insert/delete the database rows
+ * of the given models.
+ * For a BelongsToMany, only the links from the source model will be inserted/deleted, the target itself remains
+ * untouched.
+ *
+ * By default, rows that are not included in {@see self::$attach} or {@see self::$detach} remain untouched.
+ * Calling {@see self::sync()} will put the collection into replace mode, which will delete what is not contained
+ * in {@see self::$attach}
+ *
+ * Members may also be edited in place and will be cascaded by {@see EntityManager::save()} if they have
+ * been modified.
+ */
+class Collection implements IteratorAggregate, Countable
+{
+    /**
+     * Stored members, still as the unread {@see Query}
+     *
+     * Drained into {@see self::$base} once on first read, so the query is not run eagerly.
+     *
+     * @var ?Query
+     */
+    private ?Query $source = null;
+
+    /**
+     * Stored members once {@see self::$source} has been read, cached for repeated reads
+     *
+     * @var ?array<string, Model>
+     */
+    private ?array $base = null;
+
+    /** @var array<string, Model> Models whose links/rows must exist after the next {@see EntityManager::save()} */
+    private array $attach = [];
+
+    /** @var array<string, Model> Models whose links/rows must be removed on the next {@see EntityManager::save()} */
+    private array $detach = [];
+
+    /** @var bool Whether {@see self::sync()} requested a replace (remove whatever is not in {@see self::$attach}) */
+    private bool $replace = false;
+
+    /**
+     * Create a collection whose members are merged with existing ones on the next save
+     *
+     * @param iterable $values
+     *
+     * @return static
+     */
+    public static function create(iterable $values): static
+    {
+        $collection = new static();
+        foreach ($values as $model) {
+            $collection->attach($model);
+        }
+
+        return $collection;
+    }
+
+    /**
+     * Create a collection from members who are treated as already existing db rows
+     *
+     * @param Query $source
+     *
+     * @return static
+     */
+    public static function fromLoaded(Query $source): static
+    {
+        $collection = new static();
+        $collection->source = $source;
+
+        return $collection;
+    }
+
+    /**
+     * Get the {@see Query} the Collection was created from
+     *
+     * The query may be modified e.g. by calling {@see Query::filter()} on it, but only before the collection
+     * itself is iterated or {@see self::count()} is called.
+     *
+     * Modifications made to a Model yielded by the query will NOT be persisted by {@see EntityManager::save()}.
+     * Traverse the collection and modify models in place, or pass a modified {@see Model} to {@see self::attach()}
+     * if changes should be persisted.
+     *
+     * @return Query
+     *
+     * @throws LogicException If the collection was not created from a query, or the query has already been read
+     */
+    public function query(): Query
+    {
+        if ($this->source === null) {
+            throw new LogicException('Collection was created without a query, or the query was already read');
+        }
+
+        return $this->source;
+    }
+
+    /**
+     * Additively register the given model as a member of this relation
+     *
+     * @param Model $model
+     *
+     * @return $this
+     */
+    public function attach(Model $model): static
+    {
+        $id = $this->identify($model);
+        unset($this->detach[$id]);
+        if (! isset($this->base[$id])) {
+            $this->attach[$id] = $model;
+        } elseif ($model !== $this->base[$id]) {
+            throw new LogicException(
+                'Collection already contains a different model with the same primary key'
+            );
+        }
+
+        return $this;
+    }
+
+    /**
+     * Register the given model to be removed from this relation
+     *
+     * @param Model $model
+     *
+     * @return $this
+     */
+    public function detach(Model $model): static
+    {
+        $id = $this->identify($model);
+        unset($this->attach[$id]);
+        $this->detach[$id] = $model;
+
+        return $this;
+    }
+
+    /**
+     * Replace this relation's members with exactly the given models
+     *
+     * Anything currently stored that is not in $models is removed on save.
+     *
+     * @param iterable<Model> $models
+     *
+     * @return $this
+     */
+    public function sync(iterable $models): static
+    {
+        $this->attach = [];
+        $this->detach = [];
+        $this->replace = true;
+        foreach ($models as $model) {
+            $this->attach($model);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Get whether {@see self::sync()} put this collection into replace mode
+     *
+     * @return bool
+     */
+    public function isReplacing(): bool
+    {
+        return $this->replace;
+    }
+
+    /**
+     * Get the members to persist on save: staged attachments plus in-place-modified loaded members
+     *
+     * @return Model[]
+     */
+    public function getMembersToSave(): array
+    {
+        if ($this->replace) {
+            return array_values($this->attach);
+        }
+
+        return array_values(array_merge($this->base ?? [], $this->attach));
+    }
+
+    /**
+     * Get the models to delete/unlink on save
+     *
+     * @return Model[]
+     */
+    public function getDetachments(): array
+    {
+        return array_values($this->detach);
+    }
+
+    /**
+     * Whether anything must be persisted on the next save
+     *
+     * @return bool
+     */
+    public function hasPendingChanges(): bool
+    {
+        if ($this->replace || ! empty($this->attach) || ! empty($this->detach)) {
+            return true;
+        }
+
+        foreach ($this->base ?? [] as $model) {
+            if ($model->isModified()) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Reset the pending write state after the collection has been persisted
+     *
+     * @return void
+     */
+    public function clearPendingChanges(): void
+    {
+        $this->attach = [];
+        $this->detach = [];
+        $this->replace = false;
+    }
+
+    /**
+     * @return Traversable<Model>
+     */
+    public function getIterator(): Traversable
+    {
+        return new ArrayIterator($this->all());
+    }
+
+    public function count(): int
+    {
+        return count($this->all());
+    }
+
+    /**
+     * The staged members: base merged with attachments, minus detachments
+     *
+     * @return Model[]
+     */
+    private function all(): array
+    {
+        if ($this->replace) {
+            return array_values($this->attach);
+        }
+
+        $members = [];
+        foreach ($this->materializeBase() as $id => $model) {
+            if (isset($this->attach[$id]) && $this->attach[$id] !== $model) {
+                throw new LogicException(
+                    'Collection already contains a different model with the same primary key'
+                );
+            }
+
+            $members[$id] = $model;
+        }
+
+        foreach ($this->detach as $id => $_) {
+            unset($members[$id]);
+        }
+
+        return array_values(array_merge($members, $this->attach));
+    }
+
+    /**
+     * The stored members, read from the source query once and cached
+     *
+     * @return array<string, Model>
+     */
+    private function materializeBase(): array
+    {
+        if ($this->base === null) {
+            $this->base = [];
+            foreach ($this->source ?? [] as $model) {
+                $this->base[$this->identify($model)] = $model->setNew(false);
+            }
+
+            $this->source = null;
+        }
+
+        return $this->base;
+    }
+
+    /**
+     * Return the primary key of the model, or an object hash if it is not set
+     *
+     * @param Model $model
+     *
+     * @return string
+     */
+    private function identify(Model $model): string
+    {
+        $values = [];
+        foreach ((array) $model->getKeyName() as $column) {
+            if (! $model->hasProperty($column) || $model->$column === null) {
+                return 'obj#' . spl_object_id($model);
+            }
+
+            $values[] = $model->$column;
+        }
+
+        return 'pk#' . implode('|', $values);
+    }
+}
--- a/library/Notifications/Common/EntityManager.php
+++ b/library/Notifications/Common/EntityManager.php
@@ -1,0 +1,1034 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Notifications\Common;
+
+use DateTime;
+use ipl\Orm\Behaviors;
+use ipl\Orm\Query;
+use ipl\Orm\Relation;
+use ipl\Orm\Relation\BelongsTo;
+use ipl\Orm\Relation\BelongsToMany;
+use ipl\Orm\Resolver;
+use ipl\Sql\Connection;
+use ipl\Sql\Expression;
+use ipl\Sql\Select;
+use PDO;
+use RuntimeException;
+use SplObjectStorage;
+
+/**
+ * Persists models and their related models to the database.
+ *
+ * The EntityManager is the write-side counterpart to {@see Query}. Whereas a query is bound to a model
+ * class and operates on rows matching a filter, the EntityManager operates on concrete {@see Model} instances.
+ *
+ * ```
+ * $em = new EntityManager($db);
+ * $em->save($car);                 // INSERT or UPDATE, depending on whether the model is new
+ * $em->save($car->delete());       // DELETE or soft-delete, depending on the model
+ * ```
+ *
+ * Whether {@see self::save()} inserts or updates is decided by {@see Model::isNew()}. Updates only write the
+ * properties changed since the model was loaded ({@see Model::getModifiedProperties()}). Set relations are cascaded
+ * (parents first, then the model, then children and many-to-many links), all within a single transaction.
+ *
+ * Calling {@see Model::delete()} on a model and passing it to {@see self::save()} will soft or hard delete the
+ * model according to whether its table has a {@see Model::DELETED} column.
+ *
+ * In case a model supports tracking of modification times ({@see Model::getChangedAtColumn()}), the EntityManager
+ * will stamp the resulting record with the current time when it is persisted. This is done in a way that phantom
+ * reads by the daemon are avoided by locking related rows using a range lock, given the current transaction isolation
+ * level is serializable.
+ *
+ * The optional baseline time and date is used to verify that no related database record exceeds it as if that's
+ * the case, it is considered a parallel update that would be overwritten by the save. This is a simple form of
+ * optimistic locking, and is used to prevent overwriting changes made by other processes since the data was loaded.
+ * This is limited to relation models supporting tracking of modification times. For a base model the caller is
+ * responsible to detect conflicts. This is independent of the used driver or transaction isolation level.
+ *
+ * Limitations:
+ * - Saving two separate model instances that map to the same db row will result in the second save overwriting the
+ *   changes of the first
+ * - {@see self::saveGraph()} does not support cyclic graphs, passing a cyclic graph like:
+ *   ```
+ *   $parent->children = [$child];
+ *   $child->parent = $parent;
+ *   $em->save($parent);
+ *   ```
+ *   will throw a {@see RuntimeException}
+ */
+class EntityManager
+{
+    /** @var Connection The database connection to persist to */
+    protected Connection $db;
+
+    /** @var ?DateTime Baseline which no relation record must exceed */
+    protected ?DateTime $baselineAt;
+
+    /**
+     * The models currently being saved by {@see self::saveGraph()}, used to detect cyclic graphs
+     *
+     * @var SplObjectStorage<Model, null>
+     */
+    private SplObjectStorage $activeSaves;
+
+    /**
+     * Cache of table column maps populated by {@see self::getTableColumnMap()}, keyed by model class
+     *
+     * @var array<class-string, array<string, true>>
+     */
+    private array $tableColumnCache = [];
+
+    /**
+     * The resolver shared across all models
+     *
+     * @var Resolver
+     */
+    private Resolver $resolver;
+
+    /**
+     * Create a new EntityManager for the given database connection
+     *
+     * @param Connection $db
+     * @param ?DateTime $baselineAt Baseline which no relation record must exceed
+     */
+    public function __construct(Connection $db, ?DateTime $baselineAt = null)
+    {
+        $this->db = $db;
+        $this->baselineAt = $baselineAt;
+        $this->activeSaves = new SplObjectStorage();
+        $this->resolver = (new Query())
+            ->setDb($db)
+            ->getResolver();
+    }
+
+    /**
+     * Persist the given model and its set relations
+     *
+     * @param Model $model
+     *
+     * @return void
+     *
+     * @throws RuntimeException
+     */
+    public function save(Model $model): void
+    {
+        if ($this->db->inTransaction()) {
+            $this->saveGraph($model);
+        } else {
+            $this->db->transaction(function () use ($model): void {
+                $this->saveGraph($model);
+            });
+        }
+    }
+
+    /**
+     * Delete the given model's row and reset it to a fresh state
+     *
+     * Soft-deletes when the model's table has a {@see Model::DELETED} column, otherwise hard-deletes
+     * the row. Does nothing if the model is new.
+     *
+     * @param Model $model
+     *
+     * @return void
+     *
+     * @throws RuntimeException If the given model has no (complete) primary key value
+     */
+    protected function delete(Model $model): void
+    {
+        if ($model->isNew()) {
+            return;
+        }
+
+        $behaviors = $this->resolver->getBehaviors($model);
+
+        if ($model->isSoftDeletable()) {
+            $model->{$model::DELETED} = true;
+            $this->persist($model, $behaviors);
+
+            return;
+        }
+
+        $condition = $this->createPrimaryKeyCondition($model, $behaviors);
+        if ($condition === null) {
+            throw new RuntimeException(
+                sprintf(
+                    'Cannot delete %s without a primary key value',
+                    get_class($model)
+                )
+            );
+        }
+
+        $this->ensureOlderThanBaseline($model);
+        $this->db->delete($model->getTableName(), $condition);
+        foreach ((array) $model->getKeyName() as $k) {
+            unset($model->$k);
+        }
+
+        $model->clearModifiedProperties();
+        $model->setNew();
+    }
+
+    /**
+     * Recursively persist the given model and its set relations
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    protected function saveGraph(Model $model): void
+    {
+        if (! $model->isMutable()) {
+            throw new RuntimeException(sprintf(
+                'setNew() must be called on a model of type %s before saving it',
+                get_class($model)
+            ));
+        }
+
+        if ($this->activeSaves->offsetExists($model)) {
+            throw new RuntimeException('Reference loop detected, failed to save');
+        }
+
+        $this->activeSaves->offsetSet($model);
+
+        // Snapshot what to cascade before persisting, since persisting resets change tracking. Only
+        // explicitly set relations are considered (lazy loaders are closures, skipped by the iterator).
+        // A relation is cascaded when the caller (re)assigned it, or when its already-materialized value
+        // has pending changes of its own.
+        $set = iterator_to_array($model);
+        $isNew = $model->isNew();
+        $modifiedRelations = $isNew ? [] : $model->getModifiedProperties();
+
+        /** @var array<string, BelongsTo> $dependencies */
+        $dependencies = [];
+        /** @var array<string, Relation> $children */
+        $children = [];
+        /** @var array<string, BelongsToMany> $manyToMany */
+        $manyToMany = [];
+        foreach ($this->resolver->getRelations($model) as $name => $relation) {
+            if (
+                ! array_key_exists($name, $set)
+                || (! $isNew && ! isset($modifiedRelations[$name]) && ! $this->hasPendingChanges($set[$name]))
+            ) {
+                // The relation has no changes to persist
+                continue;
+            }
+
+            if ($relation instanceof BelongsTo) {
+                $dependencies[$name] = $relation;
+            } elseif ($relation instanceof BelongsToMany) {
+                $manyToMany[$name] = $relation;
+            } else {
+                $children[$name] = $relation;
+            }
+        }
+
+        if ($model->isMarkedForDeletion()) {
+            // Delete path
+            $this->saveManyToMany($model, $manyToMany, $set);
+            $this->saveChildren($model, $children, $set);
+            $this->delete($model);
+            $this->saveChangedParents($dependencies, $set);
+        } else {
+            // Insert / Update
+            $this->saveParents($model, $dependencies, $set);
+            $this->persist($model, $this->resolver->getBehaviors($model));
+            $this->saveChildren($model, $children, $set);
+            $this->saveManyToMany($model, $manyToMany, $set);
+        }
+
+        $this->activeSaves->offsetUnset($model);
+    }
+
+    /**
+     * Persist the model's BelongsTo parents and copy their keys into the model's foreign keys
+     *
+     * Parents are persisted before the model so the foreign key on this (source) table points at an
+     * existing row. An explicitly assigned `null` clears the foreign key instead.
+     *
+     * @param Model $model
+     * @param array<string, BelongsTo> $dependencies
+     * @param array<string, mixed> $set
+     *
+     * @return void
+     */
+    private function saveParents(Model $model, array $dependencies, array $set): void
+    {
+        foreach ($dependencies as $name => $relation) {
+            $related = $set[$name];
+            if ($related === null) {
+                foreach ($relation->determineKeys($model) as $sourceColumn) {
+                    $model->$sourceColumn = null;
+                }
+
+                continue;
+            }
+
+            if (! $related instanceof Model) {
+                continue;
+            }
+
+            if ($related->isMutable()) {
+                $this->saveGraph($related);
+            }
+
+            foreach ($relation->determineKeys($model) as $targetColumn => $sourceColumn) {
+                $model->$sourceColumn = $related->$targetColumn;
+            }
+        }
+    }
+
+    /**
+     * Persist the model's HasOne/HasMany children, copying the model's key into each child's foreign key
+     *
+     * Children are persisted after the model so its (possibly generated) key is known and can be copied in.
+     * When the relation's collection is in replace mode ({@see Collection::sync()}), any stored child no longer
+     * in the desired set is removed afterwards via {@see self::reconcileChildren()}. In merge mode, explicitly
+     * detached children are deleted.
+     *
+     * @param Model $model
+     * @param array<string, Relation> $children
+     * @param array<string, mixed> $set
+     *
+     * @return void
+     */
+    private function saveChildren(Model $model, array $children, array $set): void
+    {
+        foreach ($children as $name => $relation) {
+            $keys = $relation->determineKeys($model);
+            $value = $set[$name];
+            $desired = [];
+            foreach ($this->childrenOf($value) as $child) {
+                foreach ($keys as $targetColumn => $sourceColumn) {
+                    $child->$targetColumn = $model->$sourceColumn;
+                }
+
+                $this->saveGraph($child);
+                $desired[] = $child;
+            }
+
+            if ($value instanceof Collection) {
+                if ($value->isReplacing()) {
+                    if ($model->isMarkedForDeletion() && ! empty($desired)) {
+                        throw new RuntimeException(
+                            'A relation of a deleted model in replace mode should not carry attachments'
+                        );
+                    }
+
+                    $this->reconcileChildren($relation, $model, $keys, $desired);
+                } else {
+                    foreach ($value->getDetachments() as $child) {
+                        $this->delete($child);
+                    }
+                }
+
+                $value->clearPendingChanges();
+            }
+        }
+    }
+
+    /**
+     * Remove the stored children of a HasOne/HasMany relation that are absent from the desired set
+     *
+     * Only invoked for a collection in replace mode
+     *
+     * @param Relation $relation The HasOne/HasMany relation being reconciled
+     * @param Model $model The source model whose key scopes the children
+     * @param array<string, string> $keys The relation's foreign => source key map ({@see Relation::determineKeys()})
+     * @param Model[] $desired The children that should remain, already persisted
+     *
+     * @return void
+     */
+    private function reconcileChildren(Relation $relation, Model $model, array $keys, array $desired): void
+    {
+        $targetClass = $relation->getTargetClass();
+        $childModel = new $targetClass();
+        $childBehaviors = $this->resolver->getBehaviors($childModel);
+        $keyColumns = (array) $childModel->getKeyName();
+
+        // Restrict to the children sharing this model's key, persisted with the child's behaviors
+        $condition = [];
+        foreach ($keys as $foreignColumn => $sourceColumn) {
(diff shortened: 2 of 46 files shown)


test/php/Lib/EntityManager/Charm.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Tests\Icinga\Module\Notifications\Lib\EntityManager;
+
+use Icinga\Module\Notifications\Common\Model;
+use ipl\Orm\Behavior\Binary;
+use ipl\Orm\Behaviors;
+use ipl\Orm\Relations;
+
+class Charm extends Model
+{
+    public function getTableName()
```
sukhwinder33445: Please add return types to all new classes.

test/php/Lib/EntityManager/Gadget.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function createRelations(Relations $relations)
+    {
+        $relations->belongsTo('workshop', Workshop::class);
+
+        $relations->belongsToMany('stickers', Sticker::class)
+            ->through('gadget_sticker');
+
+        // Declared through a junction model that carries a `deleted` column, so the EntityManager
+        // reconciles these links with soft-deletes and revives instead of hard deletes.
```
sukhwinder33445: There is no direct functionality for soft delete.

test/php/Lib/EntityManager/GadgetTag.php
```diff
@@ -0,0 +1,42 @@
...
+use ipl\Orm\Behavior\MillisecondTimestamp;
+use ipl\Orm\Behaviors;
+
+/**
+ * Soft-delete junction model linking {@see Gadget} and {@see Tag}.
+ *
+ * Carries a `deleted` column (and `changed_at`), so the EntityManager reconciles the link table with
+ * soft-deletes and revives rather than hard deletes — mirroring the real contactgroup_member and
+ * rule_escalation_recipient junctions.
+ */
```
sukhwinder33445, line 19: There is no direct functionality for soft delete. Please update comment this here and in all other places.

library/Notifications/Common/StatefulQuery.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Mark yielded models as loaded so subsequent changes are tracked as updates
+     *
+     * @inheritDoc
+     *
+     * @return Generator
+     */
+    public function yieldResults(): Generator
+    {
+        yield from Seq::map(parent::yieldResults(), fn (Model $model) => $model->setNew(false));
```
sukhwinder33445: I would prefer to use `foreach` directly, as calling this method doesn’t offer any advantage.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Notifications\Common;
+
+use ipl\Sql\Connection;
+
+/**
+ * Base class for all Models of the module, properties of instances may be edited and can be persisted by
+ * passing them to the EntityManager
+ */
```
sukhwinder33445, line 20: This class tracks changes made to the model. Please rewrite this comment accordingly.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+     * @var list<Model>
+     */
+    private array $pendingDeletions = [];
+
+    /**
+     * Get whether this entity is new, i.e. not yet persisted to the database
+     *
+     * @return bool
+     */
+    public function isNew(): bool
```
sukhwinder33445, line 55: `Get whether this entity is newly created and does not yet exist in the database.`

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return $this->isNew;
+    }
+
+    /**
+     * Set whether this entity is new, i.e. not yet persisted to the database
```
sukhwinder33445: Same here.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return $this->dirtyProperties;
+    }
+
+    /**
+     * Reset change tracking and accept the current values as the new baseline
+     *
+     * @return $this
+     */
+    public function markClean(): static
```
sukhwinder33445: Please rename to `clearDirtyProperties()`.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+
+    /**
+     * Insert or update the given model's own row
+     *
+     * @param Model $model
+     * @param Resolver $resolver
+     *
+     * @return void
+     */
+    protected function persist(Model $model, Resolver $resolver): void
```
sukhwinder33445: Please pass the `$behavior` instead of `$resolver` as param.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+            }
+        } else {
+            foreach ($model as $property => $value) {
+                if (! isset($columns[$property])) {
+                    continue;
+                }
+
+                $data[$columns[$property]] = $behaviors->persistProperty($value, $property);
+            }
+        }
```
sukhwinder33445: I think this can be simlified as follows:
```suggestion
$toIterate = $only ?? $model;
foreach ($toIterate as $property => $_) {
    if (! isset($columns[$property])) {
        continue;
    }

    $data[$columns[$property]] = $behaviors->persistProperty($model->$property, $property);
}
```

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    /**
+     * Turn a column => value map into an equality WHERE scope (`column = ?`), as ipl-sql expects
+     *
+     * @param array<string, mixed> $columns
+     *
+     * @return array<string, mixed>
+     */
+    private function equalityScope(array $columns): array
```
sukhwinder33445: Please rename this method to something more fitting, e.g. `createCondition()`.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+     * Forget any models queued for deletion via {@see deleteOnSave()}
+     *
+     * @return $this
+     */
+    public function clearPendingDeletions(): static
+    {
+        $this->pendingDeletions = [];
+
+        return $this;
+    }
```
sukhwinder33445, line 186: As we discussed in person, you pointed out that this code isn’t really necessary and could be removed.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+            foreach ($this as $k => $v) {
+                if ($k === $key) {
+                    $hadValue = true;
+                    $original = $v;
+                    break;
+                }
+            }
+
+            if (! $hadValue || $original !== $value) {
+                $this->dirtyProperties[$key] = true;
```
sukhwinder33445: The `$key` can be either a column or a relation, please adjust the property name and documentation accordingly.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,180 @@
...
+
+        return isset($this->dirtyProperties[$property]);
+    }
+
+    /**
+     * Get the names of all properties changed since the entity was loaded as a set keyed by name
+     *
+     * @return array<string, true>
+     */
+    public function getDirtyMap(): array
```
sukhwinder33445: Please use a more appropriate and descriptive name. As the stored keys may also be relations, this should be stated in the documentation.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+
+    /**
+     * Build a WHERE scope matching the given model by its primary key, converting values for persistence
+     *
+     * @param Model $model
+     * @param Behaviors $behaviors
+     *
+     * @return ?array<string, mixed> Null if the model has no value for (part of) its primary key
+     */
+    protected function keyScope(Model $model, Behaviors $behaviors): ?array
```
sukhwinder33445: Please choose a suitable name.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+     * @return void
+     */
+    private function markLink(string $table, array $scope, string $deleted, int $changedAt): void
+    {
+        $this->db->update(
+            $table,
+            ['deleted' => $deleted, 'changed_at' => $changedAt],
+            $this->equalityScope($scope)
+        );
+    }
```
sukhwinder33445: TBD: I think that’s too much for this class. The parent repository class should explicitly set the tables.

sukhwinder33445 (changes requested):
This for now.
Note to myself: I need to debug and better understand some parts of the code before reviewing it properly (saveGraph, saveLinks, fetchLinks, reconcileSoftLinks, markLink, insertRows).

test/php/Lib/EntityManager/GadgetTag.php
BastianLedererIcinga: I made no changes yet regarding this yet, because it depends on what we agree on regarding https://github.com/Icinga/icinga-notifications-web/pull/482#discussion_r3420804258.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+            $targetBehaviors = $this->resolverFor($target)->getBehaviors($target);
+            $value = $targetBehaviors->persistProperty($target->$targetColumn, $targetColumn);
+            $desired[(string) $value] = $value;
+        }
+
+        if (in_array('deleted', $junction->getColumns(), true)) {
+            $this->reconcileSoftLinks($junction, $sourceColumns, $junctionColumn, $desired);
+
+            return;
+        }
```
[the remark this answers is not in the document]
BastianLedererIcinga: All relations using many-to-many junctions with soft deletes are now declared with the model class.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,689 @@
...
+
+        $data = $this->extract($model, $behaviors, $model->getDirtyMap());
+        if (empty($data)) {
+            // Only relations changed; there is nothing to update on this row
+            $model->markClean();
+
+            return;
+        }
```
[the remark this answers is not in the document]
BastianLedererIcinga: fixed in 52ced4b961389239ff357ea3dc9c781f558f4b85

BastianLedererIcinga:
@sukhwinder33445 As discussed the `EntitiyManager` now supports soft deletes.
I also tried clean up the code related to `EntityManger::saveGraph()`, I hope it is easier to understand now.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Mark the model for deletion on the next {@see EntityManager::save()} and return it
+     *
+     * If the model uses soft deletes this function must set the `deleted` property
+     *
+     * @return $this
+     */
+    public function markDeleted(): static
```
BastianLedererIcinga: I think it makes sense to rename this to `delete()`, but that requires a rename of the existing `Model\Rotation::delete()`, which I wanted to avoid until we agreed that the function is in the right place here.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,812 @@
...
+     *
+     * Internal helper for the delete flow: callers mark a model with {@see Model::markDeleted()} and
+     * pass it to {@see save()}. Does nothing if the model is new or has no primary key value. Nested
+     * records are expected to be removed by the database (e.g. `ON DELETE CASCADE`).
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    protected function delete(Model $model): void
```
sukhwinder33445, line 140: I think this should remain public, and internally decide whether to soft delete or permanently delete.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,812 @@
...
+        $dependencies = [];
+        /** @var array<string, Relation> $children */
+        $children = [];
+        /** @var array<string, BelongsToMany> $manyToMany */
+        $manyToMany = [];
+        foreach ($resolver->getRelations($model) as $name => $relation) {
+            if (
+                ! array_key_exists($name, $set) ||
+                ! $isNew && ! isset($modifiedRelations[$name]) && ! $this->hasPendingChanges($set[$name])
+            ) {
```
sukhwinder33445: Please fix indentation wrap secound condition in parenthesis.
```Suggestion
if (
    ! array_key_exists($name, $set)
    || (! $isNew && ! isset($modifiedRelations[$name]) && ! $this->hasPendingChanges($set[$name]))
) {
```

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,812 @@
...
+            $this->saveManyToMany($model, $manyToMany, $set);
+            $this->saveDeletedChildren($children, $set);
+
+            if ($model->isSoftDeletable()) {
+                $this->persist($model, $resolver->getBehaviors($model));
+            } else {
+                $this->delete($model);
+            }
+
+            $this->saveChangedParents($dependencies, $set);
```
sukhwinder33445, line 234: As mentioned above, this should be part of the `delete()` function, and the `delete()` function should be called here instead.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,812 @@
...
+     * @param array<string, mixed> $set
+     *
+     * @return void
+     */
+    private function saveDeletedChildren(array $children, array $set): void
+    {
+        foreach ($children as $name => $relation) {
+            foreach ($this->asTraversable($set[$name]) as $child) {
+                if ($child instanceof Model && $child->isMarkedForDeletion()) {
+                    $this->saveGraph($child);
```
sukhwinder33445: Use `delete()` instead, which performs a soft delete if necessary.

sukhwinder33445 (commented):
Once this is done, I’ll review it further, as it may require changes to other methods.

library/Notifications/Common/EntityManager.php
BastianLedererIcinga: Moving the soft/hard delete logic here is a good idea, but as discussed there is no reason to make it `public`.

library/Notifications/Common/EntityManager.php
BastianLedererIcinga: This is not only about the soft vs. hard delete decision, but also about the recursion.
A plain `delete()` call would not walk related models that still carry changes.

library/Notifications/Common/EntityManager.php
BastianLedererIcinga: I only moved
```
if ($model->isSoftDeletable()) {
    $this->persist($model, $resolver->getBehaviors($model));
} else {
    $this->delete($model);
}
```
into `delete()`, moving the rest of this code would require to pass way more parameters to `delete()`, and I don't think that would help with the readability of the code.

library/Notifications/Common/StatefulQuery.php
```diff
@@ -0,0 +1,61 @@
...
+     * Mark yielded models as loaded so subsequent changes are tracked as updates
+     *
+     * @inheritDoc
+     *
+     * @return Generator
+     */
+    public function yieldResults(): Generator
+    {
+        foreach (parent::yieldResults() as $key => $model) {
+            if ($model instanceof Model) {
```
sukhwinder33445: This condition should always be true, as the parent method always returns Model. Right?

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+ * Calling {@see Model::markDeleted()} on a model and passing it to {@see save()} will soft or hard delete the
+ * model according to the value of {@see Model::isSoftDeletable()}
+ *
+ * Limitations:
+ * - Saving two separate model instances that map to the same db row will result in the second save overwriting the
+ *   changes of the first
+ * - ManyToMany links no longer present after a save are soft- or hard deleted according to what
+ *   {@see Model::isSoftDeletable()} returns. This only works if the relation was declared using its model class
+ *   which must have a mapping for the `deleted` column, relations that use the table name will always use hard delete.
+ * - {@see saveGraph()} does not detect cycles, passing a cyclic graph like:
```
sukhwinder33445: Methods of this class must be prefixed with `self::` to ensure they are linked correctly in the PHPDoc.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+ *   ```
+ *   will throw a {@see RuntimeException}
+ */
+class EntityManager
+{
+    /** @var Connection The database connection to persist to */
+    protected Connection $db;
+
+    /**
+     * The models currently being saved by {@see saveGraph()}, used to detect cyclic graphs
```
sukhwinder33445: Same here.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+        $behaviors = $this->resolverFor($model)->getBehaviors($model);
+
+        if ($model->isSoftDeletable()) {
+            $this->persist($model, $behaviors);
+
+            return;
+        }
+
+        $condition = $this->createPrimaryKeyCondition($model, $behaviors);
+        if ($condition === null) {
```
sukhwinder33445, line 156: Why not throw an exception here, as `persist()` does?

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+            return;
+        }
+
+        $this->db->delete($model->getTableName(), $condition);
+        foreach ((array) $model->getKeyName() as $k) {
+            unset($model->$k);
+        }
+
+        $model->setNew(true);
```
sukhwinder33445: Param is superfluous here.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+     * Insert or update the given model's own row
+     *
+     * @param Model $model
+     * @param Behaviors $behaviors
+     *
+     * @return void
+     */
+    protected function persist(Model $model, Behaviors $behaviors): void
+    {
+        if (! $model->isNew() && ! $model->isModified() && ! $model->isMarkedForDeletion()) {
```
sukhwinder33445: As this method can only performs a soft-delete and `isModified()` returns `true` when the `delete` flag is set, the method call `$model->isMarkedForDeletion()` seems superfluous to me.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+            }
+        }
+
+        $data = $this->extract($model, $behaviors, $model->getModifiedProperties());
+        if (empty($data)) {
+            // Only relations changed; there is nothing to update on this row
+            $model->clearModifiedProperties();
+
+            return;
+        }
```
sukhwinder33445:
```suggestion
if (empty(array_intersect_key($model->getModifiedProperties(), $this->writableColumns($model)))) {
    // Only relations changed; there is nothing to update on this row
    $model->clearModifiedProperties();

    return;
}
```
This can be used to check whether anything has changed in the model.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+        $junctionColumn = $targetKeys[$targetColumn];
+
+        $desired = [];
+        foreach ($targets as $target) {
+            $targetBehaviors = $this->resolverFor($target)->getBehaviors($target);
+            $value = $targetBehaviors->persistProperty($target->$targetColumn, $targetColumn);
+            $desired[(string) $value] = $value;
+        }
+
+        if ($this->isSoftDeleteJunction($junction)) {
```
sukhwinder33445: This method is only used once, so please use the condition directly and remove this method.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+     * @param array<array<string, mixed>> $rows Each row as a column => value map
+     *
+     * @return void
+     */
+    protected function insertRows(string $table, array $rows): void
+    {
+        foreach ($rows as $row) {
+            $this->db->insert($table, $row);
+        }
+    }
```
sukhwinder33445: This method is only used once, so please use the code directly and remove this method.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+     *
+     * Schema-wide convention; not implemented as a behavior so individual models don't have to opt in.
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    protected function stampChangedAt(Model $model): void
+    {
+        if (isset($this->writableColumns($model)['changed_at'])) {
```
sukhwinder33445: The `changed_at` flag is still hard-coded; the `EntityManager` should not know what the flag is called.

test/php/library/Notifications/Common/EntityManagerTest.php
```diff
@@ -0,0 +1,1394 @@
...
+    public function testInsertAssignsGeneratedKeyAndMarksNotNew()
+    {
+        $workshop = new Workshop();
+        $this->assertTrue($workshop->isNew());
+        $workshop->name = 'Acme';
+
+        $this->em()->save($workshop);
+
+        $this->assertFalse($workshop->isNew(), 'A saved model is no longer new');
+        $this->assertSame(1, $workshop->id, 'The generated primary key is written back to the model');
```
sukhwinder33445: The message should explain why the test failed; this seems misleading.

```suggestion
$this->assertSame(1, $workshop->id, 'The generated primary key should be written back to the model);
```

sukhwinder33445: Please fix all the messages accordingly.

test/php/library/Notifications/Common/EntityManagerTest.php
```diff
@@ -0,0 +1,1394 @@
...
+        $gadget = new Gadget();
+        $gadget->name = 'Spanner';
+
+        $sticker = new Sticker();
+        $sticker->label = 'fragile';
+        $gadget->stickers = [$sticker];
+
+        $this->em()->save($gadget);
+
+        $this->assertFalse($sticker->isNew(), 'The target was persisted');
```
sukhwinder33445: The message should explain why the test failed; this seems misleading.

```suggestion
        $this->assertFalse($sticker->isNew(), 'A saved model should not be new');
```

test/php/library/Notifications/Common/EntityManagerTest.php
```diff
@@ -0,0 +1,1394 @@
...
+        $stamped = new Stamped();
+        $stamped->name = 'Widget';
+        $this->em()->save($stamped);
+
+        // Load it fresh — the relation comes back as a closure-backed lazy loader.
+        $loaded = null;
+        foreach (Stamped::on($this->db)->execute() as $row) {
+            $loaded = $row;
+            break;
+        }
```
sukhwinder33445: Why not `$loaded = Stamped::on($this->db)->first()`;  instead?

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,833 @@
...
+     * @return void
+     */
+    protected function saveGraph(Model $model): void
+    {
+        if ($model->isNew() && $model->isMarkedForDeletion()) {
+            return;
+        }
+
+        if ($this->activeSaves->offsetExists($model)) {
+            throw new RuntimeException('Cannot save a cyclic graph');
```
sukhwinder33445: This message is unclear. Use something like `Reference loop detected, failed to save`.

sukhwinder33445 (changes requested):

library/Notifications/Common/StatefulQuery.php
BastianLedererIcinga: The parent yields `\ipl\Orm\Model`, we check for `Icinga\Module\Notifications\Common\Model`.
The guard might still be overkill, because it is only relevant in case of a weird misconfiguration.

I added it in response to [this comment ](https://github.com/Icinga/icinga-notifications-web/pull/482#discussion_r3426741636), because I thought since you requested the copilot review and left the comment open you agreed with its suggestion.

sukhwinder33445: Yes, we check for `Icinga\Module\Notifications\Common\Model`. The condition is then fine.

sukhwinder33445 (dismissed):

library/Notifications/Common/StatefulQuery.php
nilmerg, line 1: As discussed offline already, I'd like to see a `markDeleted()` method to let the manager know that all models yielded are to be deleted without the need to traverse them and mark them individually. Though, now I think `deleteAll()` fits better, does it?

library/Notifications/Common/StatefulQuery.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Recursively mark the given model and its eagerly-loaded related models as loaded
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    private function markLoaded(Model $model): void
```
nilmerg: I see an, albeit unrealistic, risk for infinite recursion here. Just consider someone using `$query->with('service.host.service')`.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,814 @@
...
+                $this->saveGraph($model);
+            });
+        }
+    }
+
+    /**
+     * Delete the given model's row and reset it to a fresh state
+     *
+     * Soft-deletes via the configured behavior when the model is soft-deletable, otherwise hard-deletes
+     * the row. Does nothing if the model is new or has no primary key value.
```
nilmerg: *Does nothing* is a bit mis-leading in case of no primary key, don't ya think? :P

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,814 @@
...
+     *
+     * Expression columns are omitted as they cannot be written to. The result is memoized in
+     * {@see $writableColumnCache} by class, since the column set never changes once a Model
+     * class is defined; repeated calls within the same save graph reuse the cached map.
+     *
+     * @param Model $model
+     *
+     * @return array<string, string>
+     */
+    protected function writableColumns(Model $model): array
```
nilmerg: I wondered at first what this method does and why it's necessary. Now that I see it, doesn't it beg the question whether `Model::getColumns()` is the best candidate or rather a `Model::getRealColumns()` should be introduced? (Of course, the name is not final, but I didn't come up with a better name spontaneously)

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,814 @@
...
+        }
+
+        return $this->createCondition($columns);
+    }
+
+    /**
+     * Sync the junction so it links the given source to exactly the given targets
+     *
+     * Assignment is authoritative, the stored set is replaced by the given one. Links absent from $targets
+     * are deleted, new ones inserted, existing ones remain.
```
nilmerg: We discussed this already offline, that this isn't good, to remove something not explicitly marked for deletion. An empty relation (null, empty ResultSet or array) must be ignored and even if all entries are marked for deletion this doesn't mean *all* links are to be dropped.

library/Notifications/Common/Model.php
nilmerg: Makes absolute sense! My current changes removed `Rotation::delete` already so feel free to rename it to make room for this change.

nilmerg (changes requested):
Good job so far! Especially the structure of `EntityManager` is pretty clean :)

library/Notifications/Common/EntityManager.php
BastianLedererIcinga: True, I forgot to adjust the docstring, the code already thows a `RuntimeException` in this case.

library/Notifications/Common/StatefulQuery.php
BastianLedererIcinga: I implemented this in dc424d0de7722f39a056aff9747648bc47cb5059. It removes the `RuntimeException` from `Model::isNew()` that we agreed on in person, so `markLoaded()` can assume that any model that doesn't return `null` for `isNew()` has already been touched, and returns.
`EntityManager::saveGraph()` will now throw the `RuntimeException` instead if it encounters a model that still returns `null` for `isNew()`.

library/Notifications/Common/EntityManager.php
BastianLedererIcinga: This is implemented in 86df65560b21db0de04429325427f03ae28ad6d7.

library/Notifications/Common/StatefulQuery.php
BastianLedererIcinga: This is implemented in bc83d291c3e77a404c137feccda35097fa988002.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,234 @@
...
+
+            if (is_int($alias)) {
+                $columns[$column] = $column;
+            } else {
+                $columns[$alias] = $column;
+            }
+        }
+
+        return $columns;
+    }
```
nilmerg, line 196: I rather had a more static approach in mind. Differences in *columns* compared to the actual table columns manifest usually only as the primary key being missing. Expressions are possible, but really an edge case as at the moment none of our notifications models have any. So here are my suggestions:

* Rename to `getTableColumns`: It's more in-line with `getTableName` and implies the relevance already in the name. You don't have to explain this so the doc comment gets shorter as well.
* Let it return `array_merge((array) $this->getKeyName(), $this->getColumns())` by default. Any concrete model should have to override this to ensure only strings are returned. The strings being keys, is a requirement of the `EntityManager` and should therefore its business, not the model's.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Get whether the model's table uses soft deletes
+     *
+     * @return bool
+     */
+    public function isSoftDeletable(): bool
+    {
+        return in_array('deleted', $this->getColumns(), true);
```
nilmerg: This should utilize `getTableColumns`

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,234 @@
...
+     * Get the column used to store the timestamp of the most recent modification to the row
+     *
+     * `changed_at` is the schema-wide convention, the {@see EntityManager} checks whether the column
+     * exists on the model before stamping it.
+     *
+     * @return string
+     */
+    public function getChangedAtColumn(): string
+    {
+        return 'changed_at';
```
nilmerg: We just noticed recently that a particular table had no `changed_at` yet. So isn't it possible that it does not exist elsewhere yet? And even if not, it may only be *our schema's convention*, remember that I want to see this as prototype for utilization elsewhere. So just like `isSoftDeletable`, this should be able to identify a missing `changed_at` and return `null`.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Base class for all module models that tracks the changes made to a model
+ *
+ * Records which properties have changed since the model was loaded, and whether the model has been
+ * persisted yet, so the {@see EntityManager} can store a model and write only what actually changed.
+ *
+ * {@see self::setNew()} must be called explicitly when creating a new instance, it tells the {@see EntityManager}
+ * whether to insert or update
+ */
+abstract class Model extends \ipl\Orm\Model
```
nilmerg, line 21: Please add constants for `deleted` and `changed_at`. They're easier to override than methods.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,234 @@
...
+
+            if (! $hadValue || $original !== $value) {
+                $this->modifiedProperties[$key] = true;
+            }
+        }
+
+        return parent::setProperty($key, $value);
+    }
+
+    public static function on(Connection $db): StatefulQuery
```
nilmerg: I'm more and more convinced that this should possibly be a new method: `forUpdate(Connection $db): StatefulQuery`

It clearly signals the intention and `::on` keeps being the read-only case without the additional overhead.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,761 @@
...
+        }
+
+        $this->activeSaves->offsetSet($model);
+        $resolver = $this->resolverFor($model);
+
+        // Snapshot what to cascade before persisting, since persisting resets change tracking. Only
+        // explicitly set relations are considered (lazy loaders are closures, skipped by the iterator).
+        // A relation is cascaded when the caller (re)assigned it, or when its already-materialized value
+        // has pending changes of its own — so an in-place edit to a loaded related model is persisted.
+        // An explicit `null` clears the relation.
```
nilmerg: That's outdated now

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,761 @@
...
+    }
+
+    /**
+     * Get a resolver bound to the given model instance
+     *
+     * @param Model $model
+     *
+     * @return Resolver
+     */
+    protected function resolverFor(Model $model): Resolver
```
nilmerg: I don't see a reason for model-specific resolvers here. `Resolver` caches behaviors, relations etc already and this bypasses it. Try to use the same resolver each time instead.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,761 @@
...
+        $junctionColumn = $targetKeys[$targetColumn];
+
+        $desired = [];
+        foreach ($targets as $target) {
+            $targetBehaviors = $this->resolverFor($target)->getBehaviors($target);
+            $value = $targetBehaviors->persistProperty($target->$targetColumn, $targetColumn);
+            $desired[(string) $value] = $value;
+        }
+
+        if (! $junction instanceof Junction && $junction->isSoftDeletable()) {
```
nilmerg: Isn't it more fool-proof to check for `Notifications\Common\Model`?

nilmerg (changes requested):

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,761 @@
...
+     * Recursively persist the given model and its set relations
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    protected function saveGraph(Model $model): void
+    {
+        if ($model->isNew() === null) {
+            throw new RuntimeException('setNew() must be called on a model before saving it');
```
nilmerg: It'd be nice if you include the model's class here, otherwise it's hard to identify the cause.

nilmerg (commented):
That's not a final review, its only purpose is to let GitHub save what I've read already and some temporary remarks:

* `Model::isToManyRelation()` is what I've mentioned already offline. A helper to get access to relations as only the resolver knows about them. The downside, so to say, is that this defeats the intention of keeping state away from a model other than its properties. (i.e. there's now `Model::$relations`)
* The change to `Model::getProperty()`, to let it initialize empty to-many relations with a `Collection` or to transform a `StatefulQuery` to one, is a candidate to be implemented nicely if the same access to relations would be possible: As part of `Model::createDefaults()`. But _nicely_ implies not needing to call `Model::createRelations()`…
* It's not clear to me why you've overridden `Query::derive()` with `StatefulQuery::derive()`. Sure, the method doc block explains it, but my gut feeling is that this leads to a limitation/bug of the orm instead. So please show me how to reproduce this.

BastianLedererIcinga:
> * It's not clear to me why you've overridden `Query::derive()` with `StatefulQuery::derive()`. Sure, the method doc block explains it, but my gut feeling is that this leads to a limitation/bug of the orm instead. So please show me how to reproduce this.

The bug can be reproduced with this snippet.

```
$member = RotationMember::on(Database::get())->first();
$shift = $member->shift;
```
which causes:

Cannot join relation "rotation_member" in model Icinga\Module\Notifications\Model\TimeperiodEntry. Relation not found.

`RotationMember` declares the relation to `TimeperiodEntry` with the name `shift`, `TimeperiodEntry` declares its relation with the name `member`.

For the case shown above `shift` resolves to `timeperiod_entry` correctly, but `createSubQuery()` assumes that the relation declared in `TimeperiodEntry` matches the table name `rotation_member`.
 
```
public function createSubQuery(Model $target, string $targetPath, ?Model $from = null, bool $link = true): static
{
    $subQuery = (new static())
        ->setDb($this->getDb())
        ->setModel($target);


    $sourceParts = array_reverse(explode('.', $targetPath)); // ["shift", "rotation_member"]
    $sourceParts[0] = $target->getTableAlias();                       // ["timeperiod_entry", "rotation_member"]

    $subQueryResolver = $subQuery->getResolver();
    $sourcePath = join('.', $sourceParts);                                  // timeperiod_entry.rotation_member
    $subQueryTarget = $subQueryResolver->resolveRelation($sourcePath)->getTarget();
```

However the fix I implemented really makes no sense, because the error still occurs when using `Model::on()`, since it no longer uses a `StatefulQuery`.

The current override doesn't work as a general fix in ipl-orm, it still relies on the base implementation for ManyToMany relations, so I can either try to come up with a fix that covers them as well, or we accept it as a limitation and refactor the relation names. There are only a few usages of these relations (which all rely on eager loading when accessing them, so the error never surfaced there).

library/Notifications/Common/Collection.php
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Create a collection from members who are treated as already existing db rows
+     *
+     * @param iterable $source
+     *
+     * @return static
+     */
+    public static function fromLoaded(iterable $source): static
```
nilmerg: This should require an `ipl\Orm\Query`. As this is the only guaranteed reason why the resulting models should be considered loaded. The alternative, if a query isn't available, exists in form of `::create(…)`.

Once `StatefulQuery` is gone, `::materializeBase(…)` should detect that the source is a query and call `setNew(false)` on the results accordingly.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,300 @@
...
+    private ?bool $isNew = null;
+
+    /** @var bool Whether the model is marked for deletion on the next {@see EntityManager::save()} */
+    private bool $markedForDeletion = false;
+
+    /** @var array<string, true> Names of properties modified since the model was loaded */
+    private array $modifiedProperties = [];
+
+    /** @var ?Relations Lazily built relation definitions, used to tell toOne and toMany relations apart */
+    private ?Relations $relations = null;
```
nilmerg: I forgot to remove this with my commit. Please remove it on your own now.

library/Notifications/Common/Model.php
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * Allows to modify the queries created by {@see static::on()} and {@see static::forUpdate()}
+     *
+     * @param Query $query
+     *
+     * @return void
+     */
+    protected static function decorateQuery(Query $query): void
+    {
+    }
```
nilmerg: I have no further use-case for `forUpdate`. So I suggest to remove it and don't override `::on` either. The result is that `StatefulQuery` is no longer needed and can be dropped. `Collection` should then be responsible to mark loaded models as such, as suggested in my other comment.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,940 @@
...
+     */
+    private Resolver $resolver;
+
+    /**
+     * Create a new EntityManager for the given database connection
+     *
+     * @param Connection $db
+     * @param ?DateTime $baselineAt Baseline which no relation record must exceed
+     */
+    public function __construct(Connection $db, ?DateTime $baselineAt = null)
```
nilmerg, line 98: I've only introduced the argument `$baselineAt`, with no further handling. Please implement this on your own so that it behaves as described in the class-doc I added as well.

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,940 @@
...
+    /**
+     * Get the current time used for {@see self::stampChangedAt()}
+     *
+     * Override in a subclass to inject a fixed clock for testing.
+     *
+     * @return DateTime
+     */
+    protected function now(): DateTime
+    {
+        // TODO: Prevent phantom read, like in \Icinga\Module\Notifications\Common\Database::getNextChangedAt
```
nilmerg: I'd like you to implement what our `Database` trait does right now using an event listener. Since we now have a central location to handle writes, this is the perfect alternative place.

library/Notifications/Common/StatefulQuery.php
```diff
@@ -0,0 +1,129 @@
...
+     * That join walks back from the target through a relation named after the source's table alias, so
+     * a many-to-many relation is only lazily readable if the target model also declares the relation
+     * back to the source.
+     *
+     * @param string $relation
+     * @param OrmModel $source
+     *
+     * @return static
+     */
+    public function derive($relation, OrmModel $source): static
```
nilmerg: Your explanation mentions `RotationMember::$shift` as reason. Since this is unused right now, I wonder why the `EntityManager` even processes it? It is either not a `Collection` or one that has no pending changes, so it should be ignored, shouldn't it? Once `StatefulQuery` is gone, I don't think we need a replacement, but instead an issue for the aforementioned reason.

nilmerg (changes requested):

library/Notifications/Common/Collection.php
```diff
@@ -0,0 +1,269 @@
...
+     *
+     * @return Model[]
+     */
+    private function materializeBase(): array
+    {
+        if ($this->base === null) {
+            $this->base = $this->source === null ? [] : iterator_to_array($this->source, false);
+            foreach ($this->base as $model) {
+                $model->setNew(false);
+            }
```
nilmerg: Please use a single loop.

library/Notifications/Common/Collection.php
```diff
@@ -0,0 +1,269 @@
...
+    {
+        return $this->detach;
+    }
+
+    /**
+     * Whether anything must be persisted on the next save
+     *
+     * @return bool
+     */
+    public function hasPendingChanges(): bool
```
nilmerg, line 210: I just noticed an edge case here: Consider a collection that has a materialized query and one of its entries has been modified but none were attached or detached. This is right now not detected by the manager. Might it be simply an option to inspect the base here and also report a change in such a case?

library/Notifications/Common/Collection.php
```diff
@@ -0,0 +1,269 @@
...
+class Collection implements IteratorAggregate, Countable
+{
+    /**
+     * Stored members, still as the unread {@see Query}
+     *
+     * Drained into {@see self::$base} once on first read, so the query is not run eagerly.
+     *
+     * @var ?Query
+     */
+    private ?Query $source = null;
```
nilmerg, line 39: I just have the requirement to get access to the collection's query. So please add the method `query(): Query` we already discussed a few days ago. It should document how to utilize it as well:

* Calling query() only allows to adjust the query
* If modifications to the collection or its entries are desired, it must not be traversed and instead the collection must be as usual

nilmerg (changes requested):

library/Notifications/Common/EntityManager.php
```diff
@@ -0,0 +1,1022 @@
...
+     *
+     * Only children that carry their own pending deletion are touched: a live child must not be updated or
+     * inserted onto a row that is about to vanish, as that would dangle its foreign key.
+     *
+     * @param array<string, Relation> $children
+     * @param array<string, mixed> $set
+     *
+     * @return void
+     */
+    private function saveDeletedChildren(array $children, array $set): void
```
nilmerg: This misses to also delete detached children, just like `saveChildren` does.

nilmerg (dismissed):
Preliminary approval, I still have to update more current candidate usages, but it looks promising nonetheless!

nilmerg (approved):
I think it's safe to finally approve this, given the fact that I just finished refactoring the persistence of the event rule configuration which was only one left. Very good job!

Now please rebase again and we get this settled once and for all.

PR #483, opened 2026-05-29
Introduce the ability for source integrations to configure their notifications source

### LINKED ISSUE #463: Ability for source integrations to create, update and delete their Notifications configuration

sukhwinder33445:
Add `Sources` class so integrations can use it to configure and manage their notification source.

Test classes are generated using claude code.

resolves https://github.com/Icinga/icinga-notifications-web/issues/463
require https://github.com/Icinga/icinga-notifications/pull/432

Blocked by: https://github.com/Icinga/icinga-notifications-web/pull/482

### DIFF
--- a/application/controllers/SourceController.php
+++ b/application/controllers/SourceController.php
@@ -8,57 +8,77 @@ namespace Icinga\Module\Notifications\Controllers;
 use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Forms\DeleteSourceForm;
 use Icinga\Module\Notifications\Forms\SourceForm;
+use Icinga\Module\Notifications\Model\Source;
+use Icinga\Module\Notifications\Repository\SourceRepository;
 use Icinga\Web\Notification;
 use Icinga\Web\Session;
 use ipl\Html\Contract\Form;
 use ipl\Sql\Connection;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Url;
+use RuntimeException;
 
 class SourceController extends CompatController
 {
+    private Source $source;
+
     public function init(): void
     {
         $this->assertPermission('config/modules');
+
+        $source = (new SourceRepository(Database::get()))
+            ->find((int) $this->params->getRequired('id'))
+            ?->setNew(false);
+
+        if ($source === null) {
+            $this->httpNotFound($this->translate('Source not found'));
+        }
+
+        $this->source = $source;
     }
 
     public function indexAction(): void
     {
-        $sourceId = (int) $this->params->getRequired('id');
-
-        $form = (new SourceForm(Database::get()))
+        $form = (new SourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->loadSource($sourceId)
+            ->setSource($this->source)
             ->on(Form::ON_SUBMIT, function (SourceForm $form): never {
-                Database::get()->transaction(fn () => $form->editSource());
+                $source = $form->getSource();
+
+                if ($source->locked) {
+                    throw new RuntimeException('Source is locked');
+                }
+
+                Database::get()->transaction(fn (Connection $db) => (new SourceRepository($db))->update($source));
                 Notification::success(sprintf(
                     $this->translate('Updated source "%s" successfully'),
-                    $form->getSourceName()
+                    $source->name
                 ));
 
                 $this->switchToSingleColumnLayout();
             })->handleRequest($this->getServerRequest());
 
-        $this->addTitleTab(sprintf($this->translate('Source: %s'), $form->getSourceName()));
+        $this->addTitleTab(sprintf($this->translate('Source: %s'), $this->source->name));
         $this->addContent($form);
     }
 
     public function deleteAction(): void
     {
-        $sourceId = (int) $this->params->getRequired('id');
+        if ($this->source->locked) {
+            throw new RuntimeException('Source is locked');
+        }
 
         $form = (new DeleteSourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->loadSource($sourceId)
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUBMIT, function (DeleteSourceForm $form): never {
-                Database::get()->transaction(fn (Connection $db) => $form->removeSource($db));
+            ->on(Form::ON_SUBMIT, function (): never {
+                Database::get()->transaction(fn (Connection $db) => (new SourceRepository($db))->delete($this->source));
                 Notification::success($this->translate('Deleted source successfully'));
                 $this->switchToSingleColumnLayout();
             })
             ->handleRequest($this->getServerRequest());
 
-        $this->setTitle(sprintf($this->translate('Delete Source: %s'), $form->getSourceName()));
+        $this->setTitle(sprintf($this->translate('Delete Source: %s'), $this->source->name));
         $this->addContent($form);
     }
 }
--- a/application/controllers/SourcesController.php
+++ b/application/controllers/SourcesController.php
@@ -8,6 +8,7 @@ namespace Icinga\Module\Notifications\Controllers;
 use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Forms\SourceForm;
 use Icinga\Module\Notifications\Model\Source;
+use Icinga\Module\Notifications\Repository\SourceRepository;
 use Icinga\Module\Notifications\View\SourceRenderer;
 use Icinga\Module\Notifications\Web\Control\SearchBar\ObjectSuggestions;
 use Icinga\Module\Notifications\Widget\ItemList\ObjectList;
@@ -15,6 +16,7 @@ use Icinga\Web\Notification;
 use Icinga\Web\Session;
 use Icinga\Web\Widget\Tabs;
 use ipl\Html\Contract\Form;
+use ipl\Sql\Connection;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Compat\SearchControls;
 use ipl\Web\Control\LimitControl;
@@ -99,11 +101,12 @@ class SourcesController extends CompatController
 
     public function addAction(): void
     {
-        $form = (new SourceForm(Database::get()))
+        $form = (new SourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
             ->on(Form::ON_SUBMIT, function (SourceForm $form) {
-                $form->addSource();
-                Notification::success(sprintf(t('Added new source %s successfully'), $form->getSourceName()));
+                $source = $form->getSource();
+                Database::get()->transaction(fn (Connection $db) => (new SourceRepository($db))->create($source));
+                Notification::success(sprintf(t('Added new source %s successfully'), $source->name));
                 $this->switchToSingleColumnLayout();
             })
             ->handleRequest($this->getServerRequest());
--- a/application/forms/SourceForm.php
+++ b/application/forms/SourceForm.php
@@ -5,14 +5,12 @@
 
 namespace Icinga\Module\Notifications\Forms;
 
-use DateTime;
-use Icinga\Exception\Http\HttpNotFoundException;
+use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Model\Source;
 use ipl\Html\Attributes;
 use ipl\Html\HtmlDocument;
 use ipl\Html\HtmlElement;
 use ipl\Html\Text;
-use ipl\Sql\Connection;
 use ipl\Stdlib\Filter;
 use ipl\Validator\CallbackValidator;
 use ipl\Web\Common\CalloutType;
@@ -21,34 +19,19 @@ use ipl\Web\Compat\CompatForm;
 use ipl\Web\Url;
 use ipl\Web\Widget\ButtonLink;
 use ipl\Web\Widget\Callout;
-use RuntimeException;
 
 class SourceForm extends CompatForm
 {
     use CsrfCounterMeasure;
 
-    /** @var string|int The used password hash algorithm */
-    public const HASH_ALGORITHM = PASSWORD_BCRYPT;
-
-    /** @var string @var The generic source type */
+    /** @var string The generic source type */
     public const TYPE_GENERIC = 'generic';
 
     /** @var string The type for sources with an integration */
     private const TYPE_INTEGRATED = 'integrated';
 
-    /** @var Connection */
-    private Connection $db;
-
-    /** @var ?int */
-    private ?int $sourceId = null;
-
-    /** @var bool Whether the source is locked for editing */
-    private bool $isLocked = false;
-
-    public function __construct(Connection $db)
-    {
-        $this->db = $db;
-    }
+    /** @var ?Source The source to load */
+    private ?Source $source = null;
 
     protected function assemble(): void
     {
@@ -72,7 +55,7 @@ class SourceForm extends CompatForm
             [
                 'label'     => $this->translate('Source Name'),
                 'required'  => true,
-                'disabled'  => $this->isLocked
+                'disabled'  => $this->source?->locked
             ]
         );
         $this->addElement(
@@ -83,7 +66,7 @@ class SourceForm extends CompatForm
                 'required'  => true,
                 'label'     => $this->translate('Source Type'),
                 'value'     => self::TYPE_GENERIC,
-                'disabled'  => $this->isLocked,
+                'disabled'  => $this->source?->locked,
                 'class'     => 'autosubmit',
                 'options'   => [
                     self::TYPE_GENERIC    => $this->translate('Generic', 'notifications.source.type'),
@@ -112,7 +95,7 @@ class SourceForm extends CompatForm
                 [
                     'required'      => true,
                     'label'         => $this->translate('Source Identifier'),
-                    'disabled'  => $this->isLocked
+                    'disabled'  => $this->source?->locked
                 ]
             );
         }
@@ -138,14 +121,14 @@ class SourceForm extends CompatForm
             [
                 'required' => true,
                 'label' => $this->translate('Username'),
-                'disabled'  => $this->isLocked,
+                'disabled'  => $this->source?->locked,
                 'validators' => [new CallbackValidator(
                     function ($value, CallbackValidator $validator) {
                         // Username must be unique
-                        $source = Source::on($this->db)
+                        $source = Source::on(Database::get())
                             ->filter(Filter::equal('listener_username', $value));
-                        if ($this->sourceId !== null) {
-                            $source->filter(Filter::unequal('id', $this->sourceId));
+                        if ($this->source !== null) {
+                            $source->filter(Filter::unequal('id', $this->source->id));
                         }
 
                         if ($source->first() !== null) {
@@ -159,7 +142,7 @@ class SourceForm extends CompatForm
             ]
         );
 
-        if ($this->isLocked) {
+        if ($this->source?->locked) {
             $this->prependHtml(new Callout(
                 CalloutType::Info,
                 $this->translate('This source is managed by an integration, so changes can only be applied through it.')
@@ -172,8 +155,8 @@ class SourceForm extends CompatForm
             'password',
             'listener_password',
             [
-                'required'      => $this->sourceId === null,
-                'label'         => $this->sourceId !== null
+                'required'      => $this->source === null,
+                'label'         => $this->source !== null
                     ? $this->translate('New Password')
                     : $this->translate('Password'),
                 'autocomplete'  => 'new-password',
@@ -185,7 +168,7 @@ class SourceForm extends CompatForm
             'listener_password_dupe',
             [
                 'ignore'        => true,
-                'required'      => $this->sourceId === null,
+                'required'      => $this->source === null,
                 'label'         => $this->translate('Repeat Password'),
                 'autocomplete'  => 'new-password',
                 'validators'    => [new CallbackValidator(function (string $value, CallbackValidator $validator) {
@@ -204,19 +187,19 @@ class SourceForm extends CompatForm
             'submit',
             'save',
             [
-                'label' => $this->sourceId === null ?
+                'label' => $this->source === null ?
                     $this->translate('Add Source') :
                     $this->translate('Save Changes')
             ]
         );
 
-        if ($this->sourceId !== null) {
+        if ($this->source !== null) {
             $this->getElement('save')->prependWrapper(
                 (new HtmlDocument())
                     ->addHtml(
                         (new ButtonLink(
                             $this->translate('Delete'),
-                            Url::fromPath('notifications/source/delete/', ['id' => $this->sourceId])
+                            Url::fromPath('notifications/source/delete/', ['id' => $this->source->id])
                         ))->openInModal()
                     )
             );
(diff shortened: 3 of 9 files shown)


application/controllers/SourceController.php
nilmerg, line 1: In light of what I prepare currently, forms will only change a model's properties and provide the result in case of success. So all the write handling, transaction handling, etc, is still part of the controller. This should be already the case here and is then in-line with the previous implementation regarding responsibility.

`SourceForm` previously got the id of a source and was responsible to load it from the database. This _should_ change as it is required to avoid it having knowledge of the database itself. It's the last reason to keep this knowledge anyway, as the entire save handling is done elsewhere now.

Suggested changes:

`SourceForm::load(Source $source): static`
`SourceForm::get(): Source`

`SourceController::indexAction()` uses `SourceForm::get()` to pass its result to `Sources::save()`.
`SourceController::deleteAction()` only keeps the source for itself, as `DeleteSourceForm` doesn't get it anymore. (It only requires approval and only needs it for the actual deletion, which is now part of `Sources`) Again: `DeleteSourceForm` shouldn't need the model and the controller uses `Sources::delete()`.

library/Notifications/Model/Source.php
```diff
@@ -81,6 +82,7 @@ public function createBehaviors(Behaviors $behaviors): void
     {
         $behaviors->add(new MillisecondTimestamp(['changed_at']));
         $behaviors->add(new BoolCast(['deleted']));
+        $behaviors->add(new BoolCast(['locked']));
```
nilmerg: BoolCast accepts an array for a reason…

test/php/library/Notifications/Integrations/SourcesTest.php
```diff
@@ -1,20 +1,20 @@
 <?php
 
-// SPDX-FileCopyrightText: 2024 Icinga GmbH <https://icinga.com>
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
```
nilmerg: We don't raise this.

library/Notifications/Integrations/Sources.php
nilmerg, line 1: Please make this a `final class Source` instead. It should have a factory `get(string $username): self` with said username marked `readonly`. For all other properties respective setters are available (e.g. `setType(string $type): self`). Integrations are responsible to save any changes using `save()`.

nilmerg (changes requested):

test/php/library/Notifications/Integrations/SourcesTest.php
sukhwinder33445: It's actually a new file. GitHub is doing something wrong, as it only recognises the change to the filename, but in reality the namespace has also changed.

nilmerg:
please rebase

application/controllers/SourceController.php
```diff
@@ -45,20 +54,22 @@ public function indexAction(): void
 
     public function deleteAction(): void
     {
-        $sourceId = (int) $this->params->getRequired('id');
+        /** @var Source $source */
+        $source = Source::on(Database::get())
+            ->filter(Filter::equal('id', (int) $this->params->getRequired('id')))
+            ->first();
```
nilmerg: This is missing a check if it is found.

application/forms/SourceForm.php
```diff
@@ -184,139 +163,80 @@ function ($value, CallbackValidator $validator) {
...
         }
 
-        return [
-            'name' => $source->name,
-            'type' => $source->type,
-            'credentials' => [
-                'listener_username' => $source->listener_username
-            ]
-        ];
+        return $this->source->setProperties($data);
```
nilmerg: Just wondering, why do you use `setProperties` here? Assigning the properties one by one helps with static analysis after all, don't you think?

application/forms/SourceForm.php
```diff
@@ -130,14 +116,7 @@ protected function assemble(): void
...
                 'validators' => [new CallbackValidator(
                     function ($value, CallbackValidator $validator) {
-                        // Username must be unique
-                        $source = Source::on($this->db)
-                            ->filter(Filter::equal('listener_username', $value));
-                        if ($this->sourceId !== null) {
-                            $source->filter(Filter::unequal('id', $this->sourceId));
-                        }
-
-                        if ($source->first() !== null) {
```
nilmerg, line 151: Honestly, I'm not sure why you changed this? It's perfectly fine to query the database here, manually so to say. Using a class not meant for internal usage here is rather unnecessary and it changes how the validator functions anyway.

application/forms/SourceForm.php
```diff
@@ -184,139 +163,80 @@ function ($value, CallbackValidator $validator) {
...
-     * Load the source with given id
+     * Load the given source
      *
-     * @param int $id
+     * @param Source $source
      *
      * @return $this
      */
-    public function loadSource(int $id): static
+    public function loadSource(Source $source): static
```
nilmerg: I changed my mind now and think the standard setter/getter names fit better. i.e. `SourceForm::setSource()` and `SourceForm::getSource()`. That's how I do it as well in #494.

application/forms/SourceForm.php
```diff
@@ -184,139 +163,80 @@ function ($value, CallbackValidator $validator) {
...
+                            Url::fromPath('notifications/source/delete/', ['id' => $this->source->id])
                         ))->openInModal()
                     )
             );
         }
     }
 
     /**
-     * Load the source with given id
+     * Load the given source
```
nilmerg: *Set the schedule to populate the form with*

application/forms/SourceForm.php
```diff
@@ -184,139 +163,80 @@ function ($value, CallbackValidator $validator) {
...
-
-        $this->db->transaction(function (Connection $db) use ($source): void {
-            $db->insert('source', $source);
-        });
+        return $this->getValue('name');
     }
 
     /**
-     * Edit the source
+     * Get the source
```
nilmerg: *Get the schedule as configured by the user*

application/forms/SourceForm.php
```diff
@@ -184,139 +163,80 @@ function ($value, CallbackValidator $validator) {
...
     }
 
     /**
-     * Add the new source
+     * Get the source name
+     *
+     * @return string
      */
-    public function addSource(): void
+    public function getSourceName(): string
```
nilmerg: shouldn't this be obsolete now given the fact that you don't call it anymore in another case? (`schedule/index` vs `schedule/delete`)

library/Notifications/Model/Source.php
```diff
@@ -53,7 +53,8 @@ public function getColumns(): array
             'listener_username',
             'listener_password_hash',
             'changed_at',
-            'deleted'
+            'deleted',
+            'locked'
```
nilmerg, line 63: shouldn't this be in main already??

library/Notifications/Integrations/Source.php
nilmerg, line 1: Now I think it's time for the next step:

```php
final class \Icinga\Module\Notifications\Repository\SourceRepository
{
    public function __construct(Connection $db);
    public function find(int $id): ?Source;
    public function create(Source $source): void;
    public function update(Source $source): void;
    public function delete(Source $source): void;
}
```

That should be what `SourceController` and `Integrations\Source` use to manipulate the database.

Your todo in `Source::delete()` already hints at a similar thing: `RuleRepository::delete()`. But this isn't necessary to implement here, as I plan to do this in #494.

Eventually, `Integrations\Source` only has this interface:

```php
final class Source
{
    public static function get(string $username): self;
    public function setName(string $name): self;
    public function setType(string $type): self;
    public function setPassword(string $password): self;
    public function save(): void;
    public function delete(): void;
}
```

nilmerg (changes requested):

library/Notifications/Integrations/Source.php
```diff
@@ -0,0 +1,173 @@
...
+ * Utility class for managing sources in the database
+ */
+final class Source
+{
+    /** @var string The used password hash algorithm */
+    public const HASH_ALGORITHM = PASSWORD_BCRYPT;
+
+    protected SourceModel $source;
+
+    protected readonly string $username;
```
nilmerg: Since you now have the model internally, there's no need for this property anymore.

application/forms/SourceForm.php
sukhwinder33445: This was changed to remove the db dependency. I have reverted the change.

library/Notifications/Model/Source.php
sukhwinder33445: Rebased the branch and removed it.

library/Notifications/Integrations/Source.php
sukhwinder33445: The `SourceRepository` class has been added. As this class is based on the new Writer, it only requires the `save()` method intead of `create()` and `update()`.

library/Notifications/Integrations/Source.php
```diff
@@ -0,0 +1,145 @@
...
+     * @param string $password
+     *
+     * @return string
+     */
+    public static function hashPassword(string $password): string
+    {
+        // Not using PASSWORD_DEFAULT, as the used algorithm should
+        // be kept in sync with what the daemon understands
+        return password_hash($password, self::HASH_ALGORITHM);
+    }
```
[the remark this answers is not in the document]
sukhwinder33445: false only PRIOR PHP 8.0.

library/Notifications/Integrations/Source.php
nilmerg: That's interesting. Taking a look at how `SourceForm` differentiates here leads to these two differences:

* Optional changes (`listener_password`)
* Lock status assertion

It also shows how password hashing is applied.

--

Optional changes remain being tracked in the form regardless. Lock status assertion on the other hand is done by the controller upon submit and that's also fine as it's exclusive to the Notifications UI. But password hashing applies to all consumers (Integrations and Notifications alike), and should be the duty of the repository, which causes it to differentiate between *create*/*update* as in the *update* case it's not set on the model.

This should also solve the weird thing that `SourceForm` depends on `Integrations\Source::hashPassword()`. There should be nothing but tests, that requires this class.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -0,0 +1,102 @@
...
+     * @return void
+     */
+    public function delete(Source $source): void
+    {
+        if ($source->isNew()) {
+            // Source has not yet persisted, nothing to do.
+            return;
+        }
+
+        $this->db->transaction(function () use ($source): void {
```
nilmerg: Transaction handling is the *service's* job, in our case the controller.

application/controllers/SourceController.php
```diff
@@ -24,41 +27,62 @@ public function init(): void
 
     public function indexAction(): void
     {
-        $sourceId = (int) $this->params->getRequired('id');
+        /** @var ?Source $source */
+        $source = Source::on(Database::get())
+            ->filter(Filter::equal('id', (int) $this->params->getRequired('id')))
+            ->first();
```
nilmerg: Use `SourceRepository::find`

application/controllers/SourceController.php
```diff
@@ -24,41 +27,62 @@ public function init(): void
...
+            ->setSource($source)
             ->on(Form::ON_SUBMIT, function (SourceForm $form): never {
-                Database::get()->transaction(fn () => $form->editSource());
+                $source = $form->getSource();
+
+                if ($source->locked) {
+                    throw new RuntimeException('Source is locked');
+                }
+
+                (new SourceRepository(Database::get()))->save($source);
```
nilmerg: Use a transaction here

application/controllers/SourceController.php
```diff
@@ -24,41 +27,62 @@ public function init(): void
...
+        }
 
         $form = (new DeleteSourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->loadSource($sourceId)
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUBMIT, function (DeleteSourceForm $form): never {
-                Database::get()->transaction(fn (Connection $db) => $form->removeSource($db));
+            ->on(Form::ON_SUBMIT, function () use ($source): never {
+                (new SourceRepository(Database::get()))->delete($source);
```
nilmerg: Use a transaction here

application/controllers/SourceController.php
```diff
@@ -24,41 +27,62 @@ public function init(): void
...
-        $this->addTitleTab(sprintf($this->translate('Source: %s'), $form->getSourceName()));
+        $this->addTitleTab(sprintf($this->translate('Source: %s'), $source->name));
         $this->addContent($form);
     }
 
     public function deleteAction(): void
     {
-        $sourceId = (int) $this->params->getRequired('id');
+        /** @var ?Source $source */
+        $source = Source::on(Database::get())
```
nilmerg: Use `SourceRepository::find`

nilmerg (changes requested):

application/controllers/SourceController.php
```diff
@@ -8,57 +8,75 @@
...
+            throw new RuntimeException('Source is locked');
+        }
 
         $form = (new DeleteSourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->loadSource($sourceId)
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUBMIT, function (DeleteSourceForm $form): never {
-                Database::get()->transaction(fn (Connection $db) => $form->removeSource($db));
+            ->on(Form::ON_SUBMIT, function (): never {
```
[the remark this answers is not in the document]
sukhwinder33445: No, `switchToSingleColumnLayout()` redirects.

application/forms/SourceForm.php
```diff
@@ -204,145 +190,69 @@ function ($value, CallbackValidator $validator) {
...
-        if ($source === null) {
-            throw new HttpNotFoundException($this->translate('Source not found'));
+        $this->source->name = $this->getValue('name');
+        $this->source->type = $this->getValue('type', self::TYPE_GENERIC);
+        $this->source->listener_username = $this->getElement('credentials')->getValue('listener_username');
+        $pwd = $this->getElement('credentials')->getValue('listener_password');
+        if ($pwd) {
+            // The repository hashes the password on save
+            $this->source->listener_password = $pwd;
         }
```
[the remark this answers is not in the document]
sukhwinder33445: Typo fixed.

application/controllers/SourceController.php
```diff
@@ -8,57 +8,75 @@
...
+            ->setSource($this->source)
             ->on(Form::ON_SUBMIT, function (SourceForm $form): never {
-                Database::get()->transaction(fn () => $form->editSource());
+                $source = $form->getSource();
+
+                if ($source->locked) {
+                    throw new RuntimeException('Source is locked');
+                }
+
+                Database::get()->transaction(fn () => (new SourceRepository(Database::get()))->update($source));
```
nilmerg: The callback gets the connection passed, use this.

application/controllers/SourceController.php
```diff
@@ -8,57 +8,75 @@
...
+        }
 
         $form = (new DeleteSourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->loadSource($sourceId)
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUBMIT, function (DeleteSourceForm $form): never {
-                Database::get()->transaction(fn (Connection $db) => $form->removeSource($db));
+            ->on(Form::ON_SUBMIT, function (): never {
+                Database::get()->transaction(fn () => (new SourceRepository(Database::get()))->delete($this->source));
```
nilmerg: The callback gets the connection passed, use this.

application/controllers/SourcesController.php
```diff
@@ -99,11 +100,12 @@ public function indexAction(): void
...
     public function addAction(): void
     {
-        $form = (new SourceForm(Database::get()))
+        $form = (new SourceForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
             ->on(Form::ON_SUBMIT, function (SourceForm $form) {
-                $form->addSource();
-                Notification::success(sprintf(t('Added new source %s successfully'), $form->getSourceName()));
+                $source = $form->getSource();
+                Database::get()->transaction(fn () => (new SourceRepository(Database::get()))->create($source));
```
nilmerg: The callback gets the connection passed, use this.

application/forms/SourceForm.php
```diff
@@ -21,35 +19,23 @@
...
-    /** @var Connection */
-    private Connection $db;
-
-    /** @var ?int */
-    private ?int $sourceId = null;
+    /** @var ?Source The source to load */
+    private ?Source $source = null;
 
     /** @var bool Whether the source is locked for editing */
     private bool $isLocked = false;
```
nilmerg: Now that the entire source model is available to the instance, I think accessing `$this->source?->locked` is sufficient.

application/forms/SourceForm.php
```diff
@@ -204,145 +190,69 @@ function ($value, CallbackValidator $validator) {
...
-
-        if ($source === null) {
-            throw new HttpNotFoundException($this->translate('Source not found'));
+        $this->source->name = $this->getValue('name');
+        $this->source->type = $this->getValue('type', self::TYPE_GENERIC);
+        $this->source->listener_username = $this->getElement('credentials')->getValue('listener_username');
+        $pwd = $this->getElement('credentials')->getValue('listener_password');
+        if ($pwd) {
+            // The repository hashes and assign the password to `listener_password_hash` column on save
+            $this->source->listener_password = $pwd;
```
nilmerg, line 248: It **is** possible to assign any property to models, but I'm inclined to suggest to add it to the model's doc block at least with a comment explaining what the comment here says.

library/Notifications/Integrations/Source.php
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * If the source does not exist in the database, a new locked source is created.
+     * To store the source in the database, {@see self::save()} must be called.
+     *
+     * @param string $username
+     * @param ?Connection $connection The database connection to use.
+     *
+     * @return self
+     */
+    public static function get(string $username, ?Connection $connection = null): self
```
nilmerg: This is a case where I think *testable* doesn't really apply as reason for having an optional argument which isn't of any use other than tests. Please mark the constructor public instead. The test regarding the locked property can be rewritten to expect `save()` (which sets it) to issue an `$db->insert` with said column set to true. To me it doesn't really make a difference whether the source is locked from the very beginning or only after an integration wants to save it.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -0,0 +1,137 @@
...
+     * Create a new source
+     *
+     * @param Source $source
+     *
+     * @return void
+     */
+    public function create(Source $source): void
+    {
+        if (isset($source->listener_password)) {
+            $source->listener_password_hash = self::hashPassword($source->listener_password);
```
nilmerg: Unset `listener_password` afterwards please, even as part of a `try…finally` block, just in case something breaks and so we make sure the cleartext password isn't left in memory.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @return void
+     */
+    public function update(Source $source): void
+    {
+        if (isset($source->listener_password)) {
+            $source->listener_password_hash = self::hashPassword($source->listener_password);
+        }
+
+        $this->em->save($source);
```
nilmerg: It's perfectly fine to call `create` here or add `upsert` as internal method ;)

nilmerg (changes requested):

application/forms/SourceForm.php
```diff
@@ -204,145 +187,68 @@ function ($value, CallbackValidator $validator) {
...
```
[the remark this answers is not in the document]
sukhwinder33445: Its redudant now, removed.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -0,0 +1,150 @@
...
+        foreach ($source->rule->columns('id') as $rule) {
+            //TODO: add repository for rule
+            EventRuleConfigForm::removeRule($this->db, $rule);
+        }
+
+        $source->deleted = true;
+        $source->listener_username = null;
+
+        $this->em->save($source);
+    }
```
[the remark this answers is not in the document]
sukhwinder33445: `Integrations\Source` has the transaction now.

library/Notifications/Model/Source.php
```diff
@@ -24,6 +24,10 @@
  * @property string $name The user-defined name
  * @property ?string $listener_username The username for HTTP authentication
  * @property ?string $listener_password_hash
+ * @property ?string $listener_password Helper property to temporarily store the raw password (not persisted to the
+ *                                      database). The repository hashes it and assigns the result to
+ *                                      {@see self::$listener_password_hash} before saving the source to the database.
+ *                                      {@see SourceRepository::create()} and {@see SourceRepository::update()}
```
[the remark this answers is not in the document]
sukhwinder33445: It works.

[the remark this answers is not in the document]
nilmerg: What works? I don't see why that comment is wrong.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @param Source $source The source to create or update.
+     */
+    private function upsert(Source $source): void
+    {
+        if (isset($source->listener_password)) {
+            try {
+                $source->listener_password_hash = self::hashPassword($source->listener_password);
+            } finally {
+                unset($source->listener_password);
```
nilmerg, line 141: this isn't covered by the tests yet

nilmerg (changes requested):

library/Notifications/Model/Source.php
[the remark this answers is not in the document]
sukhwinder33445: I meant that it works, the IDE can resolve the references correctly. But yes, static analysers can fail, so adding the FQCN is safer and helps them too.

library/Notifications/Model/Source.php
```diff
@@ -24,6 +24,12 @@
  * @property string $name The user-defined name
  * @property ?string $listener_username The username for HTTP authentication
  * @property ?string $listener_password_hash
+ * @property ?string $listener_password Helper property to temporarily store the raw password (not persisted to the
+ *                                      database). The repository hashes it and assigns the result to
+ *                                      {@see self::$listener_password_hash} before saving the source to the database.
+ *                                      See the usage in
+ *                                      {@see \Icinga\Module\Notifications\Repository\SourceRepository::create()} and
+ *                                      {@see \Icinga\Module\Notifications\Repository\SourceRepository::update()}
```
nilmerg: A link to upsert suffices now and please `use` the class normally.

nilmerg (changes requested):

nilmerg (approved):

library/Notifications/Integrations/Source.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Save the source
+     *
+     * This makes sure the source is locked.
+     *
+     * @return void
+     */
+    public function save(): void
+    {
+        $this->source->locked = true;
```
nilmerg, line 127: Just noticed this, `save()` should throw an error in case the source is new and no name or type is set. I don't want to leave this up to the repository or database to handle.

nilmerg (changes requested):

library/Notifications/Integrations/Source.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Set the name of the source
+     *
+     * @param string $name
+     *
+     * @return $this
+     */
+    public function setName(string $name): self
```
nilmerg, line 77: Last comment, promised. I feel that there should be a `getName` as well. Integrations are not required to define their name once as it's always the same. Take different Icinga environments for kubernetes clusters for example, they all use the same type, but the name is supposedly different and a `getName` allows to display it again or make it changeable for the user. The user may have chosen a name to begin with and by not being able to *get* it later on it's not possible to visualize for the integration.

nilmerg (changes requested):

nilmerg (dismissed):

sukhwinder33445:
Only the last commit contains non-approved changes after the rebase.

application/forms/SourceForm.php
```diff
@@ -21,34 +19,19 @@
...
-use RuntimeException;
 
 class SourceForm extends CompatForm
 {
     use CsrfCounterMeasure;
 
-    /** @var string|int The used password hash algorithm */
-    public const HASH_ALGORITHM = PASSWORD_BCRYPT;
-
     /** @var string @var The generic source type */
```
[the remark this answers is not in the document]
sukhwinder33445: Done.

nilmerg (dismissed):

nilmerg (approved):

PR #486, opened 2026-06-02
Incident: Introduce column `mute_reason`

### LINKED ISSUE #485: Read muted state from incidents rather than objects
https://github.com/Icinga/icinga-notifications/pull/431 will move the `muted` flag state tracking into incidents instead of objects, so Icinga Notifications Web needs the corresponding changes as well.

nilmerg:
This was being tracked on objects before, now the mute state is tracked on incidents instead.

fixes #485


Jan-Schuppik (approved):
The failed unit-tests are not related to the changes this PR, so it's fine here

Issue #457, opened 2026-03-25
Open Escalations in Modal


flourish86:
Configure Escalations in modal dialogue

Note that the video shows a rule type distinction and a source indicator, which both are not subject of this task.

<img width="1280" height="1020" alt="Image" src="https://github.com/user-attachments/assets/17d3266a-59b9-4f9f-99e2-67e2ea8b201d" />

nilmerg:
The overview should rather describe the conditions in a less technical way.

*If the incident age reached 1 hour*
*If the severity exceeds warning and the incident age reached 2 days*

nilmerg:
This will also need to change the workflow to create or modify event rules. Once this is done, modification of an event rule is no longer possible outside a modal. The overview then is read-only and by creating a new rule, it exists at first without any escalation.

nilmerg:
#494 won't add duplication right now. It's a task of this issue here since it requires the planned interaction flow changes.

PR #494, opened 2026-06-10
Allow to duplicate configuration

### LINKED ISSUE #158: Configuration: Duplicating entries and event rules must be possible
Add duplicate button in entry form (mentioned in #149) and event rule form. This makes it possible to easily create similar entries and event rules.

nilmerg:
resolves #158

## Overview

Duplicating a configuration object means re-running the exact "persist this object and its relations" logic that previously lived inline in the forms. Rather than copy-pasting that logic, this PR first extracts persistence out of the forms into a small, reusable and testable layer, and then implements the duplication feature on top of it.

## The persistence layer

- **DTOs** (`library/Notifications/Form/Data/`) — plain, mostly-`readonly` value objects (`Contact`, `ContactGroup`, `Schedule`, `Rotation`, `EscalationRule`, `Escalation`, `EscalationRecipient`, `Source`). Forms no longer touch the database directly: they collect user input into a DTO and hand it to a repository.
- **Repositories** (`library/Notifications/Repository/`) — one per aggregate (`Contact`, `ContactGroup`, `Schedule`, `Rotation`, `Escalation`, `EscalationRule`, `Source`). They mediate between forms/controllers and the database and now hold the business logic that used to be spread across the forms: differential relation updates, priority (re)ordering, freeing unique slots on deletion, cascading deletes to dependent rotations/escalations, etc. They expose `find`/`create`/`update`/`delete`, plus `duplicate`/`move`/`wipe` where applicable.
- **`EntityManager` + change tracking** — repositories delegate the actual writes to the ORM-based `EntityManager`, which diffs a model against its loaded state, writes only what changed, and cascades related models in the correct order within one transaction. `Common\Model` gained change tracking and a `restore()` for soft-deleted rows. Repositories deliberately do **not** open transactions themselves — the calling controller owns the transaction, so multiple repository calls compose atomically.

## The feature

With that layer in place, the following can now be duplicated:

- **Schedules** — in full, including all of their rotations.
- **Rotations** — individually.
- **Contact groups**.

## Along the way

- Added the ORM relations/columns the new logic relies on (e.g. `Contact`/`Contactgroup` ↔ `rotation`/`rule_escalation`, missing `external_uuid` declarations).
- Switched the V1 API deletion paths that previously duplicated rotation/rule teardown logic over to the new repositories.

## Not included: event rule duplication

Event rule duplication is intentionally left out for now. #457 will change the interaction flow of the event rule configuration, which makes implementing duplication against the current flow unreasonable — once that change has landed, wiring duplication for event rules is a trivial follow-up.

## Testing

The repositories are covered by tests that run against real databases, once for **MySQL** and once for **PostgreSQL** (`DbTestBackends` / the `sharedDatabases` provider), exercising find/create/update/delete along with duplication, priority reshuffling and the delete-time cascade behavior. The rotation recurrence-rule generation moved into the `Rotation` DTO and keeps its dedicated unit tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


library/Notifications/Repository/RotationRepository.php
```diff
@@ -0,0 +1,480 @@
...
+            ->first()?->setNew(false);
+        if ($model === null) {
+            throw new RuntimeException('Cannot delete a rotation that does not exist in the database');
+        }
+
+        $entityManager = new EntityManager($this->db);
+        $freedPriority = $model->priority;
+
+        $model->timeperiod->timeperiod_entry = [];
+        $model->timeperiod->setNew(false)->delete();
```
BastianLedererIcinga: The `Timeperiod` must be made mutable before writing to its relation, currently deleting a rotation leaves orphan `timeperid_entry` rows.
```suggestion
        $model->timeperiod->setNew(false)->delete();
        $model->timeperiod->timeperiod_entry = [];
```

BastianLedererIcinga (dismissed):
I found no more problems while testing, the state of the databse and the runtime updates logged by the daemon match what I configure in the UI.

PR #500, opened 2026-06-17
`EscalationCondition`: Remove not required incident age operators

### LINKED ISSUE #476: Restrict incident age conditions to chronologically senseful matches
Escalations are evaluated upon each change to an incident. Conditions allow to restrict this based on an incident's age. But while the age of course changes, it may do so automatically in the sense that the conditions do not all match unless a given age is reached. This is why the daemon tracks such conditions irrespective of external triggers.

This is why configuring a condition to match incidents that are *younger* than (`<`, `<=`) does not make a difference to not match based on that. Ranges (`>=1h and <=2h`) on the other hand are only possible with these operators and are not configurable anymore without them. Though, we cannot identify a use case right now that requires ranges. Once we did, we may re-introduce them.

sukhwinder33445:
This PR removes unnecessary operators for incident age, keeping only the `>` operator.

resolves #476

### DIFF
--- a/application/forms/EventRuleConfigElements/EscalationCondition.php
+++ b/application/forms/EventRuleConfigElements/EscalationCondition.php
@@ -118,19 +118,27 @@ class EscalationCondition extends FieldsetElement
             )
         ]));
 
-        $this->addElement('select', 'operator', [
-            'required' => true,
-            'options' => [
+        $selectedColumn = $this->getPopulatedValue('column');
+        if ($selectedColumn === 'incident_age') {
+            $operators = ['>' => '>'];
+        } else {
+            $operators = [
                 '='  => '=',
                 '>'  => '>',
                 '>=' => '>=',
                 '<'  => '<',
                 '<=' => '<=',
                 '!=' => '!='
-            ]
+            ];
+        }
+
+        $this->addElement('select', 'operator', [
+            'required' => true,
+            'options' => $operators,
+            'disabled' => $selectedColumn === 'incident_age'
         ]);
 
-        if ($this->getPopulatedValue('column') === 'incident_severity') {
+        if ($selectedColumn === 'incident_severity') {
             $this->addElement('select', 'severity', [
                 'required' => true,
                 'options' => [
@@ -145,7 +153,7 @@ class EscalationCondition extends FieldsetElement
                     'emerg' => $this->translate('Emergency', 'notification.severity')
                 ]
             ]);
-        } elseif ($this->getPopulatedValue('column') === 'incident_age') {
+        } elseif ($selectedColumn === 'incident_age') {
             $noOf = $this->createElement('number', 'no_of', [
                 'required' => true,
                 'min' => 1,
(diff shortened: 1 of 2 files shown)


application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -118,19 +118,26 @@ protected function assemble(): void
...
+            '<'  => '<',
+            '<=' => '<=',
+            '!=' => '!='
+        ];
+
+        $selectedColumn = $this->getPopulatedValue('column');
+        if ($selectedColumn === 'incident_age') {
+            unset($operators['<'], $operators['<=']);
+        }
+
```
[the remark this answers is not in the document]
sukhwinder33445: These are being removed as there is no specific use case for them.

application/forms/EventRuleConfigElements/EscalationCondition.php
nilmerg, line 1: Just discussed this shortly with @yhabteab and now I think even `=` and `>=` do not make much sense for `incident_age`. The daemon cannot guarantee that the condition is evaluated at the exact time and thus an exact match is not useful.

The dropdown then only has `>` left, but that's fine. Though, it needs to be disabled to not fool users into thinking that there's another missing option. But let's keep the element in order to still visualize the actual condition.

nilmerg (changes requested):

application/forms/EventRuleConfigElements/EscalationCondition.php
nilmerg: And add a pull request description

yhabteab: > The daemon cannot guarantee that the condition is evaluated at the exact time and thus an exact match is not useful.

It cannot guarantee that it will always evaluate the condition at the exact time within a second boundry but it's still possible, so why ban the `>=` too? If the `=` operator is gone, then at least give the users the possibility to express _either at exact 10m or later_ conditions.

nilmerg: It over complicates the configuration with only very little advantage. The goal of this change is to reduce the available options to sensible cases, even if only one is left. It's about at most a full second delay this will introduce, that's fine to me.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -118,19 +118,26 @@ protected function assemble(): void
...
         ]));
 
-        $this->addElement('select', 'operator', [
-            'required' => true,
-            'options' => [
+        $selectedColumn = $this->getPopulatedValue('column');
+        if ($selectedColumn === 'incident_age') {
+            $operators = ['>' => '>'];
+        } else {
+            $operators = [
```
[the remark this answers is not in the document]
sukhwinder33445: Fixed the tests.

application/forms/EventRuleConfigElements/EscalationCondition.php
```diff
@@ -118,19 +118,26 @@ protected function assemble(): void
...
                 '<'  => '<',
                 '<=' => '<=',
                 '!=' => '!='
-            ]
+            ];
+        }
+
+        $this->addElement('select', 'operator', [
+            'required' => true,
+            'options' => $operators
```
nilmerg: Why is not disabled?

nilmerg (approved):

PR #502, opened 2026-06-26
Add `filter_name` text field to the event rule `SearchEditor`

### LINKED ISSUE #459: Add Filter Name Field
https://github.com/user-attachments/assets/d8e6e721-4887-4f57-ab21-49e662cfa572

sukhwinder33445:
This PR adds `filter_name` an optional text field to the event rule `SearchEditor`, when set, displayed as the button label — making filters self-describing without opening the editor.

**Without custom name**
image

**With custom name**
image
image

resolves #459

### DIFF
--- a/application/forms/EventRuleConfigForm.php
+++ b/application/forms/EventRuleConfigForm.php
@@ -13,7 +13,9 @@ use Icinga\Module\Notifications\Model\Rule;
 use Icinga\Module\Notifications\Model\RuleEscalation;
 use ipl\Html\Attributes;
 use ipl\Html\FormElement\SubmitButtonElement;
+use ipl\Html\HtmlDocument;
 use ipl\Html\HtmlElement;
+use ipl\Html\Text;
 use ipl\Html\ValidHtml;
 use ipl\I18n\Translation;
 use ipl\Sql\Connection;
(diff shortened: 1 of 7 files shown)


library/Notifications/Util/RuleSerializer.php
```diff
@@ -28,11 +31,13 @@ class RuleSerializer
      *
      * @param Filter\Rule $filter
      * @param array<string, string[]> $jsonPaths JSON paths keyed by column name
+     * @param ?string $filterName The name of the filter
      */
-    public function __construct(Filter\Rule $filter, array $jsonPaths)
+    public function __construct(Filter\Rule $filter, array $jsonPaths, ?string $filterName)
```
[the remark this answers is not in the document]
sukhwinder33445: Fixed and added tests.

application/forms/EventRuleConfigForm.php
```diff
@@ -131,10 +133,20 @@ protected function createObjectFilterControls(): ValidHtml
         $this->registerElement($hiddenInput);
 
         if ($hiddenInput->hasValue()) {
-            $label = new Icon('filter');
+            $parsedFilter = json_decode($hiddenInput->getValue(), true, flags: JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
sukhwinder33445: The `EventRuleController` already throws for this case. Here it should not be corrupted, but if it is, the built-in throw is sufficient.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -45,6 +50,7 @@ public function __construct(Filter\Rule $filter, array $jsonPaths)
     public function getJson(): ?string
     {
         $result = [
+            'filter_name' => $this->filterName,
```
nilmerg: Unless there's a need to store empty values, I'd rather avoid this to reduce noise in the serialized value. And since the consuming path uses `empty` it isn't necessary, so please don't store null here.

public/css/detail/event-rule-detail.less
```diff
@@ -261,3 +269,20 @@
...
+  .control-label-group {
+    line-height: 1.1em;
+    padding: .5625em .5625em .5625em 0;
+  }
+
+  input {
+    flex-grow: 1;
+    .rounded-corners();
+  }
+}
```
nilmerg: It's flex so please use native centering techniques. For design, such as rounded-corners, apply class `icinga-controls`.

```suggestion
.event-rule-filter > .filter-name {
  align-items: baseline;
  justify-content: center;
  gap: .5625em;
}
```

public/css/detail/event-rule-detail.less
```diff
@@ -105,10 +105,18 @@
...
       flex: 0;
     }
+
+    .filter-button {
+      flex-direction: column;
+
+      .name {
+        overflow: hidden;
+        .line-clamp(3);
+      }
```
nilmerg: Not sure I understand what this is supposed to do, since it looks this way for long content:

Image

nilmerg (changes requested):

public/css/detail/event-rule-detail.less
sukhwinder33445: Cannot use `icinga-controls` as it overrides the field's bg color as well. Applying `rounded-corners()` manually is the better approach, I think.

public/css/detail/event-rule-detail.less
sukhwinder33445: Added missing `word-break: break-word;`

nilmerg: Reduce the line clamp to two and the padding to a value that's less noticeable, it's currently way too much. Also, please reduce the max width somewhat.

nilmerg (changes requested):

public/css/detail/event-rule-detail.less
sukhwinder33445: Done.

nilmerg (approved):
Made some additional changes myself. Have a look at the commits for details.

PR #496, opened 2026-06-12
`RotationConfigForm`: Fix handoff interval 0

BastianLedererIcinga:
resolve #438 

The stacktrace was caused by the `Rotation Start` autosubmitting in order to calculate the first handoff, which happens before the handoff interval is validated.

To prevent this, the first handoff is only calculated for valid `$options` and a missing `GreaterThanValidator` for the interval of the multi day mode is added, the other 2 modes already use one.

PR #508, opened 2026-07-10
Fix invalid csrf token in `RotationConfigForm`

BastianLedererIcinga:
https://github.com/Icinga/icinga-notifications-web/pull/496 introduced validation of the entire form when calculating the first handoff, but the validation of the CSRF element fails when it is run there.

To fix this only the `$options` element is validated, since it should cover everything the first handoff needs.


nilmerg (approved):

nilmerg:
@cla-bot check

PR #522, opened 2026-07-23
Properly handle soft-deleted rows

nilmerg:
This introduces a proper way to hide deleted entities automatically. The previous work-around only affected a query's base table and thus didn't restrict joins the same way. Now, joins are covered as well, as are any sub-queries made by the ORM. And that is the only behavior related change: Queries made by ipl-sql are not automatically filtered anymore.

Related changes are introduced as well:
* Added `LEFT` join types to relations that were always meant to have it
* Introduced various guards when accessing relation properties as now they may be `NULL`

requires https://github.com/Icinga/ipl-orm/pull/165

### DIFF
--- a/library/Notifications/Model/Source.php
+++ b/library/Notifications/Model/Source.php
@@ -96,8 +96,10 @@ class Source extends Model
 
     public function createRelations(Relations $relations): void
     {
-        $relations->hasMany('object', Objects::class);
-        $relations->hasMany('rule', Rule::class);
+        $relations->hasMany('object', Objects::class)
+            ->setJoinType('LEFT');
+        $relations->hasMany('rule', Rule::class)
+            ->setJoinType('LEFT');
     }
 
     /**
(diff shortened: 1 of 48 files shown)


library/Notifications/Model/Source.php
```diff
@@ -131,4 +134,9 @@ public function getIcon(): Icon
 
         return $icon;
     }
+
+    public function createVisibilityFilter(Filter\Chain $filter): void
+    {
+        $filter->add(Filter::equal('deleted', 'n'));
```
BastianLedererIcinga, line 141: With this filter the incidents overview shows a deprecation warning if any of the displayed incidents belongs to a deleted source.
`Deprecated: Using null as an array offset is deprecated, use an empty string instead in /usr/share/icingaweb-modules/notifications/library/Notifications/Hook/ObjectsRendererHook.php on line 97`

The incident detail shows a similar warning from line 208 of `ObjectsRendererHook` for the same reason.

BastianLedererIcinga (changes requested):
Apart from the deprecation warning everything worked fine while testing and the changes look good.

library/Notifications/Model/Source.php
nilmerg: Done.

BastianLedererIcinga (dismissed):

nilmerg:
the latest two commits contain new changes

BastianLedererIcinga (dismissed):

BastianLedererIcinga (approved):

PR #523, opened 2026-07-24
Fix creation of orphan links by the `EntityManager`

BastianLedererIcinga:
While testing https://github.com/Icinga/icinga-notifications-web/pull/494 I noticed the following behavior:
Deleting the last member of an escalation deletes the escalation but leaves a `rule_escalation_recipient` link with `deleted` = `n`.
If the contact was not the last member of the escalation the link was properly soft deleted.

This is the snippet from `ContactRepository::delete()`
```
foreach ($model->rule_escalation as $escalation) {
    $model->rule_escalation->detach($escalation);

    $otherRecipient = $escalation->rule_escalation_recipient
        ->query()
        ->columns([new Expression('1')])
        ->filter(Filter::all(
            Filter::unequal('contact_id', $id),
            Filter::equal('deleted', 'n')
        ))
        ->first();
    if ($otherRecipient === null) {
        (new EscalationRepository($this->db))->delete($escalation->id); // The rule_escalation_recipient row is deleted here already
    }
}

(new EntityManager($this->db))->save($model);
```

Iterating the `rule_escalation` relation materializes its `$base`, so the detached escalation is included in `Collection::getMembersToSave()`.
`EntityManager::saveManyToMany()` therefore passes it to `reconcileJunction()` both in `$attach` and `$detach`.
When `reconcileSoftDeleteJunction()` queries the existing `rule_escalation_recipient` rows, the link is already soft deleted by the `EscalationRepository`, so it is not added to the `$toRemove` set.
But because the same escalation is also in the `$desired` set, the link is revived.

The problem is that the desired set of links is derived from `getMembersToSave()` instead of the explicit attachments.

So `getMembersToSave()` now only returns members that were attached or actually modified in place for collections in merge mode. Their changes are persisted in `saveManyToMany`.

`reconcileJunction()` is called with only the explicit attachments of a `Collection` exposed by `Collection::getAttachments()`, so it no longer writes to links that were never meant to be touched.

### DIFF
--- a/library/Notifications/Common/Collection.php
+++ b/library/Notifications/Common/Collection.php
@@ -189,7 +189,11 @@ class Collection implements IteratorAggregate, Countable
             return array_values($this->attach);
         }
 
-        return array_values(array_merge($this->base ?? [], $this->attach));
+        $members = array_filter($this->base ?? [], function ($model) {
+            return $model->isModified();
+        });
+
+        return array_values(array_merge($members, $this->attach));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sukhwinder33445 (approved):
LGTM

PR #527, opened 2026-07-29
Let the channel configuration also rely on the `EntityManager`

### LINKED ISSUE #525: Let the channel configuration also rely on the `EntityManager`
This includes an introduction of a repository and a suitable DTO, plus accompanying changes to the form and its controllers.

~~Eventually, the `Database` trait won't need to listen on assembled insert/update statements anymore.~~

nilmerg:
resolves #525

## Overview

Channels were the last configuration object that still assembled their own `INSERT`/`UPDATE` statements inside the form. They now use the same persistence layer as everything else: a DTO the form produces, and a repository that maps it onto the model and hands it to the `EntityManager`.

- **`Form\Data\Channel`** — plain readonly DTO (`id`, `name`, `type`, `config`).
- **`ChannelRepository`** — `find()`/`create()`/`update()`/`delete()`, with a private `upsert()` shared by create and update. It owns what the form did before: generating the `external_uuid`, encoding the config (`Json::encode()` with `JSON_FORCE_OBJECT`) and soft-deleting. It does not open transactions itself, the controller does.
- **`ChannelForm`** — `loadChannel()`, `addChannel()`, `editChannel()`, `removeChannel()` and `getChannelName()` are gone in favor of `setChannel()`/`getChannel()`. The channel id is now a hidden element, as in `SourceForm` and `ContactForm`. The only query left in the form is the one that determines whether the channel is still referenced and the delete button hence has to be disabled. The form also applies the default element decorators now, like the other forms already do, which is why the delete button is prepended by means of `prependWrapper()`.
- **`ChannelController`/`ChannelsController`** — load and persist by means of the repository, within a transaction they own. Both actions now follow the structure of the other configuration controllers: the form is populated in `ON_REQUEST`, i.e. only upon its first request, and re-rendered as a part in `ON_SENT` otherwise. `ChannelForm::onError()` emits `ON_SENT` as well, so that a failed validation still updates the form.

Note that no longer bumping `changed_at` if nothing changed, which the form did by diffing against the stored values, is now the `EntityManager`'s doing, as it only writes modified properties.

`Common\Database` still listens on assembled `INSERT`/`UPDATE` statements, as the V1 API and the incident widgets continue to use them.

## Testing

`ChannelRepositoryTest` covers find/create/update/delete against real MySQL and PostgreSQL databases, including the soft delete and that an unchanged channel isn't touched at all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/application/controllers/ChannelController.php
+++ b/application/controllers/ChannelController.php
@@ -7,8 +7,10 @@ namespace Icinga\Module\Notifications\Controllers;
 
 use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Forms\ChannelForm;
+use Icinga\Module\Notifications\Repository\ChannelRepository;
 use Icinga\Web\Notification;
 use ipl\Html\Contract\Form;
+use ipl\Sql\Connection;
 use ipl\Web\Compat\CompatController;
 
 class ChannelController extends CompatController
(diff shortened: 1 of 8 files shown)


nilmerg (commented):
Looks and works fine.

PR #535, opened 2026-08-03
Let integrations check for a contact and rework the recipient shape

### LINKED ISSUE #529: Provide integrations the option to check whether a contact with a given username is available
Integrations are now expected to be the primary frontend for their own *incidents* and thus provide a way to manage them and subscribe to them. But not every user is expected to be granted to use this capability. It is also not guaranteed that either of these actions is actually possible, since there may be no contact available that matches the currently logged-in user.

So my suggestion is to introduce:
- `\Icinga\Module\Notifications\Integrations\Incidents::canManage(\Icinga\User $user)`
- `\Icinga\Module\Notifications\Integrations\Incidents::canSubscribe(\Icinga\User $user)`

It's two new static methods because it allows us to potentially differentiate how to check this internally. For now, both just check whether a contact with the user's username exists.

### LINKED ISSUE #531: Normalize subscriber/recipient structure for integrations better
`\Icinga\Module\Notifications\Integrations\Incident::getSubscribers()` currently yields:

```
array{
  name: string,
  full_name: string,
  role: 'manager'|'subscriber',
  roleChangedAt: DateTime
}
```

Note: No `type` key as it'd always be `'contact'`.

`\Icinga\Module\Notifications\Integrations\Incident::getRecipients()` currently yields:

```
array{
  type: 'contact'|'contactgroup'|'schedule',
  name: string,
  full_name: ?string,
  roleChangedAt: DateTime
}
```

Note: No `role` key as it'd always be `'recipient'`.

But what can also be noticed, is that a recipient always has a **nullable** `full_name` key. It's nullable because contact groups and schedules have only a `name`. In the only case, when `type` is `'contact'`, the `full_name` actually carries the contact's full name. More interesting though, is that the `name` key is mandatory and is therefore set with the contact's username, which is **actually** nullable.

So, let's get this straight:

* Drop `full_name` entirely.
* Set a contact's full name as `name`
* Introduce `username: ?string`

BastianLedererIcinga:
resolve #529 
resolve #531 

Add `Incidents::canSubscribe()` and `Incidents::canManage()` so integrations can check if a subscribe/manage action is possible before attempting it.
To allow this `ContactRepository::findByUsername()` is introduced.

The  arrays yielded by `Incident::getSubscribers()` and `Incident::getRecipients()` are adjusted as suggested in #531

### DIFF
--- a/library/Notifications/Integrations/Incidents.php
+++ b/library/Notifications/Integrations/Incidents.php
@@ -9,6 +9,8 @@ use Countable;
 use Generator;
 use Icinga\Module\Notifications\Common\Database;
 use Icinga\Module\Notifications\Model\Incident as IncidentModel;
+use Icinga\Module\Notifications\Repository\ContactRepository;
+use Icinga\User;
 use ipl\Orm\Query;
 use ipl\Orm\ResultSet;
 use ipl\Sql\Connection;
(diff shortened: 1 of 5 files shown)


library/Notifications/Integrations/Incidents.php
```diff
@@ -58,6 +60,45 @@ public static function find(array $tags): static
...
+    /**
+     * Get whether the given user can manage incidents
+     *
+     * @param User $user
+     *
+     * @return bool
+     */
+    public static function canManage(User $user): bool
+    {
+        return static::hasContact($user);
```
nilmerg: Let's already think about testing this here, so this should internally instantiate self and `hasContact` isn't static anymore. And of course, a simple test for `hasContact` shouldn't hurt ;)

nilmerg (changes requested):

nilmerg (approved):

PR #541, opened 2026-08-13
Fix `Incidents::getRoles()`

BastianLedererIcinga:
`Incidents::getRoles()` can currently return several rows, keyed with `Incident` objects that share the same id, and may assign different roles to the given user.

The problem is that the current filter creates a subquery, that only checks if the given incident has an `incident_contact` row matching the `contact` with the given username. As long as this is the case, all `incident_contact` rows of the incident are contained in the result.

To fix this the `incident_contact.contact` relation is eagerly loaded, so the filter is no longer moved to a subquery and constrains the results.

### DIFF
--- a/library/Notifications/Integrations/Incidents.php
+++ b/library/Notifications/Integrations/Incidents.php
@@ -117,9 +117,12 @@ class Incidents implements IteratorAggregate, Countable
      */
     public function getRoles(User $user): Generator
     {
+        $filter = Filter::equal('incident_contact.contact.username', $user->getUsername());
+        $filter->metaData()->set('forceOptimization', false);
+
         $incidents = $this->buildQuery()
             ->withColumns(['role' => 'incident_contact.role'])
-            ->filter(Filter::equal('contact.username', $user->getUsername()));
+            ->filter($filter);
         foreach ($incidents as $incident) {
             yield new Incident($incident, $this->db) => $incident->role;
         }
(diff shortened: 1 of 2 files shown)


library/Notifications/Integrations/Incidents.php
```diff
@@ -118,8 +118,9 @@ public function hasIncident(): bool
     public function getRoles(User $user): Generator
     {
         $incidents = $this->buildQuery()
+            ->with('incident_contact.contact')
```
nilmerg: `utilize` instead of `with` should also work and doesn't have the side-effect of providing access to all contact details for the integration.

nilmerg (changes requested):

library/Notifications/Integrations/Incidents.php
BastianLedererIcinga: I just tried this and with `utilize` the added test case fails again, with the same problem that currently exists on main.

But setting `forceOptimization` to `false` in the filter's metadata fixes the original problem without eager loading the contact.

nilmerg (changes requested):
Looks fine now, but in light of what my previous review uncovered, please add a test that ensures we do not disclose any further contact details, or phrased differently, only as much columns as necessary.

nilmerg (approved):
Scratch that. There's no accessor for the incident model and we also don't accidentally give access to sub-classes since everything related is private.

PR #545, opened 2026-08-19
`Incidents`: Add separate factories for full id tags/partial id tags

BastianLedererIcinga:
Replace the `find` factory with 3 separate factories for different use cases.

`get()`: expects the full id tags of an object and return the matching incident or `null`
`getAll()`: expects an `iterable` that yields arrays if id tags, which are expected to be complete. the created `Incidents` instance will yield every matching incident.
`matchAll()`: allows partial matches.

In the case of `get()` the created `Incident` is instantiated with a query which is only executed when required.
If the query does not return an incident an `IncidentNotFoundException` is thrown.

`IncidentsTest` was adjusted with claude to match the new implementation.

### DIFF
--- a/library/Notifications/Integrations/Incident.php
+++ b/library/Notifications/Integrations/Incident.php
@@ -12,6 +12,7 @@ use Icinga\Module\Notifications\Model\Contact;
 use Icinga\Module\Notifications\Model\Incident as IncidentModel;
 use Icinga\Module\Notifications\Model\IncidentContact;
 use Icinga\Module\Notifications\Model\IncidentHistory;
+use Icinga\User;
 use InvalidArgumentException;
 use ipl\Sql\Connection;
 use ipl\Stdlib\Filter;
@@ -37,6 +38,27 @@ class Incident
     {
         $this->incident = $incident;
         $this->db = $db;
+    }
+
+    /**
+     * Get the given user's role for the incident, null if the user has no role
+     *
+     * @param User $user
+     *
+     * @return ?string
+     */
+    public function getRole(User $user): ?string
+    {
+        return IncidentContact::on($this->db)
+            ->columns('role')
+            ->filter(
+                Filter::all(
+                    Filter::equal('incident_id', $this->incident->id),
+                    Filter::equal('contact.username', $user->getUsername())
+                )
+            )
+            ->first()
+            ?->role;
     }
 
     /**
--- a/library/Notifications/Integrations/Incident.php
+++ b/library/Notifications/Integrations/Incident.php
@@ -8,35 +8,116 @@ namespace Icinga\Module\Notifications\Integrations;
 use DateTime;
 use Generator;
 use Icinga\Module\Notifications\Common\EntityManager;
+use Icinga\Module\Notifications\Integrations\Exception\IncidentNotFoundException;
 use Icinga\Module\Notifications\Model\Contact;
 use Icinga\Module\Notifications\Model\Incident as IncidentModel;
 use Icinga\Module\Notifications\Model\IncidentContact;
 use Icinga\Module\Notifications\Model\IncidentHistory;
+use Icinga\User;
 use InvalidArgumentException;
+use ipl\Orm\Query;
 use ipl\Sql\Connection;
+use ipl\Sql\Expression;
 use ipl\Stdlib\Filter;
+use LogicException;
 
 /**
  * Manage an incident's recipients and read its state
  */
 class Incident
 {
-    /** @var IncidentModel The managed incident */
-    private IncidentModel $incident;
+    /** @var ?IncidentModel The managed incident, null if it wasn't fetched yet */
+    private ?IncidentModel $incident = null;
+
+    /** @var ?Query<IncidentModel> The query to lazy load the incident */
+    private ?Query $query = null;
 
     /** @var Connection The database connection to use */
     private Connection $db;
 
-    /**
-     * Create a new wrapper for the given model
-     *
-     * @param IncidentModel $incident
-     * @param Connection $db The connection to read and persist through
-     */
-    public function __construct(IncidentModel $incident, Connection $db)
+    private function __construct()
     {
-        $this->incident = $incident;
-        $this->db = $db;
+    }
+
+    /**
+     * Create an instance from a query that should return one incident
+     *
+     * If the query does not return an incident, calling any function on the created instance throws an
+     * {@see IncidentNotFoundException}
+     *
+     * @param Query<IncidentModel> $query
+     *
+     * @return static
+     */
+    public static function fromQuery(Query $query): static
+    {
+        $incident = new static();
+        $incident->query = $query;
+        $incident->db = $query->getDb();
+
+        return $incident;
+    }
+
+    /**
+     * Create an instance from an {@see IncidentModel}
+     *
+     * Instances created with this factory will never throw an {@see IncidentNotFoundException}
+     *
+     * @param IncidentModel $model
+     * @param Connection $db
+     *
+     * @return static
+     */
+    public static function fromModel(IncidentModel $model, Connection $db): static
+    {
+        $incident = new static();
+        $incident->incident = $model;
+        $incident->db = $db;
+
+        return $incident;
+    }
+
+    /**
+     * Get the given user's role for the incident, null if the user has no role, throws if no matching incident exists
+     *
+     * @param User $user
+     *
+     * @return 'manager'|'subscriber'|'recipient'|null
+     *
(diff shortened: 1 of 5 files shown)


library/Notifications/Integrations/Incidents.php
```diff
@@ -160,16 +205,58 @@ private function incidents(): ResultSet
...
 
-        foreach ($this->tags as $tag => $value) {
-            if ($value === null) {
-                $query->filter(Filter::unlike("incident.object.tag.$tag", '*'));
-            } else {
-                $query->filter(Filter::equal("incident.object.tag.$tag", $value));
+    /**
+     * Get the filter matching the incidents of any of the tag sets
+     */
+    private function tagSetFilter(): Any
```
nilmerg: Let's make this a static method and provide one for exact matches and one for partial matches. This makes `$this->tagSetFilter` and `$this->allowPartialMatches` obsolete. There was also no need for tags to be available to the instance and thus `$this->tagSets` is also obsolete since it's the argument for the methods. To me it makes also sense to think about only passing a query instance to the constructor, even if that means any current consumer of it must clone it before use.

library/Notifications/Integrations/Incidents.php
nilmerg: Since you've introduced the dirty workaround below, let's introduce an elegant counterweight here:

```suggestion
        return $this->db->select(
            $this->buildQuery()
                ->columns([new Expression('1')])
                ->limit(1)
                ->assembleSelect()
                ->resetOrderBy()
        )->fetch() !== false;
```

library/Notifications/Integrations/Incident.php
```diff
@@ -39,6 +40,27 @@ public function __construct(IncidentModel $incident, Connection $db)
...
     }
 
+    /**
+     * Get the given user's role for the incident, null if the user has no role
+     *
+     * @param User $user
+     *
+     * @return ?string
+     */
+    public function getRole(User $user): ?string
```
nilmerg, line 89: Please rather think about another change in the `Incidents` signature. I'd like to avoid multiple queries introduced by limitations in code structure at all cost, let alone introduce another one here.

nilmerg (changes requested):

nilmerg (changes requested):
Thanks so far. But I feel there's more room for improvement 😉 

Looking at how this is used in Icinga DB Web shows that `hasIncident` and `getRoles` are now obsolete. That's rather a sign of the reason to only support a single incident, though.

But what Icinga DB Web, as an integration for a single incident, actually does are three things:

1. Showing subscribers and recipients
  a. also checks whether an incident is returned first
3. Showing a quick action to the current user
  a. only if an incident exists
5. Processing said quick action
  a. for a list of incidents, due to bulk actions

There's no reason to fetch the incident in any case, except for the reason to know it exists. But that aside, it is not needed to fetch any incident details at all. And the existence can be verified differently on the use case. `Incident::getRole` for example, may just throw an error, or return `null` and `none` for no role, to indicate a missing incident. `Incidents::get` must be lazy for this and thus may only be an alias for a factory of `Incident`.

And this approach I'd like to see also for `getAll` and `matchAll`. It must be mandatory to iterate the result to know something matched. `hasIncident` has really no use in this case and can safely be removed. `getRoles` may still have a use, but only in case of other integrations, but no known one and thus I'd object it's existence as well.

BastianLedererIcinga:
Local unit tests pass using the schema prior to https://github.com/Icinga/icinga-notifications/pull/496

I will adjust the added `external_uuid` usage in `IncidentTest` once #547 is merged.

library/Notifications/Integrations/Incident.php
```diff
@@ -8,35 +8,113 @@
...
+    }
+
+    /**
+     * Create an instance from a query that should return one incident
+     *
+     * If the query does not return an incident, calling any function on the created instance throws an
+     * {@see IncidentNotFoundException}
+     *
+     * @param Query<IncidentModel> $query
+     * @return static
```
nilmerg, line 50: empty line prior pls

library/Notifications/Integrations/Incident.php
```diff
@@ -8,35 +8,113 @@
...
+    }
+
+    /**
+     * Create an instance from an {@see IncidentModel}
+     *
+     * Instances created with this factory will never throw an {@see IncidentNotFoundException}
+     *
+     * @param IncidentModel $model
+     * @param Connection $db
+     * @return static
```
nilmerg, line 69: sam ehere

library/Notifications/Integrations/Incident.php
```diff
@@ -319,12 +407,32 @@ private function assignRole(string $username, string $role, array $noopRoles): s
...
         $history->contact_id = $contactId;
         $history->type = 'recipient_role_changed';
         $history->old_recipient_role = $oldRole;
         $history->new_recipient_role = $newRole;
         $history->time = new DateTime();
         (new EntityManager($this->db))->save($history);
     }
+
+    /**
+     * Get the value of {@see static::$incident}, fetch it if it is still null, throw if the query returns no incident
```
nilmerg: a little verbose don't you think? *"Fetch the incident lazily and return it"* suffices IMHO

library/Notifications/Integrations/Incidents.php
```diff
@@ -8,124 +8,121 @@
...
-     * @return static
+     * @param array<string, string> $tags The full set of id tags
+     *
+     * @return Incident
      */
-    public static function find(array $tags): static
+    public static function get(array $tags): Incident
     {
-        return new static($tags, Database::get());
+        // TODO: Once object_id no longer depends on the source_id, object_id should be used to optimize this
```
nilmerg: It's time 😉 so let's do it right away.

library/Notifications/Integrations/Incident.php
```diff
@@ -8,35 +8,113 @@
...
      */
-    public function __construct(IncidentModel $incident, Connection $db)
+    public function getRole(User $user): ?string
     {
-        $this->incident = $incident;
-        $this->db = $db;
+        if ($this->incident === null) {
+            $incidentContactTable = (new IncidentContact())->getTableName();
+            $contactTable = (new Contact())->getTableName();
+            $this->query
```
nilmerg: I think it's safe to assume that the caller either wants the current role or change it. And since you have two use cases already for `$this->query`, I suggest to introduce `$this->consumeQuery(): Query` which throws an exception if no query is set and unsets it otherwise prior returning. This way it's baked in that it can only be used by `getRole()` or `incident()`.

nilmerg (changes requested):

library/Notifications/Integrations/Incident.php
```diff
@@ -319,12 +410,51 @@ private function assignRole(string $username, string $role, array $noopRoles): s
...
+     * Single use getter for the query to lazy load the incident
+     *
+     * @return Query<IncidentModel>
+     *
+     * @throws LogicException If the query has already been consumed
+     */
+    private function consumeQuery(): Query
+    {
+        if ($this->query === null) {
+            throw new LogicException('Cannot fetch the incident again, the query has already been consumed');
```
nilmerg: One last bit, since it seems valid to call `getRole()` and then change the role because the first initializes `$this->incident` just like `incident()`, the message should point out this case as if it occurs the order is likely off.

nilmerg (changes requested):

library/Notifications/Integrations/Incident.php
BastianLedererIcinga: As discussed the described scenario is not possible, I adjusted the message for the only real reason, an earlier `IncidentNotFoundException`

nilmerg (dismissed):

nilmerg:
please rebase

library/Notifications/Integrations/Incidents.php
```diff
@@ -152,24 +143,80 @@ public function count(): int
...
-        $query = IncidentModel::on($this->db)->filter(Filter::unlike('recovered_at', '*'));
+        $tagSetFilter = new Any();
+        foreach ($tagSets as $tags) {
+            $tagSetFilter->add(Filter::equal('object_id', static::objectIdFor($tags)));
+        }
+
+        if ($tagSetFilter->isEmpty()) {
+            // If no tag sets were provided, this ensures no incidents are returned
+            $tagSetFilter->add(Filter::unlike('id', '*'));
+        }
```
nilmerg, line 183: This and the case in `tagSetFilterPartialMatches` should be illegal. No set at all or an empty set must throw.

nilmerg (changes requested):

nilmerg (approved):

PR #546, opened 2026-08-21
Support multiple related sources

### LINKED ISSUE #519: Support multiple related sources for objects and escalation rules
https://github.com/Icinga/icinga-notifications/issues/451 is about loosening those links and Icinga Notifications Web must also be adjusted to cope with this change.

## Incident list & detail

Both views show the source related to an incident right now. This needs to be extended to support multiple sources.

## Escalation rules (plus Notifications rules [if already implemented](https://github.com/Icinga/icinga-notifications-web/issues/452))

Rules currently expect the user to choose a specific source upon creation. This must be changed so that the user chooses a source *type* instead and creating the rule means establishing the link to all sources of this particular type.

## Source configuration

Creating a source right now does not have any side-effects. Only deletion causes all linked rules to also be deleted. Both must change.

Creating a source means establishing a link with all existing rules linked with other sources of the same type.
Deleting a source means keeping rules alive as long as one source is left, only upon deletion of the last one also rules are deleted.

BastianLedererIcinga:
resolve #519 
requires https://github.com/Icinga/icinga-notifications/pull/494

This PR allows escalation rules to be configured for a source type instead of an individual source and allows to display incidents that have more than one related source.

### Rules
For rules the configuration now asks for a source type instead of a source, and the database interaction is adjusted accordingly.
Deleting a source will now check if the source is the last remaining one of its type, and only then delete all escalation rules configured for it. The warning in `DeleteSourceForm` is adjusted to reflect this.

### Incidents
The `IncidentsController` previously eager loaded the `object.source` relation on its incidents, since an object can now be linked to several sources via `object_source`, this is no longer possible as it would duplicate incidents in the list.
To avoid having to lazy load sources for each incident the `SourceAggregator` behavior is added alongside a claude written test for it.

The incidents view now shows the icons of all related source types, and uses the source type as title rather than the name.

The detail view lists all related sources, the object link is created by the first source for which a hook is found.

Similar to the incident queries `Daemon.php` eager loaded the `object.source` relation, but it was unused by the subsequent code so I removed it in that case.

Tests pass locally on mysql with the schema changes introduced in https://github.com/Icinga/icinga-notifications/pull/494

### DIFF
--- a/application/controllers/EventRuleController.php
+++ b/application/controllers/EventRuleController.php
@@ -15,7 +15,6 @@ use Icinga\Module\Notifications\Common\SourceHookLocator;
 use Icinga\Module\Notifications\Data\NotificationConfigProvider;
 use Icinga\Module\Notifications\Forms\EventRuleConfigForm;
 use Icinga\Module\Notifications\Forms\EventRuleForm;
-use Icinga\Module\Notifications\Forms\SourceForm;
 use Icinga\Module\Notifications\Hook\V2\SourceHook;
 use Icinga\Module\Notifications\Model\Rule;
 use Icinga\Module\Notifications\Model\Source;
@@ -127,7 +126,7 @@ class EventRuleController extends CompatController
                         $this->addPart($form->prepareObjectFilterUpdate($this->session->get('object_filter')));
                         $this->addPart($form->prepareConfigUpdate(
                             $this->session->get('name'),
-                            $this->session->get('source')
+                            $this->session->get('source_type')
                         ));
                         $this->addPart(Html::tag('div', ['id' => 'event-rule-config-name'], [
                             Html::tag('h2', $this->session->get('name')),
@@ -140,7 +139,7 @@ class EventRuleController extends CompatController
                     } else {
                         $this->addPart($form->prepareConfigUpdate(
                             $this->session->get('name'),
-                            $this->session->get('source')
+                            $this->session->get('source_type')
                         ));
                         $this->addPart($form->prepareObjectFilterUpdate($this->session->get('object_filter')));
                     }
@@ -155,15 +154,15 @@ class EventRuleController extends CompatController
                     $form->setRule($rule);
 
                     $this->session->set('name', $rule->name);
-                    $this->session->set('source', $rule->source_id);
+                    $this->session->set('source_type', $rule->source_type);
                     $this->session->set('object_filter', $rule->object_filter ?? '');
                 } else {
                     $name = $this->params->getRequired('name');
-                    $source = (int) $this->params->getRequired('source');
-                    $form->populate(['name' => $name, 'source' => $source]);
+                    $source = $this->params->getRequired('source_type');
+                    $form->populate(['name' => $name, 'source_type' => $source]);
 
                     $this->session->set('name', $name);
-                    $this->session->set('source', $source);
+                    $this->session->set('source_type', $source);
                     $this->session->set('object_filter', '');
                 }
             })
@@ -204,7 +203,6 @@ class EventRuleController extends CompatController
     {
         $ruleId = (int) $this->params->getRequired('id');
         $filter = $this->params->get('object_filter', $this->session->get('object_filter'));
-        $hook = $this->resolveSourceHook($ruleId);
 
         $parsedFilter = null;
         if ($filter) {
@@ -230,6 +228,8 @@ class EventRuleController extends CompatController
                 ));
             }
         }
+
+        $hook = $this->resolveSourceHook($ruleId, (bool) ($parsedFilter['assisted'] ?? false));
 
         $editor = (new SearchEditor())
             ->addAttributes(Attributes::create(['class' => 'event-rule-filter']))
@@ -317,11 +317,17 @@ class EventRuleController extends CompatController
             };
         }
 
-        $editor->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $getJsonPaths) {
+        $assisted = $hook !== null;
+        $editor->on(Form::ON_SUBMIT, function (SearchEditor $form) use ($ruleId, $getJsonPaths, $assisted) {
             $filter = $form->getFilter();
             $this->session->set(
                 'object_filter',
-                (new RuleSerializer($filter, $getJsonPaths($filter), $form->getValue('filter_name')))->getJson()
+                (new RuleSerializer(
+                    $filter,
+                    $getJsonPaths($filter),
+                    $assisted,
+                    $form->getValue('filter_name')
+                ))->getJson()
             );
             $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
         })->handleRequest($this->getServerRequest());
@@ -349,7 +355,7 @@ class EventRuleController extends CompatController
 
     public function suggestAction(): void
     {
-        $hook = $this->resolveSourceHook((int) $this->params->getRequired('id'));
+        $hook = $this->resolveSourceHook((int) $this->params->getRequired('id'), true);
         $requestData = SearchSuggestions::parseRequest($this->getServerRequest()) ?? [];
 
         $type = $requestData['term']['type'] ?? null;
@@ -382,37 +388,31 @@ class EventRuleController extends CompatController
         $this->getDocument()->addHtml($suggestions);
     }
 
-    protected function resolveSourceHook(int $ruleId): ?SourceHook
+    protected function resolveSourceHook(int $ruleId, bool $required): ?SourceHook
     {
-        $source = null;
-        if ($ruleId !== -1) {
-            $source = Rule::on(Database::get())
-                ->columns(['id' => 'source.id', 'type' => 'source.type'])
+        $type = $this->session->get('source_type');
+        if ($type === null && $ruleId !== -1) {
+            $type = Rule::on(Database::get())
+                ->columns(['source_type'])
                 ->filter(Filter::equal('id', $ruleId))
-                ->first();
-        } elseif (isset($this->session->source)) {
-            /** @var ?Source $source */
-            $source = Source::on(Database::get())
-                ->columns(['id', 'type'])
-                ->filter(Filter::equal('id', $this->session->source))
-                ->first();
+                ->first()?->source_type;
         }
 
-        if ($source === null) {
+        if ($type === null) {
             $this->httpNotFound($this->translate('Rule not found'));
         }
 
-        if ($source->type === SourceForm::TYPE_GENERIC) {
-            return null;
-        }
+        $hook = SourceHookLocator::forType($type);
 
-        $hook = SourceHookLocator::forType($source->type);
+        if (! $required) {
+            return $hook;
+        }
 
         if ($hook === null) {
             $this->httpNotFound(sprintf($this->translate(
                 'No source integration available. Either the module supporting sources of type "%s" is not'
                 . ' enabled or you have insufficient privileges. Please contact your system administrator.'
-            ), $source->type));
+            ), $type));
         }
 
         return $hook;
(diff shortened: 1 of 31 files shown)


application/controllers/IncidentsController.php
```diff
@@ -33,8 +33,8 @@ public function indexAction(): void
         $this->addTitleTab(t('Incidents'));
 
         $incidents = Incident::on(Database::get())
-            ->with(['object', 'object.source'])
-            ->withColumns('object.id_tags');
+            ->with(['object'])
+            ->withColumns(['object.sources']);
```
BastianLedererIcinga, line 37: This will look like a mistake in the diff, but it's deliberate.
The `object.id_tags` are not used by `IncidentRenderer`. I suppose this was simply copy pasted from `IncidentController` where they actually have a purpose.

nilmerg:
> The incidents view now shows the icons of all related source types, and uses the source type as title rather than the name.

Since the detail shows both informations, `name (type)`, the list should do so as well. The type alone doesn't make much sense.

nilmerg:
The rule editor dropdown now shows the type's of the available sources. While this is fine regarding generic sources, integrated sources should be designated with a proper label. Luckily, it seems we already require a label from those as part of `\Icinga\Module\Notifications\Hook\V2\SourceHook::getSourceLabel` :tada:

application/controllers/EventRuleController.php
```diff
@@ -424,22 +420,26 @@ public function editAction(): void
 
         $eventRuleForm = (new EventRuleForm())
             ->setCsrfCounterMeasureId(Session::getSession()->getId())
-            ->setAvailableSources(
-                Database::get()->fetchPairs(
-                    Source::on(Database::get())->columns(['id', 'name'])->assembleSelect()
+            ->setAvailableSourceTypes(
+                array_column(
```
nilmerg: I know this will probably change anyway when showing a hook's source label but `array_column` looks weird to me anyway. There's definitely a better way, which I'll see if I have to suggest it when the update arrived.

application/controllers/IncidentsController.php
nilmerg: It *had* a purpose, [while name rendering was still up to the integration](https://github.com/Icinga/icinga-notifications-web/commit/f6becb68d3bac30fe8035f843567c97d4a15ab9d#diff-6b3849ec4a550aa37447949ab7fd8c2edab954c7299cd8bc98ee225a5269d10dL97).

library/Notifications/Repository/SourceRepository.php
```diff
@@ -106,9 +126,10 @@ public function delete(int $id): void
         $source->listener_username = null;
         $source->delete();
 
-        $source->rule->query()->columns('id');
```
nilmerg: That's not been a mistake, it restricts the columns being selected to what's really required. If that does not suit anymore, because `type` is missing, extend it instead.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -55,6 +56,25 @@ public function findByUsername(string $username): ?Source
...
     }
 
+    /**
+     * Get whether the given source is the only remaining one of its type
+     *
+     * @param Source $source
+     *
+     * @return bool
+     */
+    public function isLastOfItsType(Source $source): bool
```
nilmerg: I'd rather suggest to perform the query where it's required differently. This should be part of the controller, for the other usage, see below.

library/Notifications/Repository/SourceRepository.php
```diff
@@ -106,9 +126,10 @@ public function delete(int $id): void
         $source->listener_username = null;
         $source->delete();
 
-        $source->rule->query()->columns('id');
-        foreach ($source->rule as $rule) {
-            (new EscalationRuleRepository($this->db))->delete($rule->id);
+        if ($this->isLastOfItsType($source)) {
+            foreach ($source->rule as $rule) {
+                (new EscalationRuleRepository($this->db))->delete($rule->id);
+            }
```
nilmerg: Restricting the columns was not a mistake, it constraints the result to provide what's really required. I'd rather bake this in the query as a filter instead anyway, so please restore the query adjustment and extend it with `->filter(Filter::unequal('source.id', $id))`. Didn't verify this works, but even if not there's probably an easy fix.

library/Notifications/Widget/Detail/IncidentDetail.php
```diff
@@ -160,20 +164,22 @@ protected function createHistory(): array
...
+        if (empty($this->incident->object->sources)) {
             return [
                 Html::tag('h2', $this->translate('Event Source')),
                 new EmptyState($this->translate('No source information available'))
             ];
         }
 
         $list = new HtmlElement('ul', Attributes::create(['class' => 'source-list']));
-        $list->addHtml(new HtmlElement('li', null, new EventSourceBadge($this->incident->object->source)));
+        foreach ($this->incident->object->sources as $source) {
```
nilmerg, line 177: Once the update arrives, I'm sure this can avoid issuing another query and only rely on `sources`.

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,108 @@
...
+
+    final public function __construct()
+    {
+        parent::__construct(['sources']);
+    }
+
+    /** @return Source[] */
+    public function fromDb($value, $key, $context): array
+    {
+        $sources = [];
```
nilmerg, line 55: `IdTagAggregator` uses `is_string` prior, why not here? json_decode only accepts string anyway.

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,108 @@
...
+        foreach (json_decode($value ?? '', true) ?? [] as $row) {
+            $sources[] = (new Source())->setProperties($row);
```
nilmerg: `new Source($row)`

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,108 @@
...
+            $row = sprintf(
+                $isPgsql
+                    ? "json_build_object('id', %s, 'type', %s, 'name', %s)"
+                    : "json_object('id', %s, 'type', %s, 'name', %s)",
+                $subResolver->qualifyColumn('id', $alias),
+                $subResolver->qualifyColumn('type', $alias),
+                $subResolver->qualifyColumn('name', $alias)
+            );
+
+            $subQuery->columns([new Expression(($isPgsql ? 'json_agg' : 'json_arrayagg') . "($row)")]);
```
nilmerg: Have you tried using the `columns` argument of the expression?

nilmerg (changes requested):

library/Notifications/Repository/SourceRepository.php
BastianLedererIcinga: Easy fix...
This fails for the preexisting test and the one I added.
I also tried `(Filter::none(Filter::equal(...)))`, which of course behaves entirely different...
But it would only pass the added test if the relation used inner join, it uses left.

The version i implemented is different, but it also only needs one query so it's the best I can do.

library/Notifications/Model/Source.php
```diff
@@ -136,6 +139,16 @@ public function getIcon(): Icon
...
     }
 
+    /**
+     * Get the label of this source's type
+     *
+     * @return string
+     */
+    public function getLabel(): string
+    {
+        return SourceHookLocator::labelFor($this->type);
```
nilmerg: Just saw how `labelFor` is implemented. Don't you think there's some similarity with `getIcon` above?

But what I'm actually pointing out here, is that we seem both not fully okay with the case that the source model provides the accessor for both. I'm not a fan anymore of `::getIcon()`, thus I don't see the justification for `::getLabel()` either. Especially given how it's implemented here, as an alias only.

So let's drop this here, and since I cannot suggest a good place for it, let any callsite of `::getLabel()` use `SourceHookLocator::labelFor($type)` instead.

library/Notifications/Util/RuleSerializer.php
```diff
@@ -26,17 +26,22 @@ class RuleSerializer
     /** @var array<string, string[]> JSON paths keyed by column name */
     protected array $jsonPaths;
 
+    /** @var bool Whether the filter was created with a source integration */
+    protected bool $integrated;
```
nilmerg: Just a nitpick, but looking from the perspective of a self-explanatory name, wouldn't `assisted` be better?

library/Notifications/Util/RuleSerializer.php
```diff
@@ -52,6 +57,7 @@ public function getJson(): ?string
         $result = [
             'version' => self::VERSION,
             'qs'      => QueryString::render($this->filter),
+            'integrated' => $this->integrated,
```
nilmerg: The previous indentation mandates that you follow it, or change everything accordingly ;)

application/controllers/EventRuleController.php
```diff
@@ -231,6 +229,9 @@ public function searchEditorAction(): void
             }
         }
 
+        $integrated = $parsedFilter === null ? null : (bool) ($parsedFilter['integrated'] ?? false);
```
nilmerg: Doesn't this suffice?

```suggestion
        $integrated = (bool) ($parsedFilter['integrated'] ?? false);
```

application/controllers/EventRuleController.php
```diff
@@ -382,37 +389,35 @@ public function suggestAction(): void
         $this->getDocument()->addHtml($suggestions);
     }
 
-    protected function resolveSourceHook(int $ruleId): ?SourceHook
+    protected function resolveSourceHook(int $ruleId, ?bool $integrated): ?SourceHook
```
nilmerg: Yep, not nullable, deal with 😉 

```suggestion
    protected function resolveSourceHook(int $ruleId, bool $required): ?SourceHook
```

application/forms/SourceForm.php
```diff
@@ -96,7 +90,7 @@ protected function assemble(): void
             Text::create($this->translate(
                 'Sources are the most vital part of Icinga Notifications. They submit events that will be'
                 . ' processed to notify users about incidents. You can either configure sources that provide an'
-                . ' integration in Icinga Web, or use the generic type for sources that communicate directly with'
+                . ' integration in Icinga Web, or configure generic sources that communicate directly with'
                 . ' the Icinga Notifications API.'
```
nilmerg: > You can either configure sources that provide an integration in Icinga Web, or configure generic sources that communicate directly with the Icinga Notifications API.

Not sure that distinction is necessary to mention anymore?

> Sources are the most vital part of Icinga Notifications. They submit events that will be processed to notify users about incidents. You configure here how they relate to event rules and their credentials they will present to communicate with the Icinga Notifications API.

application/forms/SourceForm.php
```diff
@@ -110,48 +104,30 @@ protected function assemble(): void
...
+
+        $this->addHtml(
+            new HtmlElement(
+                'p',
+                Attributes::create(['class' => 'description']),
+                Text::create(
+                    $this->translate(
+                        'Enter the source identifier as stated in the source\'s documentation.'
+                        . ' Note that integrated sources usually provide their own configuration interface for'
+                        . ' notifications, which is the recommended way to set them up.'
```
nilmerg: > The source type is used establish a link to event rules. Enter the value as stated in the source's documentation. Note that integrated sources usually provide their own configuration interface for notifications, which is the recommended way to set them up.

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,105 @@
...
+    final public function __construct()
+    {
+        parent::__construct(['sources']);
+    }
+
+    /** @return Source[] */
+    public function fromDb($value, $key, $context): array
+    {
+        if (! is_string($value)) {
+            return [];
```
nilmerg, line 46: Let's throw here, receiving something else surely implies an error or bug?

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,105 @@
...
+use ipl\Orm\Contract\QueryAwareBehavior;
+use ipl\Orm\Contract\RewriteColumnBehavior;
+use ipl\Orm\Exception\InvalidColumnException;
+use ipl\Orm\Query;
+use ipl\Sql\Adapter\Pgsql;
+use ipl\Sql\Expression;
+use ipl\Stdlib\Filter;
+
+/**
+ * Aggregate all sources an object is linked to into the column `sources`
```
nilmerg, line 25: Please also document the value structure

test/php/library/Notifications/Repository/SourceRepositoryTest.php
```diff
@@ -339,6 +359,26 @@ public function testDeleteSoftDeletesLinkedRules(Connection $db): void
...
+
+        (new SourceRepository($db))->delete($sourceId);
+
+        $this->assertSame(
+            'y',
+            $this->loadRawEntity($db, $ruleId, Rule::class)->deleted,
+            'The rule must be soft-deleted even if an unrelated source of another type has no rules'
+        );
+    }
+
```
nilmerg: I'd question the existence of this test. It boils down to how the relationship `source<->rule` is joined, which isn't related only to deletion at all. Since this relates to how rules are identified, my guess is that this test is obsolete once I fixed the error with https://github.com/Icinga/ipl-orm/pull/171.

nilmerg (changes requested):

library/Notifications/Model/Behavior/SourceAggregator.php
```diff
@@ -0,0 +1,118 @@
...
+     */
+    public function fromDb($value, $key, $context): array
+    {
+        if ($value === null) {
+            return [];
+        }
+
+        if (! is_string($value)) {
+            throw new InvalidArgumentException(
+                sprintf('Expected JSON array or null for %s, got %s', $key, get_debug_type($value))
```
nilmerg: Use `\ipl\Stdlib\get_php_type` instead of `get_debug_type`.

nilmerg (changes requested):

nilmerg (approved):
Thank you!

PR #547, opened 2026-08-26
Handle UUID's correctly for both DBMs

yhabteab:
This PR adds support for UUID binary and text format in all UI activities involving UUIDs. For this to work, I've introduced a new `UUID` behavior in IPL orm that is used to automatically convert UUIDs between binary and text formats for MySQL but is a no-op for PostgreSQL as PDO treats the binary format as a string when not explicitly told otherwise via the `PDO::PARAM_LOB` parameter. Since I've no way to do that parameter binding in all possible cases, I've decided to leave PostgreSQL as-is.

I've also updated the API handlers to partially use the Models instead of raw SQL queries via ipl SQL, so that the UUID can be used in those cases as well. The API handlers are still using raw SQL queries in some places, but I've added an explicit conversion to binary format for those cases, so that the UUIDTransformer is not needed there.

Lastly, I've extended the `ObjectSuggestionsCursor` class to require the column name for which the suggestions are being generated, and the registered model behaviors the column belongs to, so that it also applies the behaviors, including the UUID behavior to the suggestions.

## Context

- https://github.com/Icinga/icinga-notifications/pull/496

### DIFF
--- a/library/Notifications/Api/V1/Channels.php
+++ b/library/Notifications/Api/V1/Channels.php
@@ -115,7 +115,7 @@ class Channels extends ApiV1 implements RequestHandlerInterface, EndpointInterfa
             return $this->getPlural($queryFilter, $stmt);
         }
 
-        $stmt->where(['external_uuid = ?' => $identifier]);
+        $stmt->where(['external_uuid = ?' => static::transformUUIDForDB(Database::get(), $identifier)]);
 
         /** @var stdClass|false $result */
         $result = Database::get()->fetchOne($stmt);
@@ -194,7 +194,7 @@ class Channels extends ApiV1 implements RequestHandlerInterface, EndpointInterfa
                 ->from('channel')
                 ->columns('id')
                 ->where([
-                    'external_uuid = ?' => $channelIdentifier,
+                    'external_uuid = ?' => static::transformUUIDForDB(Database::get(), $channelIdentifier),
                     'deleted = ?' => 'n'
                 ])
         );
@@ -227,6 +227,7 @@ class Channels extends ApiV1 implements RequestHandlerInterface, EndpointInterfa
 
     public function prepareRow(stdClass $row): void
     {
+        $row->id = static::getUUIDString($row->id);
         $row->config = Json::decode($row->config, true);
         unset($row->channel_id);
     }
--- a/library/Notifications/Api/V1/ContactGroups.php
+++ b/library/Notifications/Api/V1/ContactGroups.php
@@ -142,7 +142,7 @@ class ContactGroups extends ApiV1 implements RequestHandlerInterface, EndpointIn
             return $this->getPlural($queryFilter, $stmt);
         }
 
-        $stmt->where(['external_uuid = ?' => $identifier]);
+        $stmt->where(['external_uuid = ?' => static::transformUUIDForDB(Database::get(), $identifier)]);
 
         /** @var stdClass|false $result */
         $result = Database::get()->fetchOne($stmt);
@@ -419,7 +419,7 @@ class ContactGroups extends ApiV1 implements RequestHandlerInterface, EndpointIn
      */
     public static function fetchGroupIdentifiers(int $contactId): array
     {
-        return Database::get()->fetchCol(
+        $groupsUUIDs = Database::get()->fetchCol(
             (new Select())
                 ->from('contactgroup_member cgm')
                 ->columns('cg.external_uuid')
@@ -427,6 +427,8 @@ class ContactGroups extends ApiV1 implements RequestHandlerInterface, EndpointIn
                 ->where(['cgm.contact_id = ?' => $contactId, 'cgm.deleted = ?' => 'n'])
                 ->groupBy('cg.external_uuid')
         );
+
+        return array_map(static::getUUIDString(...), $groupsUUIDs);
     }
 
     /**
@@ -444,7 +446,7 @@ class ContactGroups extends ApiV1 implements RequestHandlerInterface, EndpointIn
                 ->from('contactgroup')
                 ->columns('id')
                 ->where([
-                    'external_uuid = ?' => $identifier,
+                    'external_uuid = ?' => static::transformUUIDForDB(Database::get(), $identifier),
                     'deleted = ?' => 'n'
                 ])
         );
@@ -582,6 +584,7 @@ class ContactGroups extends ApiV1 implements RequestHandlerInterface, EndpointIn
 
     public function prepareRow(stdClass $row): void
     {
+        $row->id = static::getUUIDString($row->id);
         $row->users = Contacts::fetchUserIdentifiers($row->contactgroup_id);
 
         unset($row->contactgroup_id);
--- a/library/Notifications/Model/Channel.php
+++ b/library/Notifications/Model/Channel.php
@@ -10,15 +10,17 @@ use Icinga\Module\Notifications\Common\Collection;
 use Icinga\Module\Notifications\Common\Model;
 use ipl\Orm\Behavior\BoolCast;
 use ipl\Orm\Behavior\MillisecondTimestamp;
+use ipl\Orm\Behavior\UUID;
 use ipl\Orm\Behaviors;
 use ipl\Orm\Query;
 use ipl\Orm\Relations;
 use ipl\Stdlib\Filter;
 use ipl\Web\Widget\Icon;
+use Ramsey\Uuid\UuidInterface;
 
 /**
  * @property int $id
- * @property ?string $external_uuid
+ * @property ?UuidInterface $external_uuid
  * @property string $name
  * @property string $type
  * @property ?string $config
@@ -79,6 +81,7 @@ class Channel extends Model
     {
         $behaviors->add(new MillisecondTimestamp(['changed_at']));
         $behaviors->add(new BoolCast(['deleted']));
+        $behaviors->add(new UUID(['external_uuid']));
     }
 
     public function createRelations(Relations $relations): void
--- a/library/Notifications/Test/DbTestBackends.php
+++ b/library/Notifications/Test/DbTestBackends.php
@@ -5,8 +5,10 @@
 
 namespace Icinga\Module\Notifications\Test;
 
+use ipl\Sql\Adapter\Pgsql;
 use ipl\Sql\Connection;
 use ipl\Sql\Test\SharedDatabases;
+use Ramsey\Uuid\Uuid;
 use RuntimeException;
 
 /**
@@ -61,6 +63,15 @@ SQL;
      * @return void
      */
     abstract protected static function initializeNotificationsDb(Connection $db): void;
+
+    protected static function transformUUIDForDB(Connection $db, string $uuid): string
+    {
+        if (! $db->getAdapter() instanceof Pgsql) {
+            return Uuid::fromString($uuid)->getBytes();
+        }
+
+        return $uuid;
+    }
 
     public static function setUpSchema(Connection $db, string $driver): void
     {
--- a/library/Notifications/Util/ObjectSuggestionsCursor.php
+++ b/library/Notifications/Util/ObjectSuggestionsCursor.php
@@ -5,18 +5,47 @@
 
 namespace Icinga\Module\Notifications\Util;
 
+use BackedEnum;
+use DateTime;
+use DateTimeInterface;
+use ipl\Orm\Behaviors;
+use ipl\Sql\Connection;
 use ipl\Sql\Cursor;
+use ipl\Sql\Select;
+use Ramsey\Uuid\UuidInterface;
 
 class ObjectSuggestionsCursor extends Cursor
 {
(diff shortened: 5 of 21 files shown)


library/Notifications/Api/V1/Channels.php
```diff
@@ -115,7 +117,7 @@ public function get(?string $identifier, string $queryFilter): ResponseInterface
             return $this->getPlural($queryFilter, $stmt);
         }
 
-        $stmt->where(['external_uuid = ?' => $identifier]);
+        $stmt->where($this->assembleFilter("id=$identifier", ['id'], 'external_uuid'));
```
nilmerg: That's a mis-use of the filter rendering, please find an alternative.

library/Notifications/Api/V1/Channels.php
```diff
@@ -184,22 +186,15 @@ private function getPlural(string $queryFilter, Select $stmt): ResponseInterface
...
-                    'external_uuid = ?' => $channelIdentifier,
-                    'deleted = ?' => 'n'
-                ])
-        );
+        $channel = Channel::on(Database::get())
+            ->filter(Filter::equal('external_uuid', $channelIdentifier))
+            ->first();
 
-        return $channel->id ?? false;
+        return $channel?->id;
```
nilmerg: I see no reason to mix ipl-orm (here) and ipl-sql (below). A few lines above you already don't rely on the ORM so why here? Keep using ipl-sql please.

library/Notifications/Model/Behavior/UUIDTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+use Ramsey\Uuid\Uuid;
+
+class UUIDTransformer extends PropertyBehavior implements QueryAwareBehavior
+{
+    /** @var bool Whether the query is using a pgsql adapter */
+    protected bool $isPostgres = true;
+
+    public function fromDb($value, $key, $_)
+    {
+        if (! $this->isPostgres && $value != null) {
```
nilmerg: strict unequal pls

library/Notifications/Model/Behavior/UUIDTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+            return Uuid::fromBytes($value)->toString();
+        }
+
+        return $value;
+    }
+
+    public function toDb($value, $key, $_)
+    {
+        if ($value === null) {
```
nilmerg: Also exclude `ExpressionInterface` instances here

library/Notifications/Model/Behavior/UUIDTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+
+class UUIDTransformer extends PropertyBehavior implements QueryAwareBehavior
+{
+    /** @var bool Whether the query is using a pgsql adapter */
+    protected bool $isPostgres = true;
+
+    public function fromDb($value, $key, $_)
+    {
+        if (! $this->isPostgres && $value != null) {
+            return Uuid::fromBytes($value)->toString();
```
nilmerg: I think, if it's not required due to implicit type casting, we might always return a UUID instance here. Only if that's a problem right now a string is required.

library/Notifications/Model/Behavior/UUIDTransformer.php
nilmerg, line 1: Given the likely need for this in Kubernetes Web, this should be introduced to ipl-orm.

library/Notifications/Test/DbTestBackends.php
```diff
@@ -62,6 +64,15 @@ trait DbTestBackends
      */
     abstract protected static function initializeNotificationsDb(Connection $db): void;
 
+    protected static function transformUUIDForDB(Connection $db, string $uuid): string
```
nilmerg, line 67: that's what I call a pragmatic solution! Why don't introduce such a thing for the actual api as well?

library/Notifications/Util/ObjectSuggestionsCursor.php
```diff
@@ -5,18 +5,38 @@
...
-            // We need to do it this way as the suggestion implementation bypasses ORM behaviors here and there.
-            if (is_resource($value)) {
-                $value = stream_get_contents($value);
+            $value = $this->behaviors->retrieveProperty($value, $this->column);
+            if ($value instanceof DateTime) {
+                // The search bar can't handle date time objects, so convert it back to milliseconds again.
+                $value = $value->format('Uv');
+            } elseif (is_bool($value)) {
+                // Same goes with booleans, the search bar can't render boolean values either.
+                $value = $value ? 'y' : 'n';
```
nilmerg, line 46: this missed the `EnumCast` behavior part of `Incident` and does not take the chance to render binary values as hex.

nilmerg (changes requested):

library/Notifications/Api/V1/Channels.php
yhabteab: I didn't want to change the a few lines above case because you told me yestertady that you don't want to use models for the API responses, and I didn't change the few lines below case because, well, I don't have to as it doesn't interact with the `external_uuid` column but using IPL Orm there would be perfectly fine too. I've changed this, so I don't have to work around that uuid problem since the respective models will handle this fine.

library/Notifications/Api/V1/Channels.php
yhabteab: A mis-use? Maybe, but it's not wrong either and works just fine 🤷.

library/Notifications/Model/Behavior/UUIDTransformer.php
yhabteab: Where do we use expression columns with UUIDs? I can add a check for that too but I didn't see a reason for it.

nilmerg: I noticed this flaw already in the `Binary` behavior. It's not used yet, but a safety measure as this is accepted by other parts and thus is a valid case that must be handled.

library/Notifications/Api/V1/Channels.php
nilmerg: The API implementation is deliberately not relying on the ORM for reads. But that's not the actual problem I have, it's the mixed approach. The *few lines above* also work with the identifier and thus are very well subject to the same approach here.

library/Notifications/Util/ObjectSuggestionsCursor.php
yhabteab: > this missed the `EnumCast` behavior part of `Incident`

That enum is of type string, shouldn't that be automatically casted to the respective string value somehow? I can't test it in the search bar right as I already broke the incident veiw by applying the schema changes from https://github.com/Icinga/icinga-notifications/pull/494.

> and does not take the chance to render binary values as hex.

Whoat, how? Shouldn't the `Binary` behavior have already been run on such columns?

nilmerg: Enums, value backed or not, cannot be cast to string. You'd have to call `getValue` on it first.

> Whoat, how? Shouldn't the Binary behavior have already been run on such columns?

The previous implementation did the job of the Binary behavior. Since you apply the behaviors now you reasonably removed that part, but the behavior has the same *behavior* to return a raw binary string, implied by it's name. And since the new implementation handles other behavior's *behaviors* I don't see a reason to ignore the Binary one now.

yhabteab:
Changed as requested but all the tests fail because of this dependency.
- https://github.com/Icinga/ipl-orm/pull/172

library/Notifications/Api/V1/ContactGroups.php
```diff
@@ -419,14 +419,16 @@ public function delete(string $identifier): ResponseInterface
...
+        $groupsUUIDs = Database::get()->fetchCol(
             (new Select())
                 ->from('contactgroup_member cgm')
                 ->columns('cg.external_uuid')
                 ->joinLeft('contactgroup cg', 'cg.id = cgm.contactgroup_id')
                 ->where(['cgm.contact_id = ?' => $contactId, 'cgm.deleted = ?' => 'n'])
                 ->groupBy('cg.external_uuid')
         );
+
+        return array_map(function ($uuid) { return static::getUUIDString($uuid); }, $groupsUUIDs);
```
nilmerg:
```suggestion
        return array_map(static::getUUIDString(...), $groupUUIDs);
```

nilmerg (changes requested):

library/Notifications/Api/V1/ContactGroups.php
yhabteab: Noice.

nilmerg (commented):
Still need to test, but code looks fine.

library/Notifications/Model/Channel.php
nilmerg: This is now not a string anymore, please change this to `?UUIDInterface` (plus the other two models)

test/php/library/Notifications/Repository/ContactGroupRepositoryTest.php
```diff
@@ -122,7 +124,7 @@ public function testCreateStoresTheGivenExternalUuid(Connection $db): void
         $uuid = '00000000-0000-4000-8000-0000000000e1';
         $id = (new ContactGroupRepository($db))->create(new ContactGroupData(null, 'Group A', [], $uuid));
 
-        $this->assertSame($uuid, (new ContactGroupRepository($db))->find($id)->external_uuid);
+        $this->assertSame($uuid, (string) (new ContactGroupRepository($db))->find($id)->external_uuid);
```
nilmerg: It's not a string, call `getString`

test/php/library/Notifications/Repository/ContactRepositoryTest.php
```diff
@@ -196,7 +197,7 @@ public function testCreateStoresTheGivenExternalUuid(Connection $db): void
             externalUuid: $uuid
         ));
 
-        $this->assertSame($uuid, (new ContactRepository($db))->find($id)->external_uuid);
+        $this->assertSame($uuid, (string) (new ContactRepository($db))->find($id)->external_uuid);
```
nilmerg: `getString`

library/Notifications/Util/ObjectSuggestionsCursor.php
```diff
@@ -5,18 +5,45 @@
...
         foreach (parent::getIterator() as $key => $value) {
-            // TODO(lippserd): This is a quick and dirty fix for PostgreSQL binary datatypes for which PDO returns
-            // PHP resources that would cause exceptions since resources are not a valid type for attribute values.
-            // We need to do it this way as the suggestion implementation bypasses ORM behaviors here and there.
-            if (is_resource($value)) {
-                $value = stream_get_contents($value);
+            $value = $this->behaviors->retrieveProperty($value, $this->column);
+            if ($value instanceof DateTime) {
+                // The search bar can't handle date time objects, so convert it back to milliseconds again.
+                $value = $value->format('Uv');
```
nilmerg: Use `DateTime::RFC3339_EXTENDED` instead, otherwise no match is possible, and it's more readable than a integer.

library/Notifications/Util/ObjectSuggestionsCursor.php
```diff
@@ -5,18 +5,45 @@
...
+            } elseif ($value instanceof UuidInterface) {
+                $value = $value->toString();
```
nilmerg: `(string) $value`

library/Notifications/Util/ObjectSuggestionsCursor.php
```diff
@@ -5,18 +5,45 @@
...
+            } elseif (is_bool($value)) {
+                // Same goes with booleans, the search bar can't render boolean values either.
+                $value = $value ? 'y' : 'n';
+            } elseif ($value === 'severity') {
+                $value = $value->getValue();
```
nilmerg:
```suggestion
            } elseif ($value instanceof \BackedEnum) {
                $value = $value->value;
```

nilmerg (changes requested):

library/Notifications/Util/ObjectSuggestionsCursor.php
yhabteab: Seriously? Why? What `toString()` does is, converting that value to string.

library/Notifications/Util/ObjectSuggestionsCursor.php
yhabteab: Hmm, I though, one can't use instanceof with enums.

library/Notifications/Util/ObjectSuggestionsCursor.php
yhabteab: Hrrrrmm. You're right.

yhabteab:
I've addressed all your requested changes apart from https://github.com/Icinga/icinga-notifications-web/pull/547#discussion_r3869972858 because it doesn't make any sense to me.

nilmerg (approved):

PR #557, opened 2026-09-10
Cleanup form error handling

nilmerg:
Forms rely mostly on event handling right now but that has a limitation regarding errors in the framework at the moment. The previous work-around now lead to the problem that exceptions upon submit were not rendered as form errors but silently ignored and the form just didn't submit at all without any explanation. The new way to solve the originating problem is to always render form errors but still override `onError` in forms as there is a case in which it is called but no error event is emitted: In case element validation fails. It is now possible to only rely on `Form::ON_ERROR` to ensure the client's DOM is updated to reflect the state of the form properly.

requires https://github.com/Icinga/ipl-web/pull/408


nilmerg:
@Jan-Schuppik You can test this easily by checking out https://github.com/Icinga/ipl-orm/pull/171 and trying to create a contact group.

Jan-Schuppik (approved):
Looks good and works as expected 

Image

PR #558, opened 2026-09-11
Open escalations in modals

### LINKED ISSUE #457: Open Escalations in Modal

nilmerg:
## Read-Only Detail

image

### Impossible Match Warning

Added because the form currently does not validate this and it may be the case due to migration anyway.

image

## Immediate Modal

image

## Conditional Modal

image

resolves #457


nilmerg:
@Jan-Schuppik It's quite a lot to review, but it's mostly refactoring anyway. I'd suggest to look at the commits during code review independently.

library/Notifications/Widget/EscalationRule/EscalationCondition.php
```diff
@@ -0,0 +1,385 @@
...
+                    array_intersect(
+                        array_map(fn (Severity $severity) => $severity->getValue(), $severities),
+                        array_map(fn (Severity $severity) => $severity->getValue(), $lowerSeverities)
+                    )
+                );
+            }
+        }
+
+        if (isset($byOperator['!='])) {
+            $severities = array_map(
```
Jan-Schuppik, line 195: This operator can not be used alone right now. 
An easy fix would be 
```suggestion
        if (isset($byOperator['!='])) {
        	if (empty($severities)) {  
        	    $severities = Severity::cases();  
        	}
            $severities = array_map(
```

library/Notifications/Widget/EscalationRule.php
```diff
@@ -0,0 +1,302 @@
...
+                ],
+                isset($recipient->schedule->name) => [
+                    new Icon(Icons::SCHEDULE, ['title' => $this->translate('Schedule')]),
+                    $recipient->schedule->name
+                ],
+            };
+
+            $channelLabel = Text::create(
+                $recipient->channel_label ?? $this->translate('Contact Default Channel')
+            );
```
Jan-Schuppik, line 280: $recipient->channel_label is always null here! 
The alias-prefix set by the subquery resolver (https://github.com/Icinga/ipl-orm/blob/main/src/Query.php#L677) gets truncated of the alias `channel_label` and it only exists `$recipient->nel_label`

application/controllers/EventRuleController.php
```diff
@@ -388,66 +129,60 @@ public function suggestAction(): void
...
-        $this->setTitle($this->translate('Edit Event Rule'));
+                $this->addContent($form);
+            })
+            ->on(Form::ON_SUBMIT, function (EventRuleForm $form) {
+                $rule = $form->getRule();
+
+                if ($form->hasBeenDeleted()) {
+                    Database::get()->transaction(
+                        fn(Connection $db) => (new EscalationRuleRepository(Database::get()))->delete($rule->id)
+                    );
```
Jan-Schuppik: The Icinga Web UI success notifications are missing. Previously they are only triggered if the rule was saved, but now there are more places where things are persisted, so i suggest the places where they could be handy:

```suggestion
                    );
                    Notification::success(sprintf(t('Successfully deleted event rule %s'), $rule->name));
```

application/controllers/EventRuleController.php
```diff
@@ -5,371 +5,112 @@
...
-                    $getJsonPaths($filter),
-                    $assisted,
-                    $form->getValue('filter_name')
-                ))->getJson()
-            );
-            $this->redirectNow(Links::eventRule($ruleId)->setParam('_filterOnly'));
-        })->handleRequest($this->getServerRequest());
+                Database::get()->transaction(
+                    fn(Connection $db) => (new EscalationRuleRepository(Database::get()))->update($rule)
+                );
```
Jan-Schuppik, line 84: Here would fit a success notification
```suggestion
                );
                Notification::success(t('Successfully updated object filters'));
```

application/controllers/EventRulesController.php
```diff
@@ -122,20 +124,24 @@ public function addAction(): void
...
                     Source::on(Database::get())->columns(['type'])->assembleSelect()->distinct()
                 )
             )
             ->setAction(Url::fromRequest()->getAbsoluteUrl())
-            ->on(Form::ON_SUBMIT, function ($form) {
+            ->on(Form::ON_SUBMIT, function (EventRuleForm $form) {
+                $ruleId = Database::get()->transaction(
+                    fn(Connection $db) => (new EscalationRuleRepository($db))->create($form->getRule())
+                );
+
```
Jan-Schuppik, line 141: Here would fit a success notification
```suggestion
                Notification::success(sprintf(t('New rule %s has successfully been added'), $form->getRule()->name));

```

application/controllers/RuleEscalationController.php
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                $form->setEscalation($escalation);
+            })->on(Form::ON_SUBMIT, function (EscalationForm $form) {
+                $escalation = $form->getEscalation();
+
+                if ($form->hasBeenDeleted()) {
+                    Database::get()->transaction(
+                        fn(Connection $db) => (new EscalationRepository($db))->delete($escalation->id)
+                    );
```
Jan-Schuppik, line 49: Here would fit a success notification
```suggestion
                    );
                    Notification::success(t('Successfully deleted escalation'));
```

application/controllers/RuleEscalationController.php
```diff
@@ -0,0 +1,63 @@
...
+                } else {
+                    Database::get()->transaction(
+                        fn(Connection $db) => (new EscalationRepository($db))->update($escalation)
+                    );
```
Jan-Schuppik, line 55: Here would fit a success notification
```suggestion
                    );
                    Notification::success(t('Successfully updated escalation'));
```

application/controllers/RuleEscalationsController.php
```diff
@@ -0,0 +1,54 @@
...
+            ->setCsrfCounterMeasureId(Session::getSession()->getId())
+            ->setAction(Url::fromRequest()->getAbsoluteUrl())
+            ->on(Form::ON_REQUEST, function ($_, EscalationForm $form) {
+                $form->populate([
+                    'rule_id' => $this->params->getRequired('rule'),
+                    'position' => $this->params->getRequired('position')
+                ]);
+            })->on(Form::ON_SUBMIT, function (EscalationForm $form) {
+                $escalation = $form->getEscalation();
+
```
Jan-Schuppik, line 42: Here would fit a success notification
```suggestion
                Notification::success(t('New escalation has successfully been added'));
                
```

application/controllers/EventRuleController.php
```diff
@@ -388,66 +129,60 @@ public function suggestAction(): void
...
+                        fn(Connection $db) => (new EscalationRuleRepository(Database::get()))->delete($rule->id)
+                    );
 
-        $this->addContent($eventRuleForm);
+                    $this->switchToSingleColumnLayout();
+                } elseif ($form->hasBeenDuplicated()) {
+                    $ruleId = Database::get()->transaction(
+                        fn(Connection $db) => (new EscalationRuleRepository(Database::get()))->duplicate($rule)
+                    );
+
```
Jan-Schuppik, line 180: duplicating an eventRule silently drops the object filters (EventRuleForm::getRule() always returns objectFilter=null).

an easy fix could be:
```suggestion
                } elseif ($form->hasBeenDuplicated()) {
                    $original = (new EscalationRuleRepository(Database::get()))->find($rule->id);  
                    $ruleId = Database::get()->transaction(  
                        fn(Connection $db) => (new EscalationRuleRepository(Database::get()))  
                            ->duplicate(  
                                new \Icinga\Module\Notifications\Form\Data\EscalationRule(  
                                    $rule->id,  
                                    $rule->name,  
                                    $rule->sourceType,  
                                    $original->object_filter  
                                )  
                            )  
                    );
```

application/controllers/EventRuleController.php
```diff
@@ -5,371 +5,112 @@
...
-                        ]));
-                    } else {
-                        $this->addPart($form->prepareConfigUpdate(
-                            $this->session->get('name'),
-                            $this->session->get('source_type')
-                        ));
-                        $this->addPart($form->prepareObjectFilterUpdate($this->session->get('object_filter')));
-                    }
+        $rule = (new EscalationRuleRepository(Database::get()))
+            ->find((int) $this->params->getRequired('id'));
```
Jan-Schuppik, line 43: Here is a check missing whether the rule actually exists.

```suggestion
        $rule = (new EscalationRuleRepository(Database::get()))
            ->find((int) $this->params->getRequired('id'));
        if ($rule === null) {
            $this->httpNotFound(t('Rule not found'));
        }
```

application/forms/RuleFilterForm.php
```diff
@@ -0,0 +1,225 @@
...
+            'rule_id' => $rule->id,
+            'rule_name' => $rule->name,
+            'source_type' => $rule->source_type
+        ];
+
+        if ($rule->object_filter) {
+            try {
+                $parsedFilter = json_decode($rule->object_filter, true, flags: JSON_THROW_ON_ERROR);
+            } catch (JsonException $e) {
+                Logger::error('Failed to parse rule filter configuration: %s (Error: %s)', $filter, $e);
```
Jan-Schuppik: i guess `$filter` (which doesn't exist here) should be `$rule->object_filter`

library/Notifications/Widget/EscalationRule/EscalationCondition.php
```diff
@@ -0,0 +1,385 @@
...
+        $hours = intdiv($remaining, 3600);
+        $remaining %= 3600;
+        $minutes = intdiv($remaining, 60);
+        $seconds = $remaining % 60;
+
+        if ($seconds > 0) {
+            if ($days > 0 || $hours > 0) {
+                // Be lazy for edge cases…
+                $this->addPhrase(Text::create(sprintf($this->translate(
+                    'Incident was opened %1$d days, %2$d hours, %2$d minutes and %2$d seconds ago'
```
Jan-Schuppik: here are just 2 of 4 values used 
```suggestion
                    'Incident was opened %1$d days, %2$d hours, %3$d minutes and %4$d seconds ago'
```

library/Notifications/Repository/EscalationRepository.php
Jan-Schuppik: If an escalation is removed the positions of the remaining escalations aren't updated. And because the widget derives the position for the "Add Escalation" link by counting the currently-rendered escalations, so it happens that it's not possible to create a new escalation:
`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '8-3' for key 'uk_rule_escalation_rule_id_position'`

A solution could be to replace this:

```
        $escalation->position = null;
        $escalation->rule_escalation_recipient = [];
        $escalation->delete();

        (new EntityManager($this->db))->save($escalation);
```
with this:
```
        $freedPositions = $escalation->position;
        $escalation->position = null;
        $escalation->rule_escalation_recipient = [];
        $escalation->delete();

        $entityManager = new EntityManager($this->db);
        $entityManager->save($escalation);

        $siblings = RuleEscalation::on($this->db)
            ->columns(['id', 'position'])
            ->filter(Filter::equal('rule_id', $escalation->rule_id))
            ->filter(Filter::greaterThan('position', $freedPositions))
            ->orderBy('position', SORT_ASC);

        foreach ($siblings as $sibling) {
            $sibling->setNew(false);
            $sibling->position--;
            $entityManager->save($sibling);
        }
```

application/forms/EventRuleForm.php
```diff
@@ -40,28 +42,94 @@ public function setAvailableSourceTypes(array $sourceTypes): static
...
+                'validators'    => [
+                    new CallbackValidator(function ($value, $validator) use ($ruleId) {
+                        $rules = Rule::on(Database::get())
+                            ->columns('id')
+                            ->filter(Filter::equal('name', $value));
+                        if ($ruleId !== null && ! $this->hasBeenDuplicated()) {
+                            $rules->filter(Filter::unequal('id', $ruleId));
+                        }
+
+                        if ($rules->first() !== null) {
```
Jan-Schuppik, line 128: I'm not sure if it's intended that an event rule can't be deleted if the name in the form is changed to an existing one (like it must be changed for duplicating). But the error message `An event rule with this name already exists` doesn't fit at all.
So i guess the deletion shouldn't care about the name which is temporary inserted in the form.
```suggestion
                        if ($rules->first() !== null && ! $this->hasBeenDeleted()) {
```

Jan-Schuppik (commented):
That's everything I found on the functionality side. 
CSS-related stuff still to come.

library/Notifications/Widget/EscalationRule.php
nilmerg: That's right, but I'm on https://github.com/Icinga/ipl-orm/pull/171 where this issue does not exist.

application/controllers/EventRuleController.php
nilmerg: Nah, that's the repository's responsibility because `null` means no change, hence it should apply the originals filter automatically.

library/Notifications/Widget/EscalationRule.php
Jan-Schuppik: Check 👍

application/controllers/EventRuleController.php
Jan-Schuppik: That's a good point.

public/css/detail/event-rule-detail.less
```diff
@@ -218,18 +256,29 @@
...
+    .outline,
+    .description {
+      .rounded-corners(2em);
+
+      &:has(.callout) .callout {
+        .rounded-corners(2em);
+      }
+    }
+
+    .escalations {
```
Jan-Schuppik, line 278: To keep the padding of .outline consistent in the corners, the border radius of the inner box needs to be the border radius of the outer box minus the padding between them: 
`radiusInner = radiusOuter - padding`
reason:
`The centers of the two circles defined by the border radii must be positioned exactly one above the other to maintain a constant distance between the two circles outlines.`
```suggestion
    .outline,
    .description {
      .rounded-corners(calc(2em - .5em));

      &:has(.callout) .callout {
        .rounded-corners(calc(2em - .5em));
      }
    }

    .outline {
      .rounded-corners(2em);
    }
```

public/css/detail/event-rule-detail.less
```diff
@@ -1,91 +1,143 @@
...
+    // Perfect round edges, always
+    .rounded-corners(9999px);
 
-        // fa spinner
-        content: '\f110';
+    &:has(.callout) {
+      display: contents;
+
+      .callout {
+        .rounded-corners(9999px);
```
Jan-Schuppik, line 46: Changing the border radius to the same as in two-column-view, would apply the perfect round edges only if the escalation contains exact one line, but i would solve the layout overflow.

```suggestion
    .rounded-corners(calc(2em - .5em));

    &:has(.callout) {
      display: contents;

      .callout {
        .rounded-corners(calc(2em - .5em));

```

(this change needs the suggestion that adds the border radius to .outline to keep the padding in the corner consistent)

public/css/detail/event-rule-detail.less
```diff
@@ -1,91 +1,143 @@
...
-  }
-
-  .escalation-condition,
-  .escalation-recipient {
-    > :first-child,
-    > :first-child[type="hidden"] + * {
-      .rounded-corners(.25em);
-      border-top-right-radius: 0;
-      border-bottom-right-radius: 0;
+  .outline {
```
Jan-Schuppik, line 51: this keeps the corner padding consistent
```suggestion
  .outline {
    .rounded-corners(2em);
    
```

Jan-Schuppik (commented):
The points mentioned in the previous review cycle are all solved!
 
# CSS related

## Missing shape change in single-column view
We don't care that much about mobile device formats, but applying the format change (transform to a rectangle shape if there isn't enough space) also in single-column view would help:

Tablet screen:
Image

Mobile screen:
Image
here the shape-change don't solve this completely, but it should be enough for this PR

## Immediate rule with multiple recipients looks a little bit off

The immediate escalation grows unlike the other escalation, so that it's connector-line isn't centered if there are more than one recipient on it:
Image
Image

## Many recipients
on a large screen and if it turns to rectangle shape, it fits many recipients without looking weird:
Image
Image

But on small screens in single-column view the inner boxes outgrow the outer box and the text overflows its space:
Image

## Names that are too long are not truncated
on a large screen even source names with max count of chars (255) fit in the escalation-layout (but contact names can be much longer [what fits in the type text in sql]). But if the box of the recipient-description grows the centered 'edit-escalation button' seems out of place: 
Image

But on a smaller screen or in two-column view, the overflow of a not wrappable text breaks the layout:

Image

Image

### The event-rules view would also benefit from shortening the contact names

Image

public/css/detail/event-rule-detail.less
```diff
@@ -218,18 +285,24 @@
...
   }
 }
 
-#layout.twocols:not(.wide-layout) {
-  .event-rule-config {
-    #object-filter-controls {
-      width: fit-content;
+// Based on font families set by icingaweb2/public/css/icinga/responsive.less
+#layout.compact-layout,
+#layout.poor-layout,
```
Jan-Schuppik: The default-layout is missing here

```suggestion
#layout.default-layout,
#layout.poor-layout,
```

Jan-Schuppik (commented):
The layout looks good/usable in every relevant format (if the following change is added)

application/forms/EscalationForm.php
```diff
@@ -0,0 +1,165 @@
...
+
+namespace Icinga\Module\Notifications\Forms;
+
+use Icinga\Module\Notifications\Form\ConfigProviderInterface;
+use Icinga\Module\Notifications\Form\Data\Escalation;
+use Icinga\Module\Notifications\Forms\EscalationForm\EscalationConditions;
+use Icinga\Module\Notifications\Forms\EscalationForm\EscalationRecipients;
+use Icinga\Module\Notifications\Model\RuleEscalation;
+use ipl\Html\Attributes;
+use ipl\Html\HtmlDocument;
```
Jan-Schuppik: The import `ipl\Html\HtmlDocument` isn't used anymore

```suggestion
```

library/Notifications/Repository/EscalationRepository.php
Jan-Schuppik: The import `LogicException` isn't used anymore

test/php/library/Notifications/Repository/EscalationRuleRepositoryTest.php
Jan-Schuppik: The `Icinga\Module\Notifications\Form\Data\EscalationRule` constructor has now only 4 args; the 5th constructor arg (`[]`) here is misleading

application/forms/EscalationForm/ConfigProvider.php
Jan-Schuppik: The dockblock is outdated, now it's used by `{@see EscalationForm\EscalationRecipient}` and `{@see EscalationRecipients}`

application/controllers/EventRuleController.php
```diff
@@ -388,66 +137,72 @@ public function suggestAction(): void
...
+                        fn(Connection $db) => (new EscalationRuleRepository($db))->update($rule)
+                    );
 
-        $this->addContent($eventRuleForm);
+                    Notification::success(sprintf(
+                        $this->translate('Updated escalation rule "%s"'),
+                        $rule->name
+                    ));
+                    $this->sendExtraUpdates(['#col1']);
+                    $this->closeModalAndRefreshRelatedView(Links::eventRule($rule->id));
```
Jan-Schuppik: Is there a reason that `CompatController::closeModalAndRefreshRemainingViews()` isn't used here?
```suggestion
                    $this->closeModalAndRefreshRemainingViews(Links::eventRule($rule->id));
```

application/controllers/RuleEscalationController.php
```diff
@@ -0,0 +1,68 @@
...
+
+                    Notification::success($this->translate('Updated escalation'));
+                }
+
+                $this->sendExtraUpdates(['#col1']);
+                $this->closeModalAndRefreshRelatedView(Links::eventRule($escalation->ruleId));
```
Jan-Schuppik: Here the same as above `$this->closeModalAndRefreshRemainingViews(Links::eventRule($escalation->ruleId))` could be used

application/controllers/RuleEscalationsController.php
```diff
@@ -0,0 +1,56 @@
...
+                Database::get()->transaction(
+                    fn(Connection $db) => (new EscalationRepository($db))->create($escalation)
+                );
+
+                Notification::success($this->translate('Created escalation'));
+                $this->sendExtraUpdates(['#col1']);
+                $this->closeModalAndRefreshRelatedView(Links::eventRule($escalation->ruleId));
```
Jan-Schuppik: Here the same as above `$this->closeModalAndRefreshRemainingViews(Links::eventRule($escalation->ruleId))` could be used

application/forms/EventRuleForm.php
```diff
@@ -40,28 +42,98 @@ public function setAvailableSourceTypes(array $sourceTypes): static
...
+            null
+        );
+    }
+
+    /**
+     * Check if the delete button was pressed
+     *
+     * @return bool
+     */
+    public function hasBeenDeleted(): bool
```
Jan-Schuppik: Is it intended for this method name to differ from the methods of the already existing forms `hasBeenRemoved()`?

application/forms/EscalationForm.php
```diff
@@ -0,0 +1,165 @@
...
+            (int) $this->getValue('rule_id')
+        );
+    }
+
+    /**
+     * Check if the delete button was pressed
+     *
+     * @return bool
+     */
+    public function hasBeenDeleted(): bool
```
Jan-Schuppik: same question here: differs from `hasBeenRemoved`

Jan-Schuppik (commented):
Here are a few more small cleanups and nitpicks, then everything should be fine

application/forms/EventRuleForm.php
nilmerg: yes, there's a difference between delete and remove and in this case, delete is more appropriate. That's also the case in all other forms, but not the topic of this change.

Jan-Schuppik: check 👍

Jan-Schuppik (approved):
It looks good to me 👍

PR #559, opened 2026-09-15
Add rendering for conditional configuration options

### LINKED ISSUE #554: Support conditional channel configuration options
https://github.com/Icinga/icinga-notifications/issues/514 is about the introduction to allow plugins to describe options that rely on other ones and are only visible when a specific decision is made by the user. This is essentially what many of our existing forms do and allow by utilizing the so-called *autosubmit* feature.

The method `\Icinga\Module\Notifications\Forms\ChannelForm::createConfigElements` is responsible to render a plugin's configuration options and `\Icinga\Module\Notifications\Forms\ChannelForm::getChannel` is responsible to prepare the result that's stored in the database.

## Rendering

The basic functionality is already part of the form and it's behavior can be observed by changing the channel *type*. A similar approach should be suitable here. All children of a given option must be rendered in the same order as they appear in the JSON.

Note that, for now the following limitations apply and need to be enforced:

- Only element types `bool` and `option[s]` are supported
- Children must not have children themselves

Automatic submission also has a flaw that gets overlooked often: In case the decision causes an element to have different defaults, making another choice should change to the correct default and not keep the default of the previous decision. Since this is a generalized implementation and a plugin does not have control over this, we should do better from the very start and ensure this works correctly.

## Serialization

Right now the options are all stored inside the same JSON object. This does not change and thus I don't expect many or even no changes at all here.

starlingun:
UI now supports the conditional rendering of children in channel configuration window. Drop-down lists and checkbox elements are allowed to have children that are going to be rendered depending on user's choice, e.g. depending on selected encryption type different options for certificates can show up. Works only with dropdowns and checkboxes, and children themselves cannot have children.

Closes #554


starlingun:
Closes #554

application/forms/ChannelForm.php
```diff
@@ -99,6 +99,7 @@ public function getChannel(): ChannelData
      */
     protected function assemble(): void
     {
+
```
nilmerg, line 102: please avoid unnecessary diff noise/whitespace changes like this.

application/forms/ChannelForm.php
```diff
@@ -158,7 +159,10 @@ protected function assemble(): void
 
         /** @var string $selectedType */
         $selectedType = $this->getValue('type');
-        $this->createConfigElements($selectedType, $typesConfig[$selectedType]);
+        $typeConfig = json_decode($typesConfig[$selectedType], true);
```
nilmerg, line 162: Please keep the inline variable annotation. This way the variable type matches the argument type.

application/forms/ChannelForm.php
```diff
@@ -232,15 +236,14 @@ public function hasBeenSubmitted(): bool
     }
 
     /**
-     * Create config elements for the given channel type
+     * Create config elements for the given configuration object type
      *
-     * @param string $type The channel type
-     * @param string $config The channel type config
+     * @param string $type The configuration object type
+     * @param array<int, ChannelOptionConfig>  $elementsConfig The elements type config
```
nilmerg, line 242: Single space as separator only please, unless indented. (i.e. all args to the same indentation level)

application/forms/ChannelForm.php
```diff
@@ -232,15 +236,14 @@ public function hasBeenSubmitted(): bool
...
      */
-    protected function createConfigElements(string $type, string $config): void
+    protected function createConfigElements(string $type, array $elementsConfig): void
     {
-        /** @var array<int, ChannelOptionConfig>  $elementsConfig */
-        $elementsConfig = json_decode($config, true);
+
```
nilmerg, line 246: Please watch out for remaining whitespace after refactorings and clean up accordingly. In this case, as previously mentioned already, a method/function body should not start with an empty line.

application/forms/ChannelForm.php
```diff
@@ -318,6 +347,10 @@ protected function getElementOptions(array $elementConfig): array
         if ($elementConfig['type'] === 'bool') {
             $options['checkedValue'] = 'checked';
             $options['uncheckedValue'] = 'unchecked';
+            $options['class'] = 'autosubmit';
```
nilmerg, line 350: This makes all elements of type `bool` capable of autosubmission for no reason, am I right? This should depend on the children's existence.

application/forms/ChannelForm.php
```diff
@@ -264,14 +267,25 @@ protected function createConfigElements(string $type, string $config): void
...
+            $this->addElement($configFieldset);
+        }
+
 
         foreach ($elementsConfig as $elementConfig) {
+            if ($type !== 'child') {
+                $elementName = $type . "_" . $elementConfig['name'];
+            } else {
+                $elementName = $elementConfig['name'];
+            }
```
nilmerg, line 283: We discussed prefixing all elements, didn't we? I mean, the children now have no prefix to make them more unique and thus are still subject to accidental default overrides.

application/forms/ChannelForm.php
```diff
@@ -280,6 +294,21 @@ protected function createConfigElements(string $type, string $config): void
             }
 
             $configFieldset->addElement($elem);
+
+            if ($type !== "child" && ($elem->getTag() === 'select' || $elem->getTag() === 'checkbox')) {
```
nilmerg, line 298: Please don't use the element's tag. Use the config type the element type has been inferred from prior.

application/forms/ChannelForm.php
```diff
@@ -280,6 +294,21 @@ protected function createConfigElements(string $type, string $config): void
...
+                $selectedOption = $elem->getValue();
+                $children = $elementConfig['children'];
+
+                if ($selectedOption !== null && $children !== null) {
+                    foreach ($children as $child) {
+                        $parents = $child["parent_values"];
+                        if (in_array($selectedOption, $parents)) {
```
nilmerg, line 305: `in_array` has the habit of being non-strict unless being told so. Please pass it always `strict: true` in order to avoid having to deal with false-positives. Remember, this is PHP after all, and so [type juggling applies](https://www.php.net/manual/en/language.types.type-juggling.php).

application/forms/ChannelForm.php
nilmerg: Eh, just saw that you change the element's name beforehand. That's a sign of not properly adhering to https://en.wikipedia.org/wiki/Separation_of_concerns. There are other signs such as the `$type` always being `child` during recursion and the `array()` wrap below. If you don't know what I mean, let's discuss this first before applying other changes.

nilmerg (changes requested):

nilmerg (changes requested):
And having it tested just now, I get this when opening the channel form:

```
Undefined array key "children"
#0 /usr/share/icingaweb2-modules/notifications/application/forms/ChannelForm.php(300): Icinga\Application\ApplicationBootstrap->{closure:Icinga\Application\ApplicationBootstrap::setupErrorHandling():578}(Integer, String, String, Integer)
#1 /usr/share/icingaweb2-modules/notifications/application/forms/ChannelForm.php(164): Icinga\Module\Notifications\Forms\ChannelForm->createConfigElements(String, Array)
#2 /usr/local/src/ipl-html/src/HtmlDocument.php(286): Icinga\Module\Notifications\Forms\ChannelForm->assemble()
#3 /usr/local/src/ipl-html/src/Form.php(235): ipl\Html\HtmlDocument->ensureAssembled()
#4 /usr/share/icingaweb2-modules/notifications/application/controllers/ChannelsController.php(128): ipl\Html\Form->handleRequest(Object(GuzzleHttp\Psr7\ServerRequest))
#5 /usr/share/icinga-php/vendor/vendor/icinga/zf1/library/Zend/Controller/Action.php(528): Icinga\Module\Notifications\Controllers\ChannelsController->addAction()
#6 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(78): Zend_Controller_Action->dispatch(String)
#7 /usr/share/icinga-php/vendor/vendor/icinga/zf1/library/Zend/Controller/Front.php(954): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#8 /icingaweb2/library/Icinga/Application/Web.php(255): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#9 /icingaweb2/library/Icinga/Application/webrouter.php(107): Icinga\Application\Web->dispatch()
#10 /icingaweb2/public/index.php(6): require_once(String)
#11 {main}
```

The config in the `available_channel_type.config_attrs` column for the mail channel is as follows:

```json
[{"name":"host","type":"string","label":{"de_DE":"SMTP Host","en_US":"SMTP Host"},"required":true,"min":null,"max":null},{"name":"port","type":"number","label":{"de_DE":"SMTP Port","en_US":"SMTP Port"},"required":true,"min":1,"max":65535},{"name":"sender_name","type":"string","label":{"de_DE":"Absendername","en_US":"Sender Name"},"default":"Icinga","required":true,"min":null,"max":null},{"name":"sender_mail","type":"string","label":{"de_DE":"Absenderadresse","en_US":"Sender Address"},"required":true,"min":null,"max":null},{"name":"user","type":"string","label":{"de_DE":"SMTP Benutzer","en_US":"SMTP User"},"help":{"de_DE":"Das Setzen eines SMTP Benutzers erfordert ebenfalls ein SMTP Passwort.","en_US":"When configuring an SMTP user, an SMTP password must also be set."},"min":null,"max":null},{"name":"password","type":"secret","label":{"de_DE":"SMTP Passwort","en_US":"SMTP Password"},"min":null,"max":null},{"name":"encryption","type":"option","label":{"de_DE":"SMTP Transportverschlüsselung","en_US":"SMTP Transport Encryption"},"required":true,"options":{"none":"None","starttls":"STARTTLS","tls":"TLS"},"min":null,"max":null,"children":[{"name":"ca_cert","type":"string","label":{"de_DE":"CA Zertifikat","en_US":"CA Certificate"},"help":{"de_DE":"Optionales CA Zertifikat zur Überprüfung des Zertifikats des SMTP Servers.","en_US":"Optional CA certificate to verify the SMTP server's certificate."},"min":null,"max":null,"parent_values":["starttls","tls"]},{"name":"client_cert","type":"string","label":{"de_DE":"Client Zertifikat","en_US":"Client Certificate"},"help":{"de_DE":"Optionales Client Zertifikat für gegenseitige TLS Authentifizierung.","en_US":"Optional client certificate for mutual TLS authentication."},"min":null,"max":null,"parent_values":["starttls","tls"]},{"name":"client_key","type":"secret","label":{"de_DE":"Client Schlüssel","en_US":"Client Key"},"help":{"de_DE":"Optionaler Client Schlüssel für gegenseitige TLS Authentifizierung.","en_US":"Optional client key for mutual TLS authentication."},"min":null,"max":null,"parent_values":["starttls","tls"]}]},{"name":"auth_method","type":"option","label":{"de_DE":"SMTP Authentifizierungsmethode","en_US":"SMTP Authentication Method"},"help":{"de_DE":"Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN benötigen den SMTP Benutzer und das Passwort, während OAUTHBEARER ein Bearer Token im SMTP Passwortfeld erwartet. Ist die NONE Methode gewählt, wird der SMTP Server ohne Authentifizierung kontaktiert.","en_US":"The method has to be supported by the SMTP server. PLAIN and LOGIN require the SMTP user and password, while OAUTHBEARER expects a bearer token in the SMTP password field. If the NONE method is chosen, the SMTP server will be contacted without authentication."},"required":true,"options":{"LOGIN":"LOGIN","OAUTHBEARER":"OAUTHBEARER","PLAIN":"PLAIN","none":"NONE"},"min":null,"max":null}]
```

application/forms/ChannelForm.php
nilmerg, line 1: there are still unnecessary white-space changes

application/forms/ChannelForm.php
```diff
@@ -158,7 +158,28 @@ protected function assemble(): void
 
         /** @var string $selectedType */
         $selectedType = $this->getValue('type');
-        $this->createConfigElements($selectedType, $typesConfig[$selectedType]);
+        /** @var array $typesConfig */
```
nilmerg: - The same type please
- Please ensure the variable name is correct, there's a superfluous *s*

```suggestion
        /** @var array<int, ChannelOptionConfig> $typeConfig */
```

application/forms/ChannelForm.php
nilmerg: Not done yet.

application/forms/ChannelForm.php
```diff
@@ -318,6 +340,10 @@ protected function getElementOptions(array $elementConfig): array
         if ($elementConfig['type'] === 'bool') {
             $options['checkedValue'] = 'checked';
             $options['uncheckedValue'] = 'unchecked';
+            if (isset($elementConfig['children'])) {
+                $options['class'] = 'autosubmit';
+                $options['children'] = $elementConfig['children'];
```
nilmerg: Didn't notice this prior, due the previous error, but why are the children added as option? For me this results in an additional `array to string conversion` error.

application/forms/ChannelForm.php
```diff
@@ -355,6 +385,7 @@ protected function getElementOptions(array $elementConfig): array
         return $options;
     }
 
+
```
nilmerg: If you have problems catching such, I'd suggest you take a look at the diff by yourself 😉

application/forms/ChannelForm.php
```diff
@@ -232,54 +253,55 @@ public function hasBeenSubmitted(): bool
...
+            $elementType = $this->getElementType($elementConfig['type']);
+
             /** @var BaseFormElement $elem */
             $elem = $this->createElement(
-                $this->getElementType($elementConfig['type']),
-                $elementConfig['name'],
-                $this->getElementOptions($elementConfig)
+                $elementType,
+                $elementName,
+                $this-> getElementOptions($elementConfig)
```
nilmerg: superfluous space after `->`

application/forms/ChannelForm.php
```diff
@@ -232,54 +253,55 @@ public function hasBeenSubmitted(): bool
...
+                $children = $elementConfig['children'];
+                $childrenToRender = [];
+
+                if ($selectedOption !== null) {
+                    foreach ($children as $child) {
+                        $parents = $child["parent_values"];
+                        if (in_array($selectedOption, $parents, true)) {
+                            $childrenToRender[] = $child;
+                        }
+                    }
```
nilmerg, line 299: please put an empty line after closing braces if an expression follows.

application/forms/ChannelForm.php
```diff
@@ -335,6 +361,10 @@ protected function getElementOptions(array $elementConfig): array
             if ($elementConfig['type'] === 'options') {
                 $options['multiple'] = true;
             }
+            if (isset($elementConfig['children'])) {
```
nilmerg, line 363: empty line prior please

nilmerg (changes requested):
please rebase with main and make sure my commit is gone here.

Also, have you tried opening a channel with the new dependent options set? I suppose we missed this when talking about changing the names of elements.

application/forms/ChannelForm.php
```diff
@@ -158,7 +158,28 @@ protected function assemble(): void
...
+                                'Could not decode options for type \'%s\'.'
+                                . ' Check if your database\'s character set is correctly configured.'
+                            ),
+                        )
+                    )
+                )
+            );
+        }
+
+        $this->createConfigElements($selectedType, $typeConfig, false);
```
nilmerg: `createConfigElements` was returning early in case of a empty config and did not create any elements, so this should be in the else-branch of this if expression.

nilmerg (changes requested):
As promised earlier, the currently serialized configuration is invalid as the email channel e.g. expects `ca_cert` but we store it as e.g. `tls_ca_cert`.

oxzi:
Thanks for this PR. To get this working with a checkbox as a parent, I patched it as follows:

```diff
diff --git a/application/forms/ChannelForm.php b/application/forms/ChannelForm.php
index f658d0f..5594ddc 100644
--- a/application/forms/ChannelForm.php
+++ b/application/forms/ChannelForm.php
@@ -293,6 +293,9 @@ class ChannelForm extends CompatForm
                 if ($selectedOption !== null) {
                     foreach ($children as $child) {
                         $parents = $child["parent_values"];
+                       if ($elementType === 'checkbox' && $selectedOption === 'checked') {
+                           $selectedOption = true;
+                       }
                         if (in_array($selectedOption, $parents, true)) {
                             $childrenToRender[] = $child;
                         }
```

PR #560, opened 2026-09-15
Properly re-render forms when being auto-submitted

nilmerg:
Form event handling is still not fully reliable, thus every controller that depends on events only listens for `Form::ON_SENT` now to detect auto submissions.

This slipped through during #557.


Jan-Schuppik (approved):
all the controller actions and forms work again, the no-content error on auto-submit is gone

